PR #19, opened 2018-08-18
Different colors for labels based on their hash

sandhose:
A hash is computed for each label, which is then used to determine the label's color.
In the future the label's colors might be somehow configurable, but at least they now have different colors.

image

image

(Part of the diff is due to [prettier](https://prettier.io) re-formatting the code)


MichaelMure:
It does looks neat :+1:

PR #20, opened 2018-08-18
Update bug.go

dispostable420:
hello


MichaelMure:
Alright, that seems completely boggus.

PR #40, opened 2018-09-04
docs: fix terminal UI recording src

tplk:


tplk:
Not sure why git registers a difference in the last line.
Probably something to do with line endings, I edited the file via GitHub UI.

PR #41, opened 2018-09-09
Cleanup queries.md

j-f1:
This cleans up the wording a little and clarifies a few points.

### DIFF
--- a/doc/queries.md
+++ b/doc/queries.md
@@ -1,6 +1,6 @@
 # Searching bugs
 
-You can search bugs using a micro query language for both filtering and sorting. A query could looks like this:
+You can search bugs using a micro query language for both filtering and sorting. A query could look like this:
 
 ```
 status:open sort:edit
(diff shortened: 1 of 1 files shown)


doc/queries.md
```diff
@@ -1,16 +1,16 @@
...
 ```
 
 A few tips:
 
-- query are case insensitive.
-- you can combine as much qualifier as you want.
-- you can use quotation for multi-word search term. For example, use `author:"René Descartes"` to search with a complete name.
+- queries are case insensitive.
+- you can combine as many qualifiers as you want.
+- you can use double quotes for multi-word search terms. For example, `author:"René Descartes"` searches for bugs opened by René Descartes, whereas `author:René Descartes` looks for issues opened by anyone named René containing “Descartes” in their content.
```
j-f1: Does the search normalize diacritical marks (i.e. searching for “rene” would still find issues opened by someone named “René”)?

MichaelMure: It should, I'm not sure it is already working though. I will try to make it work.

MichaelMure: `author:` will only match on the person that opened the bug. Comments are not indexed at the moment, it will come at a later stage.

MichaelMure:
Thanks, really helpful :+1: 

It's a pain to write these ...

doc/queries.md
j-f1: Does https://github.com/MichaelMure/git-bug/pull/41/commits/fc0f058a00eb190d79fcaff8b63dc1f669b89d45 make this better?

MichaelMure: It will be correct eventually, but text search is not implemented at the moment because it would require a clever full-text indexing. `author:"René Descartes"` will return an error.

MichaelMure:
Awesome, thanks !

PR #57, opened 2018-10-02
Add open/close in bug view

### LINKED ISSUE #29: feat: Close bug from termui's show view
Hello,

Thanks for your work on this! I looked through the source code and I don't think I'm missing anything, but it would be great to be able to close a bug via `termui` when you are in a particular bug's view.

If that is in fact a feature that's missing, I'd be happy to work on a PR to add it. My apologies if it already exists and I just haven't figured it out yet! :smile:

Thanks!

adamslc:
Makes the key 'o' toggle a bug open or closed in the `termui` bug view. Feel free to bikeshed the specific key to activate this functionality (or even if a toggle behavior is appropriate).

I have very little experience with go, so if this isn't idiomatic, feel free to point it out.

Closes  #29.

### DIFF
--- a/termui/show_bug.go
+++ b/termui/show_bug.go
@@ -94,12 +94,12 @@ func (sb *showBug) layout(g *gocui.Gui) error {
 	}
 
 	v.Clear()
-	fmt.Fprintf(v, "[q] Save and return [←↓↑→,hjkl] Navigation [o] Toggle open/close ")
+	fmt.Fprintf(v, "[q] Save and return [←↓↑→,hjkl] Navigation ")
 
 	if sb.isOnSide {
 		fmt.Fprint(v, "[a] Add label [r] Remove label")
 	} else {
-		fmt.Fprint(v, "[c] Comment [t] Change title")
+		fmt.Fprint(v, "[o] Toggle open/close [c] Comment [t] Change title")
 	}
 
 	_, err = g.SetViewOnTop(showBugInstructionView)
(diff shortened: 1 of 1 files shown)


termui/show_bug.go
```diff
@@ -94,7 +94,7 @@ func (sb *showBug) layout(g *gocui.Gui) error {
 	}
 
 	v.Clear()
-	fmt.Fprintf(v, "[q] Save and return [←↓↑→,hjkl] Navigation ")
+	fmt.Fprintf(v, "[q] Save and return [←↓↑→,hjkl] Navigation [o] Toggle open/close ")
```
MichaelMure: To avoid displaying too much instructions at the same time, it would be best to display this one only when the cursor is on the bug timeline. Could you move it a few line below, with the instructions for comments/title ?

termui/show_bug.go
```diff
@@ -603,6 +609,14 @@ func (sb *showBug) setTitle(g *gocui.Gui, v *gocui.View) error {
 	return setTitleWithEditor(sb.bug)
 }
 
+func (sb *showBug) toggleOpenClose(g *gocui.Gui, v *gocui.View) error {
+	if sb.bug.Snapshot().Status.String() == "open" {
```
MichaelMure: Could you use a switch/case with the constants (https://github.com/MichaelMure/git-bug/blob/master/bug/status.go#L12-L13) instead of a string ? In the future, a bug might have more status than just open/close, might as well prepare the code for that.

PR #58, opened 2018-10-03
Make comments editable in termui

adamslc:
This PR makes it possible to edit comments in `termui` mode with 'e'. As it stands now, only comments are editable, however it might make sense to put more functionality under edit. For example, editing labels and the title could also be triggered with 'e' when the appropriate view is selected.

Also, there seems to be a lot of code duplication in `termui`. Would you be welcome to a PR to reduce this duplication?

### DIFF
--- a/input/input.go
+++ b/input/input.go
@@ -113,6 +113,42 @@ func BugCommentEditorInput(repo repository.RepoCommon) (string, error) {
 	return message, nil
 }
 
+const bugEditCommentTemplate = `%s
+
+# Please enter the comment message. Lines starting with '#' will be ignored,
+# and an empty message aborts the operation.
+`
+
+// BugEditCommentEditorInput will open the default editor in the terminal with a
+// template for the user to fill. The file is then processed to extract a comment.
+func BugEditCommentEditorInput(repo repository.RepoCommon, preMessage string) (string, error) {
+	template := fmt.Sprintf(bugEditCommentTemplate, preMessage)
+	raw, err := launchEditorWithTemplate(repo, messageFilename, template)
+
+	if err != nil {
+		return "", err
+	}
+
+	lines := strings.Split(raw, "\n")
+
+	var buffer bytes.Buffer
+	for _, line := range lines {
+		if strings.HasPrefix(line, "#") {
+			continue
+		}
+		buffer.WriteString(line)
+		buffer.WriteString("\n")
+	}
+
+	message := strings.TrimSpace(buffer.String())
+
+	if message == "" {
+		return "", ErrEmptyMessage
+	}
+
+	return message, nil
+}
+
 const bugTitleTemplate = `%s
 
 # Please enter the new title. Only one line will used.
--- a/input/input.go
+++ b/input/input.go
@@ -78,7 +78,7 @@ func BugCreateEditorInput(repo repository.RepoCommon, preTitle string, preMessag
 	return title, message, nil
 }
 
-const bugCommentTemplate = `
+const bugCommentTemplate = `%s
 
 # Please enter the comment message. Lines starting with '#' will be ignored,
 # and an empty message aborts the operation.
@@ -86,43 +86,8 @@ const bugCommentTemplate = `
 
 // BugCommentEditorInput will open the default editor in the terminal with a
 // template for the user to fill. The file is then processed to extract a comment.
-func BugCommentEditorInput(repo repository.RepoCommon) (string, error) {
-	raw, err := launchEditorWithTemplate(repo, messageFilename, bugCommentTemplate)
-
-	if err != nil {
-		return "", err
-	}
-
-	lines := strings.Split(raw, "\n")
-
-	var buffer bytes.Buffer
-	for _, line := range lines {
-		if strings.HasPrefix(line, "#") {
-			continue
-		}
-		buffer.WriteString(line)
-		buffer.WriteString("\n")
-	}
-
-	message := strings.TrimSpace(buffer.String())
-
-	if message == "" {
-		return "", ErrEmptyMessage
-	}
-
-	return message, nil
-}
-
-const bugEditCommentTemplate = `%s
-
-# Please enter the comment message. Lines starting with '#' will be ignored,
-# and an empty message aborts the operation.
-`
-
-// BugEditCommentEditorInput will open the default editor in the terminal with a
-// template for the user to fill. The file is then processed to extract a comment.
-func BugEditCommentEditorInput(repo repository.RepoCommon, preMessage string) (string, error) {
-	template := fmt.Sprintf(bugEditCommentTemplate, preMessage)
+func BugCommentEditorInput(repo repository.RepoCommon, preMessage string) (string, error) {
+	template := fmt.Sprintf(bugCommentTemplate, preMessage)
 	raw, err := launchEditorWithTemplate(repo, messageFilename, template)
 
 	if err != nil {
--- a/termui/select_popup.go
+++ b/termui/select_popup.go
@@ -1,0 +1,217 @@
+package termui
+
+import (
+	"fmt"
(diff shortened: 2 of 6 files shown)


MichaelMure:
> This PR makes it possible to edit comments in `termui` mode with 'e'. As it stands now, only comments are editable, however it might make sense to put more functionality under edit. For example, editing labels and the title could also be triggered with 'e' when the appropriate view is selected.

Interesting idea. I have to say, I have no well defined general UX plan for the termui. Suggestions welcome :)

> Also, there seems to be a lot of code duplication in `termui`. Would you be welcome to a PR to reduce this duplication?

Sure! I'm curious what you would deduplicate though. What the termui also need is more comments and tests. This part of the code is really behind the pack ...

input/input.go
```diff
@@ -113,6 +113,42 @@ func BugCommentEditorInput(repo repository.RepoCommon) (string, error) {
...
+		if strings.HasPrefix(line, "#") {
+			continue
+		}
+		buffer.WriteString(line)
+		buffer.WriteString("\n")
+	}
+
+	message := strings.TrimSpace(buffer.String())
+
+	if message == "" {
```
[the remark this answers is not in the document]
MichaelMure: The more I think about it, the more I think that:

1. CREATE should *allow* an empty comment
2. ADD_COMMENT should *deny* an empty comment
3. EDIT_COMMENT should *allow* an empty comment. In that case, it should be considered as a deleted comment.

termui/select_popup.go
```diff
@@ -0,0 +1,220 @@
...
+
+		v.Frame = true
+		v.Title = sp.title
+	}
+
+	y0 += 1
+
+	for i, name := range sp.options {
+		viewname := fmt.Sprintf("view%d", i)
+		v, err = g.SetView(viewname, x0+2, y0, x0+width-2, y0+2)
```
golangcibot: ineffectual assignment to `err`

input/input.go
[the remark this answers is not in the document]
adamslc: That was what I was thinking. It might be nice to add a visual indicator of the deleted comment (e.g. replace the author with `(deleted)`) so it is clear what happened.

termui/show_bug.go
```diff
@@ -24,6 +24,7 @@ type showBug struct {
 	bug                *cache.BugCache
 	childViews         []string
 	mainSelectableView []string
+	mainTimelineIndex  []int
```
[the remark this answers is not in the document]
adamslc: I don't think that would be enough because I also need to get the message text from the timeline item. One thing that might work is using the index of the timeline as the `selected` variable.

Is there ever a time when the will be something in `mainSelectableView` that isn't in `Timeline` or vice versa?

adamslc:
I've added a new input widget, `selectPopup`, that is now used to add/remove labels. I also changed the edit commands for labels to only use 'e' now. Try it out and see what you think.

termui/show_bug.go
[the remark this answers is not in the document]
MichaelMure: You can use `Snapshot.SearchTimelineItem` to get the target operation and then get the message.

MichaelMure:
I tried your version, that's neat :)

A few things:

- have a look at `RepoCache.ValidLabels`, it will give you all the labels already used in the repo
- instead of a popup, maybe use the full window ? That is, implement the `window` interface, like `showBug` does. It will give you more space, solve your responsiveness problem and would avoid having a popup over another popup.
- could you do this change in another branch ? That way we can merge things as soon as they are ready

Thanks a lot !

adamslc:
Ok, I've updated `mainSelectableView` to use the hash, and made the label selector display all labels used in the repo. I will work on making the label selector full screen in a separate commit.

MichaelMure:
> I will work on making the label selector full screen in a separate commit.

I was thinking editable comment in a PR, label edition in another ;-)

I'll see if I can merge what is ready already.

adamslc:
Sorry, my git-foo is not that strong, but I just merged this branch with master locally and it was sufficiently annoying to show me the error in my ways...

MichaelMure:
Alright, don't move, I'll try to split this PR in two properly ;-)

MichaelMure:
Well actually the two features are already too merged together to easily split them. Let's try to finish both instead.

termui/show_bug.go
```diff
@@ -241,7 +233,7 @@ func (sb *showBug) renderMain(g *gocui.Gui, mainView *gocui.View) error {
 	y0 += lines + 1
 
 	for i, op := range snap.Timeline {
```
MichaelMure: `i` is not used anymore so the compilation fail. Use `for _, op := range snap.Timeline {`

adamslc:
If its alright with you, I think I'll just close this PR, sort out the mess locally and open new (separate PRs).

termui/select_popup.go
```diff
@@ -0,0 +1,220 @@
...
+}
+
+func(sp *selectPopup) selectItem(g *gocui.Gui, v*gocui.View) error {
+	sp.optionSelect[sp.selected] = !sp.optionSelect[sp.selected]
+
+	return nil
+}
+
+func (sp *selectPopup) addItem(g *gocui.Gui, v *gocui.View) error {
+	c := ui.inputPopup.Activate("Add item")
```
MichaelMure: "Add a new label"

termui/show_bug.go
```diff
@@ -605,33 +597,60 @@ func (sb *showBug) comment(g *gocui.Gui, v *gocui.View) error {
...
 
-func (sb *showBug) setTitle(g *gocui.Gui, v *gocui.View) error {
-	return setTitleWithEditor(sb.bug)
-}
+func (sb *showBug) editLabels(g *gocui.Gui, v *gocui.View, snap *bug.Snapshot) error {
+	repoLabels := sb.bug.Repo().ValidLabels()
+	bugLabels := snap.Labels
+	labelStrings := make([]string, len(repoLabels))
+	labelSelect := make([]bool, len(repoLabels))
+	for i, repoLabel := range repoLabels {
```
MichaelMure: You can simplify this:

```go
for i, repoLabel := range repoLabels {
		labelStrings[i] = string(repoLabel)

		for _, bugLabel := range bugLabels {
			if repoLabel == bugLabel {
				labelSelect[i] = true
			}
		}
	}
```

termui/show_bug.go
```diff
@@ -643,37 +662,46 @@ func (sb *showBug) addLabel(g *gocui.Gui, v *gocui.View) error {
...
+	if sb.isOnSide {
+		return sb.editLabels(g, v, snap)
+	}
 
-		labels := strings.FieldsFunc(input, func(r rune) bool {
-			return r == ' ' || r == ','
-		})
+	op, err := snap.SearchTimelineItem(git.Hash(sb.selected))
+	if err != nil {
+		ui.msgPopup.Activate(msgPopupErrorTitle, "Selected field is not editable.")
```
MichaelMure: This should not be a popup but a panic. You can only have an error here if something is wrong in the code.

termui/termui.go
```diff
@@ -230,6 +241,42 @@ func addCommentWithEditor(bug *cache.BugCache) error {
...
+	ui.g = nil
+
+	message, err := input.BugCommentEditorInput(ui.cache, preMessage)
+
+	if err != nil && err != input.ErrEmptyMessage {
+		return err
+	}
+
+	if err == input.ErrEmptyMessage {
+		// TODO: Allow comments to be deleted?
```
MichaelMure: Leave that as a TODO for now, we will deal with deleted comments later.

MichaelMure:
> If its alright with you, I think I'll just close this PR, sort out the mess locally and open new (separate PRs).

As you prefer. There is not that much work left to finish though

PR #62, opened 2018-10-05
Allow setting a default remote

### LINKED ISSUE #11: feat: configuration for the default remote
For now, the default git remote is hard-coded as `"origin"`. While this will work most of the time, having this being configurable would be nice.

Git handle this configuration separately for each branch so we can't reuse that system.

I guess the best way to do that would be using the git config system and having an entry like `gitbug.defaultremote=origin`.

In git-bug, we will also need a config package that read and expose this.

adamslc:
Adds some code to read and write gitconfig files, and uses it to enable default remotes.

This is based off code at [go-gitconfig](https://github.com/tcnksm/go-gitconfig). I would have just used that code, but the project seems abandoned, and it is licensed as MIT.

Closes #11.

### DIFF
--- a/commands/push.go
+++ b/commands/push.go
@@ -5,7 +5,12 @@ import (
 	"fmt"
 
 	"github.com/MichaelMure/git-bug/cache"
+	"github.com/MichaelMure/git-bug/util/git"
 	"github.com/spf13/cobra"
+)
+
+var (
+	defaultRemote bool
 )
 
 func runPush(cmd *cobra.Command, args []string) error {
@@ -13,9 +18,31 @@ func runPush(cmd *cobra.Command, args []string) error {
 		return errors.New("Only pushing to one remote at a time is supported")
 	}
 
-	remote := "origin"
+	if len(args) == 0 && defaultRemote {
+		str, err := git.GetConfig("gitbug.defaultremote")
+		if _, ok := err.(*git.ErrNotFound); ok {
+			return errors.New("No default remote set")
+		} else {
+			return err
+		}
+
+		fmt.Println("Default remote: ", str)
+		return nil
+	}
+
+	remote := ""
 	if len(args) == 1 {
 		remote = args[0]
+	} else if str, err := git.GetConfig("gitbug.defaultremote"); err == nil {
+		remote = str
+	} else {
+		return errors.New("No remote provided, and no defaults specified.")
+	}
+
+	if defaultRemote {
+		if err := git.SetConfig("gitbug.defaultremote", remote); err != nil {
+			return err
+		}
 	}
 
 	backend, err := cache.NewRepoCache(repo)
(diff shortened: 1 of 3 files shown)


MichaelMure:
Damn, you are on fire :tada: :tada: :tada: 

Instead of importing part of go-gitconfig, you can use the interface function `RepoCommon.ReadConfigs()`. It's implemented by both the low level `GitRepo` and `RepoCache`. This function will read all values matching a prefix. If that's inconvenient, you can add a new function to read a single entry.

commands/push.go
```diff
@@ -43,4 +70,9 @@ var pushCmd = &cobra.Command{
 
 func init() {
 	RootCmd.AddCommand(pushCmd)
+
+	pushCmd.Flags().SortFlags = false
+
+	pushCmd.Flags().BoolVarP(&defaultRemote, "default", "d", false,
+		"If a remote is provided, set the default remote for future pushes. Otherwise list the default remote if one exists.")
```
MichaelMure: In term of UX that's not good. The `push` command should not have a lasting effect on the config. All the git-bug config manipulation should be centralized in a new `config` command. I guess it would be better to open an issue, see what it should cover and design this new command.

adamslc: Hmm, I see the argument, but a standard `git push` can already have a lasting effect on the repo config with e.g. `--set-upstream`. Even if there is a config command, it still seems like you should be able to both push to a remote and set that remote as the default at the same time.

MichaelMure:
Sorry to rain on the parade :-\

Maybe before starting a new feature, contact me so I can give you the useful pointers and we can design the thing together. There is either the [gitter lobby](https://gitter.im/the-git-bug/Lobby), a github issue or my mail (in my github profile). As I'm starting a new job on Monday I will be a bit less responsive though.

Btw, if you are looking for a new target, it would be awesome to have a proper window in the termui to edit the search query, with widget and selectors (a bit like the toolbar in https://github.com/MichaelMure/git-bug/issues).

Again, sorry to be annoying :-|

adamslc:
No worries. This didn't take very long at all to knock up, so I'm not that attached to it.

MichaelMure:
This is still a valid feature but the code for commands has moved so much that it would need to be reimplemented. At the same time, reading the git config is now integral part of `Repo` so it would be way easier to do.

Sorry :-\

mcepl:
@adamslc Is there any hope that you would revive this PR? I usually reserve `origin` for the original repo I branch from, and so it is usually the one repository I have no access to.

sudoforge:
> @adamslc Is there any hope that you would revive this PR? I usually reserve `origin` for the original repo I branch from, and so it is usually the one repository I have no access to.

@mcepl The issue is still open: #62

This is something I'm loosely planning on including in `0.9.0`.

mcepl:
> @mcepl The issue is still open: #62

??? **This** is the issue 62.

sudoforge:
> Did you mean https://github.com/MichaelMure/git-bug/pull/1076 ?

Whoops! I was auto-completing using github's little drop-down with a search of "default remote" and must have selected the wrong entry.

I meant #11.

PR #70, opened 2018-10-22
[DOC] Usage examples in README.md

auyer:
Current README.md docs are missleading. 
1 - The "new" command has been changed, 
1 - git-bug is being represented without the dash  ( "git bug" instead of "git-bug").


MichaelMure:
> 1 - The "new" command has been changed,

Good catch !

> 1 - git-bug is being represented without the dash ( "git bug" instead of "git-bug").

This one is correct though. If the binary `git-bug` is in your `$PATH`, git will pick it up as a new sub-command.

auyer:
It wasn't happening in my path. I will conduct more tests latter.
For now, I rebased commit.

Issue #42, opened 2018-09-09
feat: catch properly Ctrl+C / SIGINT in commands


MichaelMure:
Right now, commands get killed by a SIGINT and the lock file is not properly removed.

As the cache is smart enough to see that the previous process is gone and clean the file, it's not that much of a problem, but it would be cleaner.

auyer:
I'll start looking into this issue.
Are there exported functions responsible by creating and removing the lock files ? If there are, the solution is even cleaner and I wont need to re-write any code.

MichaelMure:
Thanks.

What you need to trigger is `RepoCache.Close()`, that is the `backend.Close()` that you can see in most of the commands.

MichaelMure:
Dealt with by @auyer :+1:

PR #71, opened 2018-10-23
[WIP] Cleaning lock files on Interruption

auyer:
This PR attemps to solve issue #42 

### What was done: 
I created a "cleaner" package where all lockfile removal functions will be registered, and in can any interruption signal is recieved, the these functions will be executed first.

### How it was tested:
As far as I know, there is no way to create unit tests for this. ~~I created an [Issue](https://github.com/auyer/git-bug/issues/1) on my fork in case someone figures out how to test this.~~
As far as manual tests go, I repetedly performend manual interruptions on a long running `git bug ls` several times, and killed the text editor responsible by `git bug add`. 

Please, test it with your test cases, and report the results here.

### DIFF
--- a/cleaner/cleaner.go
+++ b/cleaner/cleaner.go
@@ -21,7 +21,8 @@ func Register(f t) {
 			ch := make(chan os.Signal, 1)
 			signal.Notify(ch, syscall.SIGINT, syscall.SIGTERM, os.Interrupt)
 			<-ch
-			fmt.Println("")
+                        // Prevent un-terminated ^C character in terminal
+			fmt.Println()
 			clean()
 			os.Exit(1)
 		}()
--- a/cleaner/cleaner.go
+++ b/cleaner/cleaner.go
@@ -1,0 +1,38 @@
+package cleaner
+
+import (
+	"fmt"
+	"os"
+	"os/signal"
+	"syscall"
+)
+
+type t func() error
+
+var cleaners []t
+var inactive bool
+
+// Register a cleaner function. When a function is registered, the Signal watcher is started in a goroutine.
+func Register(f t) {
+	cleaners = append(cleaners, f)
+	if !inactive {
+		inactive = false
+		go func() {
+			ch := make(chan os.Signal, 1)
+			signal.Notify(ch, syscall.SIGINT, syscall.SIGTERM, os.Interrupt)
+			<-ch
+			fmt.Println("")
+			clean()
+			os.Exit(1)
+		}()
+	}
+}
+
+// Clean invokes all registered cleanup functions
+func clean() {
+	fmt.Println("Cleaning")
+	for _, f := range cleaners {
+		_ = f()
+	}
+	cleaners = []t{}
+}
--- a/cleaner/cleaner.go
+++ b/cleaner/cleaner.go
@@ -7,7 +7,7 @@ import (
 	"syscall"
 )
 
-type t func() error
+type Cleaner func() error
 
 var cleaners []t
 var inactive bool
--- a/cleaner/cleaner.go
+++ b/cleaner/cleaner.go
@@ -9,11 +9,11 @@ import (
 
 type Cleaner func() error
 
-var cleaners []t
+var cleaners []Cleaner
 var inactive bool
 
 // Register a cleaner function. When a function is registered, the Signal watcher is started in a goroutine.
-func Register(f t) {
+func Register(f Cleaner) {
 	cleaners = append(cleaners, f)
 	if !inactive {
 		inactive = false
@@ -21,7 +21,7 @@ func Register(f t) {
 			ch := make(chan os.Signal, 1)
 			signal.Notify(ch, syscall.SIGINT, syscall.SIGTERM, os.Interrupt)
 			<-ch
-                        // Prevent un-terminated ^C character in terminal
+			// Prevent un-terminated ^C character in terminal
 			fmt.Println()
 			clean()
 			os.Exit(1)
@@ -35,5 +35,5 @@ func clean() {
 	for _, f := range cleaners {
 		_ = f()
 	}
-	cleaners = []t{}
+	cleaners = []Cleaner{}
 }
--- a/commands/add.go
+++ b/commands/add.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 
 	"github.com/MichaelMure/git-bug/cache"
+	"github.com/MichaelMure/git-bug/cleaner"
 	"github.com/MichaelMure/git-bug/input"
 	"github.com/spf13/cobra"
 )
@@ -22,6 +23,7 @@ func runAddBug(cmd *cobra.Command, args []string) error {
 		return err
 	}
 	defer backend.Close()
+	cleaner.Register(backend.Close)
 
 	if addMessageFile != "" && addMessage == "" {
 		addTitle, addMessage, err = input.BugCreateFileInput(addMessageFile)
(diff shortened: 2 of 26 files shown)


cleaner/cleaner.go
```diff
@@ -0,0 +1,38 @@
...
+			os.Exit(1)
+		}()
+	}
+}
+
+// Clean invokes all registered cleanup functions
+func clean() {
+	fmt.Println("Cleaning")
+	for _, f := range cleaners {
+		f()
```
golangcibot: Error return value of `f` is not checked

cleaner/cleaner.go
```diff
@@ -0,0 +1,38 @@
...
+var inactive bool
+
+// Register a cleaner function. When a function is registered, the Signal watcher is started in a goroutine.
+func Register(f t) {
+	cleaners = append(cleaners, f)
+	if !inactive {
+		inactive = false
+		go func() {
+			ch := make(chan os.Signal, 1)
+			signal.Notify(ch, syscall.SIGINT, syscall.SIGKILL, syscall.SIGTERM, os.Interrupt, os.Kill)
```
golangcibot: syscall.SIGKILL cannot be trapped (did you mean syscall.SIGTERM?)

cleaner/cleaner.go
```diff
@@ -0,0 +1,38 @@
...
+var inactive bool
+
+// Register a cleaner function. When a function is registered, the Signal watcher is started in a goroutine.
+func Register(f t) {
+	cleaners = append(cleaners, f)
+	if !inactive {
+		inactive = false
+		go func() {
+			ch := make(chan os.Signal, 1)
+			signal.Notify(ch, syscall.SIGINT, syscall.SIGTERM, os.Interrupt, os.Kill)
```
golangcibot: os.Kill cannot be trapped (did you mean syscall.SIGTERM?)

cleaner/cleaner.go
```diff
@@ -0,0 +1,38 @@
...
+// Register a cleaner function. When a function is registered, the Signal watcher is started in a goroutine.
+func Register(f t) {
+	cleaners = append(cleaners, f)
+	if !inactive {
+		inactive = false
+		go func() {
+			ch := make(chan os.Signal, 1)
+			signal.Notify(ch, syscall.SIGINT, syscall.SIGTERM, os.Interrupt)
+			<-ch
+			fmt.Println("")
```
MichaelMure: I suppose it's to have a clean line in the shell ? 
```suggestion
                        // Prevent un-terminated ^C character in terminal
			fmt.Println()
```

cleaner/cleaner.go
```diff
@@ -0,0 +1,38 @@
+package cleaner
```
MichaelMure: Could you move it to `util/interrupt` and name the main function `RegisterCleaner` ?

cleaner/cleaner.go
```diff
@@ -0,0 +1,38 @@
...
+
+import (
+	"fmt"
+	"os"
+	"os/signal"
+	"syscall"
+)
+
+type t func() error
```
MichaelMure:
```suggestion
type Cleaner func() error
```

cleaner/cleaner.go
```diff
@@ -0,0 +1,38 @@
...
+
+var cleaners []t
+var inactive bool
+
+// Register a cleaner function. When a function is registered, the Signal watcher is started in a goroutine.
+func Register(f t) {
+	cleaners = append(cleaners, f)
+	if !inactive {
+		inactive = false
```
MichaelMure: Shouldn't that be `inactive = true` ? Inverting the meaning and using `active` would make it clearer.

cleaner/cleaner.go
```diff
@@ -0,0 +1,38 @@
...
+			os.Exit(1)
+		}()
+	}
+}
+
+// Clean invokes all registered cleanup functions
+func clean() {
+	fmt.Println("Cleaning")
+	for _, f := range cleaners {
+		_ = f()
```
MichaelMure: Print the error if any ?

cleaner/cleaner.go
```diff
@@ -0,0 +1,38 @@
...
+	"syscall"
+)
+
+type t func() error
+
+var cleaners []t
+var inactive bool
+
+// Register a cleaner function. When a function is registered, the Signal watcher is started in a goroutine.
+func Register(f t) {
```
MichaelMure: Would be neat to have a [variadic](https://gobyexample.com/variadic-functions) here, so we can register multiple cleaner easily. And now that I think about it, the cleaners should be executed in the reverse order they have been added, to handle possible dependency between them (as `defer` in go does).

cleaner/cleaner.go
auyer: Precisely !

cleaner/cleaner.go
auyer: Moved.

cleaner/cleaner.go
auyer: I believe printing errors here would be unnecessary.

First, if the user is shutting the command down by force, it is likely that they do not want to see a bunch of information on the terminal.
Second but more important : If previous lockfiles had alredy been removed from disk by the deferables, a "file now found" error will occour, but that does not affect the user in any way. Implementing a "removal" function to remove previously registered cleaners that have became unnecessary is going to add complexity, and I see no real benefit.

Tell me what you prefere: leve it as is, print errors as is, or add the removal functionality.

cleaner/cleaner.go
auyer: You are perfeclty right. Done.

cleaner/cleaner.go
auyer: Done ! This was my first time using this concept in go ! Thanks !
It also inpired me to test the register function without uing the `os.Exit` part.

auyer:
> Sorry for the avalanche of comments

Thanks for the comments! I agree that they improved my contribuiton.
Please check the one I left unresolved.

cleaner/cleaner.go
MichaelMure: As far as I can tell, you are not reversing the order. A simple way to fix that would be to prepend cleaners instead of appending them. See https://codingair.wordpress.com/2014/07/18/go-appendprepend-item-into-slice/

MichaelMure:
Good thing to have some tests :+1:

cleaner/cleaner.go
auyer: I missed that part in your comment. Doing it now.

cleaner/cleaner.go
auyer: > You are writing a reusable package.

To make it more reusable, I am considering separating the anonymous function and giving it a name. 
It would be possible to summon it manually, or parsing it to the RegisterCleaner as an argument. 
```
func RegisterCleaner(watcher func(), f ...Cleaner) {
	for _, fn := range f {
		cleaners = append([]Cleaner{fn}, cleaners...)
		if !active {
			active = true
			go watcher()
		}
	}
}
```

The calling packages would do something like : 
```
interrupt.RegisterCleaner( interrupt.SignalWatcher, backend.Close)
```

It would be easy to create different watchers to deal with the errors in a different way, print in a different way...

How do you like this approach ?

MichaelMure: It's highly subjective, but I'd say that it would be over engineering. There is no need for that currently in the project and I don't foresee something that would use it, so better keep things simple.

auyer:
For future reference, in [this discussion](https://github.com/MichaelMure/git-bug/pull/71#discussion_r228334325) was suggested the functional way to deal with detecting the signal and running the `Clean()` routine.

MichaelMure:
Merged, thanks :+1:

PR #79, opened 2018-12-09
WIP: Initial Launchpad bridge.

Steap:


bridge/launchpad/import.go
```diff
@@ -0,0 +1,70 @@
...
+		Name:      b.Owner.Name,
+		Email:     "",
+		Login:     b.Owner.Login,
+		AvatarUrl: "",
+	}
+}
+
+func (li *launchpadImporter) ImportAll(repo *cache.RepoCache) error {
+	lpAPI := new(launchpadAPI)
+	lpAPI.Init()
```
golangcibot: Error return value of `lpAPI.Init` is not checked (from `errcheck`)

bridge/launchpad/import.go
```diff
@@ -0,0 +1,70 @@
...
+	for _, lpBug := range lpBugs {
+		lpBugID := fmt.Sprintf("%d", lpBug.ID)
+		_, err := repo.ResolveBugCreateMetadata(keyLaunchpadID, lpBugID)
+		if err != nil && err != bug.ErrBugNotExist {
+			return err
+		}
+
+		if err == bug.ErrBugNotExist {
+			createdAt, _ := time.Parse(time.RFC3339, lpBug.CreatedAt)
+			repo.NewBugRaw(
```
golangcibot: Error return value of `repo.NewBugRaw` is not checked (from `errcheck`)

bridge/launchpad/launchpad_api.go
```diff
@@ -0,0 +1,148 @@
...
+// Person describes a person on Launchpad (a bug owner, a message author, ...).
+type launchpadPerson struct {
+	Name  string `json:"display_name"`
+	Login string `json:"name"`
+}
+
+func (owner *launchpadPerson) UnmarshalJSON(data []byte) error {
+	type launchpadPersonX launchpadPerson // Avoid infinite recursion
+	var ownerLink string
+	json.Unmarshal(data, &ownerLink)
```
golangcibot: Error return value of `json.Unmarshal` is not checked (from `errcheck`)

bridge/launchpad/import.go
```diff
@@ -0,0 +1,70 @@
...
+				li.makePerson(lpBug),
+				createdAt.Unix(),
+				lpBug.Title,
+				lpBug.Description,
+				nil,
+				map[string]string{
+					keyLaunchpadID: lpBugID,
+				},
+			)
+		} else {
```
golangcibot: empty branch (from `megacheck`)

bridge/launchpad/launchpad_api.go
```diff
@@ -0,0 +1,148 @@
...
+		return nil
+	}
+
+	defer resp.Body.Close()
+
+	var p launchpadPersonX
+	if err := json.NewDecoder(resp.Body).Decode(&p); err != nil {
+		return nil
+	}
+	*owner = launchpadPerson{p.Name, p.Login}
```
golangcibot: should convert p (type launchpadPersonX) to launchpadPerson instead of using struct literal (from `megacheck`)

Steap:
So, this is not ready to merge, and requires a bit more work. I'm mostly willing to know if I'm headed into the right direction. You can test this code by doing:

```
$ ./git-bug bridge configure
# Configure a bridge using the "launchpad" target
```

Or just add this to your .git/config:
```
[git-bug "bridge.launchpad.test-lp"]
	project = python-glanceclient # Or whatever project you want :)
```

Then, as usual:

```
./git-bug bridge pull
./git-bug bridge termui
```

I think I would need to use bug.AddCommentRaw to add all comments posted on the bugs. I'm not really sure how I can add the status of the bug, and could not figure it out by reading the github bridge code.

MichaelMure:
That's awesome ! Thank you

Does the Lauchpad API provide some king of history of each bug ? It would be nice to have an incremental importer, able to support comment edition and such.

> I think I would need to use bug.AddCommentRaw to add all comments posted on the bugs.

Yes, most likely.

> I'm not really sure how I can add the status of the bug

The problem here is that git-bug only support two status currently, open and closed. Launchpad, according to https://help.launchpad.net/Bugs/Statuses support 10 of them. As you can see [here](https://framacalc.org/bugtracker), there is a lot of variability across bug trackers. Some support a fixed set, some support a configurable set, and of course they don't match each others ...

For now, this is an unsolved problem in git-bug ... So, what can we do ?

Option 1) map the Launchpad statuses to the two git-bug statuses, open and closed. Easy enough but some information is lost.
Option 2) import the Launchpad status anyway, and change the `Status.Validate()` function to not complain about the unknow status. UI will be broken as `unknown status` will be displayed. No information is lost, but it's not really helping and might create bugs elsewhere.
Option 3) actually solve the status problem properly in git-bug core and use that to implement the importer. Obviously the best outcome, but it's hard, and I'm not even sure what the best way would be to do that.

I would go with the Option 1) for now, and document properly in the importer what is missing, what information is lost, etc ... No need to do too much at a time, and even less in a single PR.

MichaelMure:
> I'm mostly willing to know if I'm headed into the right direction.

Yes

MichaelMure:
>  I would like to add a "last_pulled" key to the configuration, that would hold a timestamp. It would allow us to only query "new" changes from Launchpad, thus reducing the amount of requests sent.

That's very neat, but it's not what I was talking about. To make an importer that import everything, you need to have access to the history of a bug, not just the current state. As an example, Github expose a timeline of event (https://developer.github.com/v4/object/issue/#timeline) for an issue, a comment and so on. That means that the importer can go look for everything that happened in the past and import properly, for example, comment editions.

If launchpad doesn't provide that, that's fine. That means we will import a snapshot instead of all the history.

> Would this be an acceptable first patch?

Yes, very much. But in that case, maybe name the target `launchpad-preview` so users can expect that there is some missing pieces ?

bridge/launchpad/import.go
```diff
@@ -0,0 +1,80 @@
...
+				createdAt.Unix(),
+				lpBug.Title,
+				lpBug.Description,
+				nil,
+				map[string]string{
+					keyLaunchpadID: lpBugID,
+				},
+			)
+			if err != nil {
+				fmt.Println("Failed to add bug #", lpBugID)
```
MichaelMure: Some nitpicking ;)

- most likely, you want to stop the process and return the error here, because something went very wrong, so better stop before corrupting the local data somehow
- instead of manually piecing together you error message, you can use the `errors.wrap()` function,  which also have the side effect of keeping a stacktrace internally for easier debugging
- error message in go usually start with a lowercase letter, no formatting

so here that would be `return errors.Wrapf(err, "failed to add bug id #%s", lpBugID)`

bridge/launchpad/launchpad_api.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+
+	// First, try to gather info about the bug owner using our cache.
+	if cachedPerson, hasKey := personCache[ownerLink]; hasKey {
+		*owner = cachedPerson
+		return nil
+	}
+
+	// If the bug owner is not already known, we have to send a request.
+	req, err := http.NewRequest("GET", ownerLink, nil)
```
MichaelMure: It's a little odd to have an `UnmarshalJSON` function do a network request, as they are usually just about decoding the data. Not wrong though, no worries.

MichaelMure:
Keep up the good work :)

Btw, tell me when you want to merge the first batch.

bridge/launchpad/import.go
Steap: I agree that we should probably just stop if something goes wrong. I used errors.Wrapf in the new version.

bridge/launchpad/launchpad_api.go
Steap: Yeah, I was not sure how to do this. I could store the owner link in the Bug structure, and turn it into a Person structure later, but that is not super convenient.

Steap:
I'd be ok with merging this if it is good enough for you :)

MichaelMure:
Merged !

As you will see in https://github.com/MichaelMure/git-bug/commit/47b2aa4cc1c20ae37d846da4ea71e09c0c918192, I made a very minor cleaning (defering a Close() in a loop might lead to resources leak) and implemented the config validation.

PR #81, opened 2018-12-10
tests: configure user ident in repos used by tests

rafasc:
Some git operations require the user to have an identity configured and
will exit with failure if none is set (or if git can't guess it). As a
direct consequence of this, the test suite may fail depending on the
user local configuration.

The error itself is justified as regular users *should* configure their
identity themselves. However, when building in chrooted environments
it's unlikely the git identity will be set making the test suite fail
unnecessarily.

To prevent such unnecessary failures, let's make a dummy identity for
repos created and used by the test suite.

### DIFF
--- a/bug/bug_actions_test.go
+++ b/bug/bug_actions_test.go
@@ -31,6 +31,13 @@ func createRepo(bare bool) *repository.GitRepo {
 		log.Fatal(err)
 	}
 
+	if err := repo.StoreConfig("user.name", "testuser"); err != nil {
+		log.Fatal("failed to set user.name for test repository: ", err)
+	}
+	if err := repo.StoreConfig("user.email", "testuser@example.com"); err != nil {
+		log.Fatal("failed to set user.email for test repository: ", err)
+	}
+
 	return repo
 }
 
--- a/commands/select/select_test.go
+++ b/commands/select/select_test.go
@@ -96,6 +96,13 @@ func createRepo() *repository.GitRepo {
 		log.Fatal(err)
 	}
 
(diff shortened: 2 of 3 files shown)


bug/bug_actions_test.go
```diff
@@ -30,6 +30,8 @@ func createRepo(bare bool) *repository.GitRepo {
 	if err != nil {
 		log.Fatal(err)
 	}
+	repo.StoreConfig("user.name", "testuser")
```
golangcibot: Error return value of `repo.StoreConfig` is not checked (from `errcheck`)

bug/bug_actions_test.go
```diff
@@ -30,6 +30,8 @@ func createRepo(bare bool) *repository.GitRepo {
...
+	repo.StoreConfig("user.email", "testuser@example.com")
```
golangcibot: Error return value of `repo.StoreConfig` is not checked (from `errcheck`)

commands/select/select_test.go
```diff
@@ -95,6 +95,8 @@ func createRepo() *repository.GitRepo {
 	if err != nil {
 		log.Fatal(err)
 	}
+	repo.StoreConfig("user.name", "testuser")
```
golangcibot: Error return value of `repo.StoreConfig` is not checked (from `errcheck`)

MichaelMure:
As the golangbot said, it would be nice to check the error. You can either do a `if err != nil { log.Fatal(err) }` or pass the `t testing.TB` to the function and use `assert.NoError(t, repo.StoreConfig(...))`

PR #82, opened 2018-12-21
Launchpad bridge: fetch comments.

Steap:


bridge/launchpad/import.go
```diff
@@ -70,6 +73,26 @@ func (li *launchpadImporter) ImportAll(repo *cache.RepoCache) error {
...
+			err = b.AddCommentRaw(
+				li.makePerson(lpMessage.Owner),
+				createdAt.Unix(),
+				lpMessage.Content,
+				nil,
+				nil)
+			if err != nil {
+				return errors.Wrapf(err, "failed to add comment to bug #%s", lpBugID)
+			}
+			err = b.CommitAsNeeded()
```
Steap: So, I'm not sure why I need to "manually" commit the comments, when the bug itself is automagically committed. I would to get an explanation :)

Steap:
So, in my next patch, I'd like to refactor this bridge a bit. Most Launchpad API answers are paginated. Both SearchTasks() and queryMessages() handle pagination. I was thinking of having a single function handle the pagination, and then smaller functions do the actual processing. Here is an example of what it would look like:

http://paste.debian.net/1057015/


I am not too familiar with Golang though, so I'm kind of stuck because my paginate() function cannot return []LPBug in one case, and []LPMessage in another. Do you have any idea how I could handle pagination in a clean way?

bridge/launchpad/import.go
MichaelMure: It's the other way around. All new edit on a bug pile up in a staging area, kind of like git (see https://github.com/MichaelMure/git-bug/blob/master/bug/bug.go#L64). These changes are visible by any UI but are not yet written in git. Once it's time to write all that (this time is determined by each UI), a `Commit()` will write all edits in a single commit. `CommitAsNeeded()` is the same thing but only commit if there is actually something to commit so you don't get an error. All that is meant to limit the number of commit and improve the perfomance, be it reading speed or network transfer.

Now, the tricky thing is that the cache need to know the bug id to store a new bug. But a bug only get it's id when it's commited the first time (it's the hash of the first commit). That's why creating a new bug through the cache automatically commit it.

bridge/launchpad/import.go
```diff
@@ -70,6 +73,26 @@ func (li *launchpadImporter) ImportAll(repo *cache.RepoCache) error {
 			fmt.Println("TODO: Update bug")
 		}
 
+		/* Handle messages */
+		// The Launchpad API returns the bug description as the first
+		// comment, so skip it.
+		for _, lpMessage := range lpBug.Messages[1:] {
```
MichaelMure: I would add a check just before to make sure there is actually at least one element here. It would still be an import failure, but at least a polite failure ;)

bridge/launchpad/launchpad_api.go
```diff
@@ -175,5 +187,50 @@ func (lapi *launchpadAPI) queryBug(url string) (LPBug, error) {
...
+		var result launchpadMessageAnswer
+
+		err = json.NewDecoder(resp.Body).Decode(&result)
+		_ = resp.Body.Close()
+
+		if err != nil {
+			return nil, err
+		}
+
+		for _, messageEntry := range result.Entries {
```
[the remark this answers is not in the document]
MichaelMure: this bot is smart :heart:

MichaelMure:
Also, yeah for the PR :-D

bridge/launchpad/import.go
Steap: So I should commit as little as possible, right? In this case, I should probably only commit when returning from the function, am I right?

bridge/launchpad/import.go
Steap: I'm so focused on learning Go that I forgot about the basics :) Nice catch.

Steap:
> If you approach this problem as you would do with a language like, say, Java or C#, you will quickly end up in a dead end, as you can see. This is because go doesn't have generics and so you can't write a function that return one type and then another in another context.

Yeah, I was looking for something inheritance-related (like in Python) or macro-related (like in C), but I found out that generic types are not really something you want to do in Go.

> There is a few way you could go from here.

Yes, I think I may stick with option 2 (copy/paste) for now. This is probably the most readable solution.

bridge/launchpad/import.go
MichaelMure: That's correct. The github importer commit at the end of each individual bug import.

MichaelMure: Yes there is a way. Bugs, comment and everything you import should be tagged in the metadata with the corresponding launchpad id. That way, you can look in the bug edition for any operation with the id you want to import so see if it's already there or not.

in the github importer:
lookup --> https://github.com/MichaelMure/git-bug/blob/master/bridge/github/import.go#L390
tagging --> https://github.com/MichaelMure/git-bug/blob/master/bridge/github/import.go#L448

Steap: OK. Launchpad comments do not have their own ID, per say. I'll probably end up using "<bug-id>-<comment-index>" if this identifier is just used internally.

MichaelMure: You can use anything as long as its unique and stable. Maybe an url ? a made up id from the bug id and a comment index ? a hash of the content ?

MichaelMure: It's not used internally. Metadata are for your own usage, that's why they are prefixed with `launchpad-`

MichaelMure:
Hey @Steap, any chance you would continue working on this ?

Steap:
The new patch seems to be working. There is only one thing that I find a bit weird. Before applying this match, this is what termui would show:

gitbug-before

And after applying this patch:

gitbug-after

The "last edit" is updated, as expected. But now the bugs are not sorted by date of last edit any more (the order is the same before and after applying the patch). This might be a termui issue and not related to my patch, though. What do you think?

MichaelMure:
Awesome.

This behavior is by design: the default sorting in the termui is by creation time. This is to have a more stable list and allow a user to memorize the bug positions on screen or relative to each other, so they can browse quickly a long list.

You can change the query if you want though: https://github.com/MichaelMure/git-bug/blob/master/doc/queries.md

You can't change the default query yet though so it will reset each time you close the termui.

I'll review the code and merge asap but it looks good at first glance.

Steap:
Yeah, I tried using "sort:edit-desc" as a search string, and got this:

gitbug

This still does not look right. I'm not sure whether the bug is in my new code or not, though.

bridge/launchpad/import.go
```diff
@@ -70,6 +73,47 @@ func (li *launchpadImporter) ImportAll(repo *cache.RepoCache) error {
...
 
+		/* Handle messages */
+		if len(lpBug.Messages) == 0 {
+			return errors.Wrapf(err, "failed to fetch comments for bug #%s", lpBugID)
+		}
+
+		// The Launchpad API returns the bug description as the first
+		// comment, so skip it.
+		for i, lpMessage := range lpBug.Messages[1:] {
+			commentID := fmt.Sprintf("%d-%d", lpBug.ID, i)
```
MichaelMure: Are you sure that the list of comments you get from Launchpad is stable ? That is, can a comment be deleted and if yes, does that change the list ? Because, as you use through `i` the position of the comment in the response, that could break the mapping. Maybe use `self_link` that seems more stable (taken from https://launchpad.net/+apidoc/1.0.html#message, i suppose that the correct one) ?

bridge/launchpad/import.go
```diff
@@ -70,6 +73,47 @@ func (li *launchpadImporter) ImportAll(repo *cache.RepoCache) error {
...
+				createdAt.Unix(),
+				lpMessage.Content,
+				nil,
+				map[string]string{
+					keyLaunchpadID: commentID,
+				})
+			if err != nil {
+				return errors.Wrapf(err, "failed to add comment to bug #%s", lpBugID)
+			}
+			err = b.CommitAsNeeded()
```
MichaelMure: As discussed, the `CommitAsNeeded()` should probably be on the bug loop, not the comment loop, to reduce the overhead of commit, blob and all.

MichaelMure:
Ok, so what I think is happening about the bug order in the termui, is that the `LAST EDIT` column display the last edition based on the **timestamp**, while the list is ordered based on the  **Lamport time**.

That means that the ordering you see in the termui is correct: it shows in which order the bugs have been edited last. But that's regardless of the actual timestamp from Launchpad.

What you could do to fix that, is to query the bugs from launchpad ordered by edition. That way you would insert the changes with the same order.

bridge/launchpad/import.go
Steap: I did it.

Steap:
> What you could do to fix that, is to query the bugs from launchpad ordered by edition. That way you would insert the changes with the same order.

OK, I did that, and now I get an ordered list by default (oldest first) and can change the order by using "sort:edit-asc", which is nice!

MichaelMure:
:+1: :+1: :tada: :tada: 

Just FYI, I'm close to finishing https://github.com/MichaelMure/git-bug/pull/89/files#diff-6acca1a99ee701cba2dbae2e85b1dc92 which introduce a proper identity management. This changes impact this importer but I took care of it in my PR (although I have a conflict now ;) ). 

What's next ? labels ?

PR #83, opened 2018-12-22
Improve termui paging

Steap:


termui/bug_table.go
```diff
@@ -340,8 +340,13 @@ func (bt *bugTable) renderFooter(v *gocui.View, maxX int) {
 
 func (bt *bugTable) cursorDown(g *gocui.Gui, v *gocui.View) error {
 	_, y := v.Cursor()
-	y = minInt(y+1, bt.getTableLength()-1)
 
+	// If we are at the bottom of the page, switch to the next one.
+	if y+1 > bt.getTableLength()-1 {
```
MichaelMure: I see where it came from, but that's a weird test ;)
```suggestion
 	if y+1 >= bt.getTableLength() {
```
?

Steap:
> That gives a sort of natural scrolling flow. 

Yep, that is what I was going for. It really felt frustrating compared to mutt.

> Now, I'm less convinced by the reset of the cursor position when explicitly changing the page with left/right. I think left/right would be used mainly when the user already scrolled a few page and want to go back to a specific bug, or go from one bug to another. Key point here, is that the user already know visually _where_ the bug is, so where he wants to go. If the cursor is reset at each page change, it breaks this flow. You can't think "2 pages before, 3 lines up", because the cursor will move away as soon as you change the page.

I honestly do not think I would ever think "I want to go to page 3, line 17", but that might be because my brain does not work that way. Also, the order in which the bugs are shown will keep changing: users who sort them by date of last update will never get bugs in the same order, for instance.

This might be visually annoying  though. Maybe we could only set the selectCursor when pressing 'j' or 'k'? Would that work?

termui/bug_table.go
Steap: Yep, this might be more readable :)

MichaelMure:
> I honestly do not think I would ever think "I want to go to page 3, line 17", but that might be because my brain does not work that way. Also, the order in which the bugs are shown will keep changing: users who sort them by date of last update will never get bugs in the same order, for instance.

I didn't meant to say that a user would memorize the order and position of bugs day after day (although with an ordering by creation, some might), but rather remember the position of a bug they have seen 3 seconds before while browsing. I would be surprised if you don't do that as well, maybe unconsciously. 

As an example, let say you move files in a file browser. You see your target folder, go select the files you want to move, and you instinctively move them in the direction of the target folder. Even if it's not visible anymore. Because you know it's there. Does that ring a bell ?

If you don't mind, I'll take only 1174265. Hope that's ok.

Steap:
Oh, I see what you mean thanks to that file manager example. Yeah just take 1174265. I could change the second patch so that:

- pressing 'h' and 'l' does not change selectCursor;
- pressing 'j' and 'k' resets selectCursor.

What do you think?

Steap:
No idea you could push on my branch either :)

The behaviour seems nice. Maybe you could merge fb87d44888ff7d7e528156c6d4b7b508e17ddbae and 87098cee7b70679849e76ba608f28f1b94a5316e and then push this?

MichaelMure:
It's all in the branch and merged now :+1:

Issue #18, opened 2018-08-18
References bugs in commit messages


ianwalter:
Kind of like how it works in GitHub where you can references an issue # (or maybe short hash, but a number would be easier to remember) in a commit message and that reference displays a link in the issue timeline (and closes the issue if you put something like `Closes #1`?).

MichaelMure:
We can't have numbers (see https://github.com/MichaelMure/git-bug/blob/master/doc/model.md#you-cant-have-a-simple-consecutive-index-for-your-bugs) but we can use the shorter hash.

The tricky part is that git-bug is entirely decoupled from the normal code workflow (by design). Having this kind of feature would require git-bug to track changes in the normal branches and edit bugs accordingly.

Maybe git-bug could install a hook to spy on the normal workflow and trigger these changes ? The problem is that every single commiter of the repo would need to have git-bug installed or some of these events would be missed.

ianwalter:
Ah, of course, that makes sense. I'll dig more into the documentation.

Installing a commit hook sounds like a good idea to me. I'm okay with some of these events being missed if contributors don't have git-bug installed because I feel like even if teams don't buy into this fully, it would still be helpful to those on the team that are using it (especially if conveniences are added like the automatic bug closing feature I mentioned). 

This opens up the opportunity for a lot of features though and it's probably a good idea to think about an extension/plugin system early on so that the project is less burdensome to maintain. Personally, I am really excited about the idea of using a git-bug extension to generate a CHANGELOG because I think it would be a lot smoother than alternatives like commitizen.

edi9999:
This could be both a git hook to "notify" git-bug of the commit-message, and git-bug could also be automatically at startup check whether all commit messages have already been parsed, and then apply the "actions" starting from the oldest "unseen" commit

seeduvax:
I'm very interested in such feature.
First let me introduce myself a little bit. I'm working for an Aerospace Company and have 15 years background in industrial and critical software engineering. So to master changes with such software we have two strong requirement for the configuration management:
- for any source file, I shall always be able to recall for every change, why it was changed.
- for any change request (issue), I shell be able to retrieve all changes performed on all files. 

[...]

PR #84, opened 2018-12-22
Allow empty messages in comments.

Steap:
Some bug trackers, like Launchpad, allow messages to be empty (when adding a
file to the bug, for instance).

### DIFF
--- a/bug/op_add_comment.go
+++ b/bug/op_add_comment.go
@@ -58,10 +58,6 @@ func (op *AddCommentOperation) Validate() error {
 		return err
 	}
 
-	if text.Empty(op.Message) {
-		return fmt.Errorf("message is empty")
-	}
-
 	if !text.Safe(op.Message) {
 		return fmt.Errorf("message is not fully printable")
 	}
--- a/bug/operation_test.go
+++ b/bug/operation_test.go
@@ -47,7 +47,6 @@ func TestValidate(t *testing.T) {
 		NewSetTitleOp(rene, unix, "title", "multi\nline"),
 		NewSetTitleOp(rene, unix, "title\u001b", "title2"),
 		NewSetTitleOp(rene, unix, "title", "title2\u001b"),
-		NewAddCommentOp(rene, unix, "", nil),
 		NewAddCommentOp(rene, unix, "message\u001b", nil),
 		NewAddCommentOp(rene, unix, "message", []git.Hash{git.Hash("invalid")}),
 		NewSetStatusOp(rene, unix, 1000),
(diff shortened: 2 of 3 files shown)


MichaelMure:
Ha, I would have liked to keep this constraint, but I guess there is no way around it ...

What would be neat though, is that the data model allow empty comments (as you do with this PR), but each UI would still make sure to refuse to create empty message.

Could you check if your change doesn't create a weird bug in the command line tools and in the termui ? Including during the initial creation of a bug, when adding a comment, and editing a comment (empty or not). It should be working fine most of the time but I'm not sure it's 100% here.

Steap:
Well, I did not really think about that, but I think it might look kind of weird in termui. Maybe we could display "<git-bug: empty comment>" or something like that? Not sure whether that should be handled by the backends or by the core of git-bug though.

MichaelMure:
That's one thing i wanted to do. Something similar to what github do for the first message when empty in an issue (so just the title set), but adapted for a terminal. Maybe in italics or with a colored background ?

In any case, it should be handled by each UI. The core hold the data in a neutral way. This problem is a matter of presentation and should be dealt with by each UI, potentially in a different way.

MichaelMure:
A little help here, I implemented the placeholder thing for empty messages. 

image

:+1: or :-1:  ?

Steap:
I added a placeholder for the "show" command. What do you think?

MichaelMure:
I would just match the way it looks in the termui (seee https://github.com/MichaelMure/git-bug/blob/master/termui/show_bug.go#L381). But yeah, good idea :+1:

Steap:
I'm afraid that, because we do not use colors in the output, people might think that "no description provided" is the actual comment, which might be confusing.

MichaelMure:
There **is** colors in the `show` command currently (author, status, id). Color won't work everywhere though, but that's another question. In the future, I'd like to have a way to disable colors. No sure yet how though, it could be automatic by detecting if the output is a terminal or not, a `--no-color` flag, or follow the git config (`color.ui` key). Either way, a variable should be available to each command to know if colors should be used or not, and that will allow to render things differently.

Another thing I'd like to have is a flag to output the content in a scripting friendly way (see https://github.com/MichaelMure/git-bug/issues/18#issuecomment-447659419). If we have that, there is almost no need to disable colors anymore.

Steap:
Oh yeah, there are colors! The patch should be ok now :)

MichaelMure:
Alright, I failed to find a bug, so let's merge :-)

Issue #85, opened 2018-12-22
CJK contents does not display correctly in termui


ProgramFan:
When I add issues in Chinese, the characters are not displayed correctly in termui. For example, I get
<img width="939" alt="default" src="https://user-images.githubusercontent.com/32396/50376124-d05f6900-0643-11e9-937c-a270a4cfa12f.png">
for
<img width="939" alt="default" src="https://user-images.githubusercontent.com/32396/50376132-eb31dd80-0643-11e9-8595-f557ee346242.png">

ProgramFan:
This issue shall be caused by the inability of gocui to handle unicode texts correctly, as suggested in https://github.com/jroimartin/gocui/issues/157. The nice thing is that somone provides a solution in https://github.com/jroimartin/gocui/issues/157. Would it possible to fix this in go-bug?

MichaelMure:
I looked at this, and it's actually a problem in git-bug, not in gocui. Working on a fix.

MichaelMure:
Could you try the latest revision from git to see if it's fixed ?

ProgramFan:
Still not working, why?
<img width="939" alt="default" src="https://user-images.githubusercontent.com/32396/50400003-60c1b900-07be-11e9-96f3-750d51a2e9c0.png">

ProgramFan:
I found that the Chinese characters are skipped interleaved. Like the above example:

改**进** LibFind 支**持** Targets，**可**更**方**便**地**支**持**二**维**三**维**混**用**、**编**译**参**数**依**赖

The ones in bold are skipped. The English chars are kept.

MichaelMure:
Indeed, more fixes are needed. What happen is that the Chinese characters actually take **two** column in a terminal instead of always one as with the latin alphabet. It breaks an algorithm in git-bug, but it seems it also break gocui itself.

There is a PR with a fix in gocui (https://github.com/jroimartin/gocui/pull/173) but the project seems unmaintained for some month now.... That means I would need to fork gocui and maintain it ...

ProgramFan:
After looking into the PR, I step into the fork from lazygit authors: https://github.com/jesseduffield/gocui, which seems to have fixed the problems (also 28+ commits newer than original gocui). This fork is used in lazygit, and it works:
<img width="939" alt="default" src="https://user-images.githubusercontent.com/32396/50407791-289b9400-0819-11e9-9ff3-f565c4326064.png">

ProgramFan:
[...]

PR #86, opened 2018-12-26
Implement CJK support in termui

ProgramFan:
This PR implements CJK support in termui and addresses #85 
default
default


util/text/text.go
```diff
@@ -17,97 +16,99 @@ func Wrap(text string, lineWidth int) (string, int) {
...
+				width += wl + 1
 			}
 		}
+		word = ""
+	}
 
-		if lineBuffer.Len() > 0 {
-			textBuffer.WriteString(pad + strings.TrimRight(lineBuffer.String(), " "))
-			lineBuffer.Reset()
+	for _, r := range []rune(s) {
```
golangcibot: should range over string, not []rune(string) (from `megacheck`)

ProgramFan:
I am new to the go world, please kindly help me by pointing out any problems. I can refine them in following updates.

misc/zsh_completion/git-bug
```diff
@@ -17,6 +17,12 @@ case $state in
   ;;
   level2)
     case $words[2] in
+      status)
```
MichaelMure: Please don't commit that file. It's a generated file but the generator is not determinist (should be fixed now).

util/text/text.go
```diff
@@ -116,63 +117,13 @@ func wordLen(word string) int {
...
-
 		if char == 'm' {
 			escape = false
 		}
 	}
 
 	return length
 }
-
-// splitWord split a word at the given length, while ignoring the terminal escape sequences
```
MichaelMure: Even if it's not used anymore, let's keep that function (and the corresponding tests) around for now. It's working and might be useful in the future.

misc/zsh_completion/git-bug
ProgramFan: OK

ProgramFan:
I found it extremely difficult to pass the tests in `text_test.go`. Even for ascii-only tests, the rules of how the spaces are handled are extremely difficult to deduce. Is it acceptable to use the following simplified rune?
The text is wrapped as if it is first soft-wrapped and then have spaces trimmed with:
1. Lines consisting of only white spaces are trimmed, eg, just an empty line.
2. Trailing spaces are removed.
3. Leading spaces are preserved on the first line of a paragraph, and removed otherwise.

util/text/text.go
```diff
@@ -16,98 +15,199 @@ func Wrap(text string, lineWidth int) (string, int) {
...
+		return string(s)
+	}
+	// Assume the original string contains no new line and the wrapped only insert
+	// new lines. So that we can recover the position where we insert the term
+	// escapes.
+	var out string = ""
+
+	currPos := 0
+	currItem := 0
+	for _, r := range []rune(s) {
```
golangcibot: should range over string, not []rune(string) (from `megacheck`)

util/text/text.go
```diff
@@ -16,98 +15,199 @@ func Wrap(text string, lineWidth int) (string, int) {
...
 		}
+	}
 
-		if lineBuffer.Len() > 0 {
-			textBuffer.WriteString(pad + strings.TrimRight(lineBuffer.String(), " "))
-			lineBuffer.Reset()
+	var chunks []string
+	var word string
+	wordType := NONE
+	for _, r := range []rune(newStr) {
```
golangcibot: should range over string, not []rune(string) (from `megacheck`)

ProgramFan:
I adapted two tests in `text_test.go` so that they conform to the proposed wrapping algorithm. Two extra tests for mixed wide-normal words are added.

I will refactor the `text.go` and add tests for extra functions. Once they are ready, this PR is ready to merge.

ProgramFan:
I am ready for the merge. @MichaelMure

util/text/left_padded_test.go
```diff
@@ -28,13 +28,13 @@ func TestLeftPadMaxLine(t *testing.T) {
 		},
 		{
 			"foo",
-			"  f…",
+			"  ...",
```
MichaelMure: Please keep using the `…` character, it allows to save space and have more useful characters displayed. It's very helpful on a small terminal area.

ProgramFan: The problem here is that '…' is counted two cells but displayed as '…' + VOID in the terminal. This makes the background of the selected item in bug table missing one char:
default

I guess this can be resolved in gocui? If it can be fixed there, I am very glad to change it back to unicode ellipsis.

ProgramFan: The problem here is that '…' is counted two cells but displayed as '…' + VOID in the terminal. This makes the background of the selected item in bug table missing one char:
default

I guess this can be resolved in gocui? If it can be fixed there, I am very glad to change it back to unicode ellipsis.

ProgramFan:
Indeed, we need to change the wrapping algorithm bottom-up to make it suitable for CJK contents.

The original algorithm breaks lines at spaces, and breaks long words in the middle, which emulates the behavior of vim `set formatoptions-=Mm`. This algorithm will leave some lines very short, like:
default
But we really need to emulate `set formatoptions+=Mm` for CJK contents, like:
default

Given that the original is tightly coupled with the formal assumption, incremental changes will simply not do. This is why the change is to introduce a new algorithm.

ProgramFan:
The new algorithm preserves almost all behaviors of the original one, so for English documents, the results are the same (almost). As you see, the algorithm passes all but two tests. The two failing tests are concerned with `\t`, where the original algorithm inserts extra newlines while the new algorithm simply treats them as four spaces. The new algorithm produces the same texts as vim (while the original does not), so I would prefer the new algorithm.

ProgramFan:
The new algorithm uses a quite different logic: it first softwarps the line, then strips out heading and trailing spaces. The softwarp is done by firstly segmenting the line into chunks, then inserting linebreaks among these chunks.

The fact that the text contains terminal escapes makes the softwrap difficult, as they can appear anywhere in the line and shall be treated in pairs to ensure the space-trimming works properly. I tried many ways to deal with it directly in the softwrap+trimming process, and none succeeded. My final choice is to extract out the terminal escapes before softwrap and insert them back after softwrap. This algorithm passes all tests, including new tests I added.

ProgramFan:
The new algorithm is not Chinese-specific. It shall work with all CJK contents.

util/text/left_padded_test.go
ProgramFan: Luckily runewidth exports some variables so we can workaround this in our code.

util/text/text.go
```diff
@@ -1,12 +1,18 @@
...
-
 	"github.com/mattn/go-runewidth"
+	"strings"
+	"unicode/utf8"
 )
 
+// Force runewidth not to treat ambiguous runes as wide chars, so that things
+// like unicode ellipsis/up/down/left/right glyphs can have correct runewidth
+// and can be displayed correctly in terminals.
+func init() {
```
MichaelMure: `util/text/left_padded.go` and `util/text/text.go` are in the same package (`text`) so a single `init()` will be enough.

ProgramFan: Thanks for pointing this out and helpping me understand it. I am fixing it.

util/text/text.go
```diff
@@ -16,98 +22,241 @@ func Wrap(text string, lineWidth int) (string, int) {
...
-	return textBuffer.String(), nbLine
+	return chunks
+}
+
+// Rune categories
+//
+// These categories are so defined that each category forms a non-breakable
+// chunk. It IS NOT the same as unicode code point categories.
+//
+const (
```
MichaelMure: There is a few things that could be improved here:
- These are useful only inside the package, so they should not be exported
- CamelCase is generally used in go
- iota could be better used
- an explicit type could be used

It could be something like this:

```go
const (
	none int = -1 + iota
	wideChar
	invisible
	shortUnicode
	space
	visibleAscii
)
```

You probably don't need the `-1` but I'll let you decide.

util/text/text_test.go
```diff
@@ -43,7 +44,7 @@ func TestWrap(t *testing.T) {
 		// A tab counts as 4 characters.
 		{
 			"foo\nb\t r\n baz",
-			"foo\nb\n  r\n baz",
+			"foo\nb\nr\n baz",
```
MichaelMure: The idea here was that a `\t` character is an explicit 4 spaces and should probably not be trimmed or reduced to a single space. In other words, the algorithm should trim spaces only after the wrapping as been done.

MichaelMure: If I had to choose I would keep the previous behavior here (the tabulation spacing is preserved across lines), but your behavior is not really wrong either. If you can make it works, that's good, but it's fine otherwise.

util/text/text.go
```diff
@@ -16,98 +22,241 @@ func Wrap(text string, lineWidth int) (string, int) {
...
+}
 
-			if !firstWord {
-				lineBuffer.WriteString(" ")
-				spaceLeft -= 1
+// EscapeItem: Storage of terminal escapes in a line. 'item' is the actural
+// escape command, and 'pos' is the index in the rune array where the 'item'
+// shall be inserted back. For example, the escape item in "F\x1b33mox" is
+// {"\x1b33m", 1}.
+type EscapeItem struct {
```
MichaelMure: It's only used internally, it should not be exported.

ProgramFan: Thanks, I will resolve this.

util/text/text.go
ProgramFan: Thanks for the suggestions. I will improve this.

util/text/text_test.go
ProgramFan: I can try to preserve the previous behavior. But I don't quite understand the rule here. `b\t r\n` results in `b\n  r\n`, which means the 4 spaces are trimmed in a specific way.

ProgramFan:
It looks the word is correctly segmented, but the padding gets inserted in the wrong place. This shall be a bug in the new algorithm. I will fix it.

util/text/text_test.go
MichaelMure: The idea is that the tabulation is replaced by a 4 character long space and keep providing a constant spacing. I didn't try with your algorithm, but my guess is that if the tabulation is at the end of the line, it will be trimmed and disappear. The previous behavior is that the next line would be pushed with an extra padding to keep this spacing.

ProgramFan:
I identified the bug. The softwarping logic used a wrong width to split words. Fixed in 8ea5c6b41de.

util/text/text_test.go
ProgramFan: Sorry for not getting the idea. Why 'b\tSr\n' results in `b\nSSr\n` given a 4-char limit? Shouldn't it be `b\nSSSS\nSr\n`?

ProgramFan: Another thing I failed to understand is the line https://github.com/MichaelMure/git-bug/blob/433c4fc46a7a2cabb69eee5226bb8ea87b6c5661/util/text/text.go#L26, which substitute tabs for 4 spaces, but only for the first 4 occurrences? But if tabs are substituted before processing, how do we know when we shall treat the tabs in a special way?

MichaelMure: Ho well, you are right, the previous code was buggy. 

> Sorry for not getting the idea. Why 'b\tSr\n' results in `b\nSSr\n` given a 4-char limit? Shouldn't it be `b\nSSSS\nSr\n`?

You are correct. That said, this behavior is more a wish than a requirement. I think that would look better, but if you can't make it work, it's fine.

MichaelMure:
I'll have a quick test but that looks ready to be merged !

MichaelMure:
It works better than before, so I merged.

Thank you for your tenacity :-)

PR #88, opened 2019-01-02
show command - id switch and use example in hook script/

seeduvax:
Added --id switch to the show command to only get the selected bug id. This is introduced to ease the implementation of a hook script to pre-fill the commit message with the current bug id as hashtag. Getting only the id prevent to use acrobatic pipes of sed that may suffer change in the regular text ouptut format.
The prepare-commit-msg git hook script I use with this hack for my own projects is provided in the new contrib directory included in this pull request.

### DIFF
--- a/commands/show.go
+++ b/commands/show.go
@@ -12,6 +12,9 @@ import (
 	"github.com/spf13/cobra"
 )
 
+var (
+	showId	bool
+)
 func runShowBug(cmd *cobra.Command, args []string) error {
 	backend, err := cache.NewRepoCache(repo)
 	if err != nil {
@@ -27,6 +30,10 @@ func runShowBug(cmd *cobra.Command, args []string) error {
 
 	snapshot := b.Snapshot()
 
+	if showId {
+		fmt.Printf("%s\n",snapshot.HumanId())
+		return nil
+	}
 	if len(snapshot.Comments) == 0 {
 		return errors.New("Invalid bug: no comment")
 	}
@@ -90,4 +97,5 @@ var showCmd = &cobra.Command{
 
 func init() {
 	RootCmd.AddCommand(showCmd)
+	showCmd.Flags().BoolVarP(&showId,"id","i",false,"Select field to display: id")
 }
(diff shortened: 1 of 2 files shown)


seeduvax:
I planned to make a 2nd pull request for commit 5dac6fb, but it's there. I'm still discovering github features. So feel free to reject if you don't want to mix the two topics in a single request.

j-f1:
> I planned to make a 2nd pull request for commit 5dac6fb, but it's there.

The typical way to prevent that is to use a separate branch for each PR (this also has the bonus that GitHub will give you the option to delete the branch when the PR is merged so it doesn’t clutter up your repo).

commands/show.go
```diff
@@ -90,4 +97,5 @@ var showCmd = &cobra.Command{
 
 func init() {
 	RootCmd.AddCommand(showCmd)
+	showCmd.Flags().BoolVarP(&showId,"id","i",false,"Select field to display: id")
```
MichaelMure: While I completely agree that it's a good feature to have (I had that in mind already), I'm not a fan of the current UX:

- `id` refers to the full 40 character long identifier. "Human ID" is more an internal naming from the geek world, so let's call that "short id"
- it should be open to more field than just the id. There is at least
  - id
  - short id
  - status
  - title
  - last comment
  - last edit time (regular time)
  - last edit time (lamport time)
  - labels (one by line)
  - author's name
  - author's email

So ...

What about a unique `--field=` to select what field to display ? This could very well combine with a future `--enc=` flage to select what kind of output we want (normal, json, ...). In the future, a multi valued `--fields=` could be added to select many fields at the same time as well.

There is an example of how to deal with these flags in the `ls` command.

MichaelMure: Yeah, leave the json thing for later, it needs extensive plumbing and a battle plan on its own.

I wrote an overview of the data model here: https://github.com/MichaelMure/git-bug/blob/master/doc/model.md

There is also the comment, either directly in the code or through https://godoc.org/github.com/MichaelMure/git-bug

But in short:

- `bug.Bug` hold the raw data, that is the list of edit operations for a bug
- `bug.Snapshot` hold a compiled version for the various UI to use
- `bug.WithSnapshot` is a special purpose `Bug` that maintain a `Snapshot` up to date incrementally, each time a new operation is added, for performance
- `bug.Interface` is an interface to manipulate `Bug` or `WithSnapshot` transparently

Issue #87, opened 2019-01-02
bug identifier size vary across commands


seeduvax:
ls command gives id with 7 char while show command gives Id with 8 char.
After quick code review, it appear the HumanId method implementation is not consistent in bug.go and snapshot.go. First use a constant that is defined in the same file. Second is hard coded in the Sprintf call.
A first trivial fix (very easy I may send a pull request today) is to change the hard coded value in snapshot.go (7 char seems to be the right value to match the 7 char ids generally provided in git and github). A better fix should be to use the same constant. To be defined is the way to manage such constant: into a dedicated go file or to import bug.go in snapshot.go.
To be checked: is there other alternative HumanId() functions hidden elsewhere?

MichaelMure:
Fixed by #90

PR #90, opened 2019-01-04
Issue #87: use same id format function in snapshot.go than in bug.go

seeduvax:
After some fight with my git, here is the best fix I can actually do for the short id format issue. May be some there are aother alternate HumanID() implementation somewhere. I didn't find any by doing basic search with grep.


MichaelMure:
Perfect :+1:

PR #91, opened 2019-01-04
Cmd show fields

seeduvax:
Introduced --fields switch to select what to be displayed by show command.
Also added use example in misc/git_hooks as previously suggested.

### DIFF
--- a/commands/show.go
+++ b/commands/show.go
@@ -12,6 +12,10 @@ import (
 	"github.com/spf13/cobra"
 )
 
+var (
+	showFieldsQuery	[]string
+)
+
 func runShowBug(cmd *cobra.Command, args []string) error {
 	backend, err := cache.NewRepoCache(repo)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


commands/show.go
```diff
@@ -33,49 +37,75 @@ func runShowBug(cmd *cobra.Command, args []string) error {
...
+				case "status": fmt.Printf("%s ",snapshot.Status)
+				case "title": fmt.Printf("%s ",snapshot.Title)
+				default:
+					unknownFields+=field+" "
+					err=true
+			}
+		}
+		fmt.Printf("\n")
+		if err {
+			return errors.New(fmt.Sprintf("Unsupported fields requested: %s\n",unknownFields))
```
golangcibot: should use fmt.Errorf(...) instead of errors.New(fmt.Sprintf(...)) (from `megacheck`)

commands/show.go
```diff
@@ -33,49 +37,75 @@ func runShowBug(cmd *cobra.Command, args []string) error {
...
+			}
+
+			fmt.Printf("%s%s\n\n\n",
+				indent,
+				message,
+			)
+		}
+	} else {
+		unknownFields:=""
+		err:=false
```
MichaelMure: This very not like go. In go, an `err` variable is very much expected to be of type `error`, not a boolean. I think you can simply return an error as soon as you find a malformed query (the `default` case).

MichaelMure:
We are close :) Just fix the weirdish error handling and I'll merge. Thanks !

commands/show.go
seeduvax: Made it shorter, then no more need of this ambiguous err flag.

MichaelMure:
Hmm, I used the feature, but when selecting multiple field at the same time, they are written in the same line, which is probably unusable most of the time.

What about sticking with a single field for now (`--field`) ? The flag would be of type `StringVarP`.

seeduvax:
It is qui te useable while you are requesting fields that are single Word (like statuts or id). But you're right, if you fetch user and title it's unclear  where fields stars and stop. I see two options:
- use à specific separator char like ';' (then output May be processed by anything that processed csv files), but the extracted fields could also contain the separator char. Escaping will be needed... 
- output each field on a new line. Surely the easiest to implement: change one char and remove one line !

MichaelMure:
@seeduvax I quickly made the change. What do you think ?

> output each field on a new line

That's not perfect either, because you will have some field like a comment that will already be multi line.

I think it's best to stick with a single field for each call. Multiple values will be useful when we can output for example JSON.

MichaelMure:
Alright, thanks for the help :+1: :-)

If you want to tackle something else, let me know :-)

seeduvax:
@MichaelMure Next for me will be probably to draft something about relationships between issues as I suggested on my first message to you. It will probably take some weeks before I'll come back with something consistent to be discussed.

MichaelMure:
Alright, neat :-)

Although, please come forward early in this process, either on a github issue or in [Gitter](https://gitter.im/Arbore/Lobby) so we can discuss that. It's a tricky subject, both technically and for the UX. I'll have to make sure that it fits nicely with other features and workflows.

Issue #93, opened 2019-01-20
feat: easier Github bridge configuration


MichaelMure:
So I was thinking.

The first step of the Github bridge configurator is to ask the user for the project URL (ex: `https://github.com/MichaelMure/git-bug`). But the configurator could be smart and look for the configured remote to provide a list the user could pick from.

```
$ git remote --verbose
origin	git@github.com:MichaelMure/git-bug.git (fetch)
origin	git@github.com:MichaelMure/git-bug.git (push)
adamslc	git@github.com:adamslc/git-bug.git (fetch)
adamslc	git@github.com:adamslc/git-bug.git (push)
```

Would give:

```
Detected projects:
[1]: github.com/MichaelMure/git-bug
[2]: github.com/adamslc/git-bug

[0]: another project
```

If [0] is selected, the current prompt (`Github project URL: `) is shown.

sladyn98:
@MichaelMure we could use the 
``` os/exec``` package to do the job
```out, err := exec.Command("git remote --verbose").Output()```
	```if err != nil {```
		```fmt.Printf("%s", err)```
	```}```
	```output := string(out[:])```
	```fmt.Println(output)```
we could then pick pipe the user input to a buffer and take the necessary action.

MichaelMure:
You could, but please implement a generic function `ListRemotes` in the interface `RepoCommon` and the type that implement it. You can then use this new function from the `Configure(repo repository.Repo)`.

corobo-sladyn:
I added the function declaration to ```RepoCommon```interface in  ```repo.go``` and the the implementation in the ```git.go``` file.Does the the function have to be called  in the ```config.go``` of ```launchpad``` or ```github``` .

sladyn98:
I added the function declaration to RepoCommoninterface in repo.go and the the implementation in the git.go file.Does the the function have to be called in the config.go of launchpad or github .

MichaelMure:
Lets start with github. I suppose it could be done for launpad as well but i don't know launchpad enough to tell.

sladyn98:
I added the new function in config.go for github. How do I test that the new features work ? Is there a custom development script or do I have to compile the binaries.

MichaelMure:
Dealt with by https://github.com/MichaelMure/git-bug/pull/153

PR #95, opened 2019-01-29
easier Github bridge configuration

sladyn98:
Added a List Remotes function which fetches the local remotes configured making it easier to configure a bridge.

Fixes: https://github.com/MichaelMure/git-bug/issues/93

### DIFF
--- a/bridge/github/config.go
+++ b/bridge/github/config.go
@@ -11,6 +11,7 @@ import (
 	"net/http"
 	"os"
 	"regexp"
+	"strconv"
 	"strings"
 	"syscall"
 	"time"
@@ -36,7 +37,7 @@ func (*Github) Configure(repo repository.RepoCommon) (core.Configuration, error)
 	// fmt.Println("The token will have the \"repo\" permission, giving it read/write access to your repositories and issues. There is no narrower scope available, sorry :-|")
 	fmt.Println()
 
-	projectUser, projectName, err := promptURL()
+	projectUser, projectName, err := promptURL(repo)
 	if err != nil {
 		return nil, err
 	}
@@ -205,13 +206,69 @@ func promptUsername() (string, error) {
 	}
 }
 
-func promptURL() (string, string, error) {
+func promptURL(repo repository.RepoCommon) (string, string, error) {
 	for {
-		fmt.Print("Github project URL: ")
 
-		line, err := bufio.NewReader(os.Stdin).ReadString('\n')
+		fmt.Println("Detected Projects:")
+
+		repos, err := repo.ListRemotes()
+
 		if err != nil {
 			return "", "", err
+		}
+
+		reposMap := make(map[int]string)
+		i := 1
+		for _, value := range repos {
+
+			if checkGithubURL(value) {
+				shortenedURL, err := shortenedURL(value)
+				if err != nil {
+					fmt.Println(err)
+					continue
+				}
+				fmt.Printf("[%d]: %s\n", i, shortenedURL)
+				reposMap[i] = value
+				i++
+			}
+		}
+
+		if len(reposMap) == 0 {
+			fmt.Printf("No github projects found\n")
+		}
+
+		fmt.Printf("\n[0]: Another project\n")
+		fmt.Println()
+		output, err := bufio.NewReader(os.Stdin).ReadString('\n')
+		if err != nil {
+			fmt.Println(err)
+			return "", "", fmt.Errorf("input error")
+		}
+		output = strings.TrimRight(output, "\n")
+
+		outputInteger, err := strconv.Atoi(output)
+		var line string
+		if err != nil {
+			fmt.Println(err)
+			return "", "", fmt.Errorf("error in string conversion")
+		}
+
+		if outputInteger == 0 {
+			fmt.Println("Enter Project URL:")
+			line, err = bufio.NewReader(os.Stdin).ReadString('\n')
+			if err != nil {
+				return "", "", fmt.Errorf("input error")
+			}
+			line = strings.TrimRight(line, "\n")
+
+			if !checkGithubURL(line) {
+				fmt.Printf("Please Enter valid github URL\n")
+				fmt.Println()
+				continue
+			}
+
+		} else {
+			line = reposMap[outputInteger]
 		}
 
 		line = strings.TrimRight(line, "\n")
@@ -232,8 +289,20 @@ func promptURL() (string, string, error) {
 	}
 }
 
+func checkGithubURL(url string) bool {
+
+	re, err := regexp.Compile(`^[a-z]*\W*github\.com[a-z]*.*$`)
+
+	if err != nil {
+		panic(err)
+	}
+	res := re.FindStringSubmatch(url)
+	return !(len(res) == 0)
+
+}
+
 func splitURL(url string) (string, string, error) {
-	re, err := regexp.Compile(`github\.com\/([^\/]*)\/([^\/]*)`)
+	re, err := regexp.Compile(`github\.com[\/:]([^\/]*)\/([^\/]*)`)
 	if err != nil {
 		panic(err)
 	}
@@ -245,6 +314,23 @@ func splitURL(url string) (string, string, error) {
 	}
 
 	return res[1], res[2], nil
+}
+
+func shortenedURL(url string) (string, error) {
+
+	re, err := regexp.Compile(`github\.com[\/:][^\/]*\/[^\/]*`)
+	if err != nil {
+		panic(err)
+	}
+
+	res := re.FindStringSubmatch(url)
+
+	if res == nil {
+		return "", fmt.Errorf("bad github project url")
+	}
+
+	return res[0], nil
+
 }
 
 func validateUsername(username string) (bool, error) {
--- a/cache/repo_cache.go
+++ b/cache/repo_cache.go
@@ -130,6 +130,11 @@ func (c *RepoCache) RmConfigs(keyPrefix string) error {
 	return c.repo.RmConfigs(keyPrefix)
 }
 
+// ListRemotes lists all the local remotes
+func (c *RepoCache) ListRemotes() (map[string]string, error) {
+	return c.repo.ListRemotes()
+}
+
 func (c *RepoCache) lock() error {
 	lockPath := repoLockFilePath(c.repo)
 
--- a/repository/git.go
+++ b/repository/git.go
@@ -209,6 +209,34 @@ func (repo *GitRepo) RmConfigs(keyPrefix string) error {
 	return err
 }
 
+//ListRemotes lists all the local remotes
+func (repo *GitRepo) ListRemotes() (map[string]string, error) {
+	out, err := repo.runGitCommand("remote", "--verbose")
+
+	if err != nil {
+		return nil, err
+	}
+
+	splitRemotes := strings.Split(out, "\n")
+
+	remotesMap := make(map[string]string, len(splitRemotes))
+
+	for _, line := range splitRemotes {
+
+		remoteValues := strings.Fields(line)
+		if len(remoteValues) != 3 {
+			return nil, fmt.Errorf("error in parsing remote: %s", line)
+		}
+
+		if _, ok := remotesMap[remoteValues[0]]; ok {
+			continue
+		}
+		remotesMap[remoteValues[0]] = remoteValues[1]
+
+	}
+	return remotesMap, nil
+}
+
 // FetchRefs fetch git refs from a remote
 func (repo *GitRepo) FetchRefs(remote, refSpec string) (string, error) {
 	stdout, err := repo.runGitCommand("fetch", remote, refSpec)
--- a/repository/mock_repo.go
+++ b/repository/mock_repo.go
@@ -84,6 +84,15 @@ func (r *mockRepoForTest) RmConfigs(keyPrefix string) error {
 	return nil
 }
 
+func (r *mockRepoForTest) ListRemotes() (map[string]string, error) {
+
+	remotesMap := make(map[string]string)
+	for key, value := range r.config {
+		remotesMap[key] = value
+	}
+	return remotesMap, nil
+}
+
 // PushRefs push git refs to a remote
 func (r *mockRepoForTest) PushRefs(remote string, refSpec string) (string, error) {
 	return "", nil
--- a/repository/repo.go
+++ b/repository/repo.go
@@ -31,6 +31,9 @@ type RepoCommon interface {
 
 	// RmConfigs remove all key/value pair matching the key prefix
 	RmConfigs(keyPrefix string) error
+
+	//ListRemotes lists all the remotes configured in the current repository
+	ListRemotes() (map[string]string, error)
 }
 
 // Repo represents a source code repository.


repository/git.go
```diff
@@ -207,6 +207,18 @@ func (repo *GitRepo) RmConfigs(keyPrefix string) error {
 	return err
 }
 
+//ListRemotes lists all the local remotes
+func (repo *GitRepo) ListRemotes() ([]string, error) {
```
MichaelMure: This function should be generic, usable for other purpose than just what you want it to do right now. That means in particular:

- you need to deduplicate the remotes (`git remote --verbose` generally list twice each remotes, for fetch and push). If they are identical, you should list them only once.
- you should do more processing of the raw output of the command here, not just split each lines 
- the name of the remote is important as well, instead of returning a `[]string` you could return a map with the name of the remote as the key, and the URL as the value

repository/mock_repo.go
```diff
@@ -82,6 +82,11 @@ func (r *mockRepoForTest) RmConfigs(keyPrefix string) error {
 	return nil
 }
 
+func (r *mockRepoForTest) ListRemotes() ([]string, error) {
+	testRemotes := make([]string, 2)
```
MichaelMure: A mock (see https://stackoverflow.com/questions/214092/what-is-a-mock-and-when-should-you-use-it) should return correct values, and an array of two empty string is not a correct return value for `ListRemotes`

bridge/github/config.go
```diff
@@ -205,11 +205,32 @@ func promptUsername() (string, error) {
...
+
+		var choice int
+		_, err := fmt.Scanf("%d", &choice)
+
+		var line string
+		if choice == 0 {
+			fmt.Print("Github project URL: ")
+			line, err = bufio.NewReader(os.Stdin).ReadString('\n')
+		} else {
+			line = listOfRemotes[choice-1]
```
MichaelMure: This is not working (you should run the project and see if it works). Here, you need to extract the `projectName` and `projectUser` from the remote. Maybe do something similar as `splitURL` with a regex ?

bridge/github/config.go
```diff
@@ -205,13 +205,34 @@ func promptUsername() (string, error) {
...
+			i++
+		}
+
+		fmt.Printf("[0]: Another project\n")
+		fmt.Println("Enter project fork eg:origin,myfork")
+		fmt.Println("For custom url enter custom:")
+		fmt.Println()
+		var line, output string
+		var err error
+		output, err = bufio.NewReader(os.Stdin).ReadString('\n')
```
golangcibot: this value of `err` is never used (from `staticcheck`)

bridge/github/config.go
```diff
@@ -205,13 +205,34 @@ func promptUsername() (string, error) {
...
+		output = strings.TrimRight(output, "\n")
+
+		if repos[output] != "" {
+			line = repos[output]
+			fmt.Printf("Url chosen is %s\n", repos[output])
+		} else {
+			fmt.Println("Enter Project URL:")
+			line, err = bufio.NewReader(os.Stdin).ReadString('\n')
```
golangcibot: this value of `err` is never used (from `staticcheck`)

repository/git.go
sladyn98: Yeah sure and instead of entering a number we could promt the user to enter the branch name eg myfork or origin.

bridge/github/config.go
```diff
@@ -205,13 +205,42 @@ func promptUsername() (string, error) {
...
+		fmt.Println("For custom url enter custom:")
+		fmt.Println()
+		var line string
+		output, err := bufio.NewReader(os.Stdin).ReadString('\n')
+		output = strings.TrimRight(output, "\n")
+
 		if err != nil {
-			return "", "", err
+			fmt.Println(err)
+			continue
```
MichaelMure: this is a critical error, there is no point in trying again.

bridge/github/config.go
```diff
@@ -205,13 +205,42 @@ func promptUsername() (string, error) {
...
 	for {
-		fmt.Print("Github project URL: ")
 
-		line, err := bufio.NewReader(os.Stdin).ReadString('\n')
+		fmt.Println("The remotes are:")
+
+		repos, nil := repo.ListRemotes()
+
+		i := 0
+		for key, value := range repos {
```
MichaelMure: you need to filter the remotes and suggest only the one that can work here, that is, the remotes from github

bridge/github/config.go
```diff
@@ -205,13 +205,42 @@ func promptUsername() (string, error) {
 	}
 }
 
-func promptURL() (string, string, error) {
+func promptURL(repo repository.RepoCommon) (string, string, error) {
```
MichaelMure: Could you try to match the behavior I detailed in https://github.com/MichaelMure/git-bug/issues/93#issue-401117315 ?

In particular, the user should be able to select an option with just the number, and the `[0]` should allow to enter an arbitrary project, as the code does before your PR.

repository/mock_repo.go
```diff
@@ -82,6 +82,17 @@ func (r *mockRepoForTest) RmConfigs(keyPrefix string) error {
 	return nil
 }
 
+func (r *mockRepoForTest) ListRemotes() (map[string]string, error) {
+	remotes := make(map[string]string)
+	for key, val := range r.config {
+		if _, ok := remotes[key]; ok {
```
MichaelMure: this would match much more than just the remotes. Have a look at the output of `git config --list`.

bridge/github/config.go
MichaelMure: It's not important how much code we need to write. What is important is to build a tool that is fast, easy to use and understand.

sladyn98: Okay wonderful I will map the index values to the URL and submit a PR.

bridge/github/config.go
```diff
@@ -232,6 +277,23 @@ func promptURL() (string, string, error) {
...
+
+	re, err := regexp.Compile(`^https?:\/\/github.com\/*`)
+
+	if err != nil {
+		fmt.Printf("we are entering hree")
+		panic(err)
+	}
+	res := re.FindStringSubmatch(url)
+
+	if len(res) == 0 {
```
golangcibot: should use 'return <expr>' instead of 'if <expr> { return <bool> }; return <bool>' (from `gosimple`)

bridge/github/config.go
```diff
@@ -205,13 +206,57 @@ func promptUsername() (string, error) {
...
+
+		repos, err := repo.ListRemotes()
+		reposMap := make(map[string]string)
+
+		if err != nil {
+			return "", "", fmt.Errorf("Error in fetching remotes")
+		}
+
+		i := 0
+		for _, value := range repos {
```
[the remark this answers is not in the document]
sladyn98: Yeah I know I tried that to make the code readable but that's a string which I realised converting and using it as an integer required extra code, but I think it should be enhanced for readability. I will submit a ammendment.

cache/repo_cache.go
```diff
@@ -90,6 +90,11 @@ func (c *RepoCache) RmConfigs(keyPrefix string) error {
 	return c.repo.RmConfigs(keyPrefix)
 }
 
+//ListRemotes lists all the local remotes
```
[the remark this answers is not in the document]
sladyn98: Yeah Implemented :+1:

repository/mock_repo.go
```diff
@@ -82,6 +82,17 @@ func (r *mockRepoForTest) RmConfigs(keyPrefix string) error {
 	return nil
 }
 
+func (r *mockRepoForTest) ListRemotes() (map[string]string, error) {
+	remotesMap := make(map[string]string)
+
+	for key, value := range r.config {
```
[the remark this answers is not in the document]
sladyn98: Yeah since I changed the function to take the remote names as keys..... I will change this as well :+1:

MichaelMure:
Tested it, it's not working with ssh-type remotes:

```
The remotes are:
[1]: git@github.com:seeduvax/git-bug.git
[2]: git@github.com:vandesign/git-bug.git
[3]: git@github.com:adamslc/git-bug.git
[4]: git@github.com:JonLundy/git-bug.git
[5]: git@github.com:MichaelMure/git-bug.git
[6]: git@github.com:Steap/git-bug.git
[7]: git@github.com:ProgramFan/git-bug.git
[8]: git@github.com:sladyn98/git-bug.git
[0]: Another project

2
Url chosen is git@github.com:vandesign/git-bug.git
bad github project url
```

repository/mock_repo.go
```diff
@@ -82,6 +82,17 @@ func (r *mockRepoForTest) RmConfigs(keyPrefix string) error {
...
+		if strings.HasPrefix(key, "remote.origin.url") {
```
MichaelMure: The remote could be different than `origin`. As an example, here is what I have in my config:

```
git config --list | grep "remote\."
remote.origin.url=git@github.com:MichaelMure/git-bug.git
remote.origin.fetch=+refs/heads/*:refs/remotes/origin/*
remote.JonLundy.url=git@github.com:JonLundy/git-bug.git
remote.JonLundy.fetch=+refs/heads/*:refs/remotes/JonLundy/*
remote.vandesign.url=git@github.com:vandesign/git-bug.git
remote.vandesign.fetch=+refs/heads/*:refs/remotes/vandesign/*
remote.adamslc.url=git@github.com:adamslc/git-bug.git
remote.adamslc.fetch=+refs/heads/*:refs/remotes/adamslc/*
remote.steap.url=git@github.com:Steap/git-bug.git
remote.steap.fetch=+refs/heads/*:refs/remotes/Steap/*
remote.programfan.url=git@github.com:ProgramFan/git-bug.git
remote.programfan.fetch=+refs/heads/*:refs/remotes/programfan/*
remote.seeduvax.url=git@github.com:seeduvax/git-bug.git
remote.seeduvax.fetch=+refs/heads/*:refs/remotes/seeduvax/*
remote.sladyn98.url=git@github.com:sladyn98/git-bug.git
remote.sladyn98.fetch=+refs/heads/*:refs/remotes/sladyn98/*
```

repository/git.go
```diff
@@ -207,6 +207,34 @@ func (repo *GitRepo) RmConfigs(keyPrefix string) error {
...
+
+	splitRemotes := strings.Split(out, "\n")
+
+	remotesMap := make(map[string]string, len(splitRemotes))
+
+	for _, line := range splitRemotes {
+
+		remoteValues := strings.Fields(line)
+		if len(remoteValues) != 3 {
+			return nil, fmt.Errorf("Error in remote: %s", line)
```
MichaelMure: `"Error in remote: %s"` doesn't actually said what the problem is. We already know that there is an error because we get an error. Also, error message in go should not start with a capital letter.

You can say `"failed to parse remote %s"`

bridge/github/config.go
```diff
@@ -205,13 +206,58 @@ func promptUsername() (string, error) {
...
-		line, err := bufio.NewReader(os.Stdin).ReadString('\n')
+		fmt.Println("The remotes are:")
+
+		repos, nil := repo.ListRemotes()
+		reposMap := make(map[int]string)
+		i := 1
+		for _, value := range repos {
+
+			if checkGithubURL(value) {
+				fmt.Printf("[%d]: %s\n", i, value)
```
MichaelMure: Could you make sure to match the following UI (display a correct and short URL, space between the detected project and the custom option, `Detected projects`) ?

```
Detected projects:
[1]: github.com/MichaelMure/git-bug
[2]: github.com/adamslc/git-bug

[0]: another project
```

repository/git.go
sladyn98: Yeah I think you are right we need to change the error message.

repository/mock_repo.go
sladyn98: So what do you want the tests to check for? Do you want me to create a custom array of origins or names and test for each?

MichaelMure: You should be able to do:

`mockRepo.StoreConfig("remote.origin.url", "git@github.com:MichaelMure/git-bug.git")`
`mockRepo.ListRemotes()` // return a map with `origin` --> `git@github.com:MichaelMure/git-bug.git`

repository/mock_repo.go
```diff
@@ -84,6 +84,13 @@ func (r *mockRepoForTest) RmConfigs(keyPrefix string) error {
 	return nil
 }
 
+func (r *mockRepoForTest) ListRemotes() (map[string]string, error) {
+
+	r.StoreConfig("remote.origin.url", "git@github.com:example/example.git")
```
golangcibot: Error return value of `r.StoreConfig` is not checked (from `errcheck`)

repository/mock_repo.go
```diff
@@ -84,6 +84,13 @@ func (r *mockRepoForTest) RmConfigs(keyPrefix string) error {
...
+	repos, nil := r.ListRemotes()
```
golangcibot: SA5007: infinite recursive call (from `staticcheck`)

bridge/github/config.go
```diff
@@ -205,13 +206,61 @@ func promptUsername() (string, error) {
...
+					fmt.Println(err)
+					continue
+				}
+				fmt.Printf("[%d]: %s\n", i, shortenedURL)
+				reposMap[i] = value
+				i++
+			}
+		}
+
+		if i == 1 {
```
A-Hilaly: Maybe you dont need `i` at all. You can just check the length of `reposMap`.

bridge/github/config.go
```diff
@@ -205,13 +206,61 @@ func promptUsername() (string, error) {
...
 
-func promptURL() (string, string, error) {
+func promptURL(repo repository.RepoCommon) (string, string, error) {
 	for {
-		fmt.Print("Github project URL: ")
 
-		line, err := bufio.NewReader(os.Stdin).ReadString('\n')
+		fmt.Println("Detected Projects:")
+
+		repos, nil := repo.ListRemotes()
```
A-Hilaly: You need to check for error here. `nil` is a special key word in Go, it would be better to do something like
```go
repos, err := repo.ListRemotes()
if err != nil {
    return "", "", err
}
```

bridge/github/config.go
```diff
@@ -205,13 +206,61 @@ func promptUsername() (string, error) {
...
+		}
+
+		if i == 1 {
+			fmt.Printf("No github projects found\n")
+		}
+
+		fmt.Printf("\n[0]: Another project\n")
+		fmt.Println()
+		var line string
+		output, err := bufio.NewReader(os.Stdin).ReadString('\n')
```
A-Hilaly: Moving error check here would be nice

bridge/github/config.go
```diff
@@ -205,13 +206,61 @@ func promptUsername() (string, error) {
...
+			line = strings.TrimRight(line, "\n")
+
+			if !checkGithubURL(line) {
+				fmt.Printf("Please Enter valid github URL\n")
+				fmt.Println()
+				continue
+			}
+
+			if err != nil {
+				return "", "", fmt.Errorf("Input error")
```
A-Hilaly: Go errors should not be capitalised: https://github.com/golang/go/wiki/CodeReviewComments#error-strings

bridge/github/config.go
```diff
@@ -232,8 +281,25 @@ func promptURL() (string, string, error) {
...
+	if err != nil {
+		panic(err)
+	}
+	res := re.FindStringSubmatch(url)
+
+	if len(res) == 0 {
+		value = false
+	}
+
+	return value
```
A-Hilaly: this is can be done without declaring `value`
```go
if len(res) == 0 {
    return false
}

return true
```

bridge/github/config.go
```diff
@@ -205,15 +206,67 @@ func promptUsername() (string, error) {
...
+
+		if outputInteger == 0 {
+			fmt.Println("Enter Project URL:")
+			line, err = bufio.NewReader(os.Stdin).ReadString('\n')
+			line = strings.TrimRight(line, "\n")
+
+			if !checkGithubURL(line) {
+				fmt.Printf("Please Enter valid github URL\n")
+				fmt.Println()
+				continue
```
A-Hilaly: since this is printing "Please enter a valid github URL" can we just directly recall the url reader ?

bridge/github/config.go
```diff
@@ -205,15 +206,67 @@ func promptUsername() (string, error) {
...
+		fmt.Println()
+		var line string
+		output, err := bufio.NewReader(os.Stdin).ReadString('\n')
+		if err != nil {
+			fmt.Println(err)
+			return "", "", fmt.Errorf("input error")
+		}
+		output = strings.TrimRight(output, "\n")
+
+		outputInteger, _ := strconv.Atoi(output)
```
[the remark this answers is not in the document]
sladyn98: Yeah done

bridge/github/config.go
```diff
@@ -205,15 +206,67 @@ func promptUsername() (string, error) {
...
+			}
+		}
+
+		if len(reposMap) == 0 {
+			fmt.Printf("No github projects found\n")
+		}
+
+		fmt.Printf("\n[0]: Another project\n")
+		fmt.Println()
+		var line string
```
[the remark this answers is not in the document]
sladyn98: Yeah right.

bridge/github/config.go
```diff
@@ -232,8 +285,25 @@ func promptURL() (string, string, error) {
 	}
 }
 
+func checkGithubURL(url string) bool {
+
+	re, err := regexp.Compile(`^[a-z]*\W*github\.com[a-z]*.*$`)
```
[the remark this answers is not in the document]
sladyn98: Would it make much of a difference if we do this.Rather keeping it via multiple calls would still be sane IMO.

bridge/github/config.go
sladyn98: We could, but it would be best if we start he procedure all over again.The previous implementation had it this way so IMO this should do just well

bridge/github/config.go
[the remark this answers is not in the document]
A-Hilaly: The problem with this is that: If there is an error in the regex expression, the only way to detect it would be to try it manually (runtime). Using `MustCompile` (outside of the function) will prevent that by checking the expressions in the compile time.

Compiling regexes is costing a lot also, you can check the difference your self with this test file.
```golang
package git

import (
	"regexp"
	"testing"
)

var Sink bool

func BenchmarkCompileRun(b *testing.B) {
	for i := 0; i < b.N; i++ {
		rx := regexp.MustCompile(`[\w\.+-]+@[\w\.-]+\.[\w\.-]+`)
		Sink = rx.MatchString("123456789 foo@bar.etc")
	}
}

func BenchmarkRun(b *testing.B) {
	rx := regexp.MustCompile(`[\w\.+-]+@[\w\.-]+\.[\w\.-]+`)
	for i := 0; i < b.N; i++ {
		Sink = rx.MatchString("123456789 foo@bar.etc")
	}
}

```

```bash
$ go test -bench=.

goos: darwin
goarch: amd64
pkg: gitlab.com/a-hilaly/blockchain
BenchmarkCompileRun-8             300000              5192 ns/op
BenchmarkRun-8                   2000000               938 ns/op
PASS
ok      gitlab.com/a-hilaly/blockchain   4.458s
```

bridge/github/config.go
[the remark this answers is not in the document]
A-Hilaly: I meant moving it after `strconv.Atoi` call and it err checking. `line` would be useless if the function exists with an error

MichaelMure:
This has been adopted by @A-Hilaly in https://github.com/MichaelMure/git-bug/pull/153, now merged. Thanks @sladyn98 for the help :)

PR #97, opened 2019-02-08
A command to list matching bug id from a prefix

### LINKED ISSUE #47: feat: a command to list matching bug id from a prefix
To be able to auto-complete bug id from a prefix, a hidden command to query the repo for matching id is necessary. 

`git bug ls-id [<prefix>]`

If no prefix is given, output all existing id.

sladyn98:
This file adds the ls-id command which returns
the bug id matching the prefix the user enters.
If no prefix entered it lists all the BugId's

Closes #47

### DIFF
--- a/commands/ls-id.go
+++ b/commands/ls-id.go
@@ -1,0 +1,80 @@
+package commands
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/MichaelMure/git-bug/bug"
+	"github.com/spf13/cobra"
+)
+
+func runLsID(cmd *cobra.Command, args []string) error {
+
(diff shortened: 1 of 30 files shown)


commands/ls-id.go
```diff
@@ -0,0 +1,75 @@
...
+	"strings"
+
+	"github.com/MichaelMure/git-bug/bug"
+	"github.com/spf13/cobra"
+)
+
+func runLsID(cmd *cobra.Command, args []string) error {
+
+	if len(args) < 1 {
+		ListAllID()
```
golangcibot: Error return value of `ListAllID` is not checked (from `errcheck`)

PR #104, opened 2019-03-13
Use Timeline API instead of raw operations

### LINKED ISSUE #94: WebUI needs to be migrated to use the Timeline
After editing the first comment (aka description) of an issue `git bug show $ID` correctly shows the latest version of that comment.

The WebUI however, only shows the first version of that comment.

sandhose:
Fixes #94 

The webui now uses the `Timeline` API instead of the raw `Operation`s.

### DIFF
--- a/webui/src/Author.js
+++ b/webui/src/Author.js
@@ -1,28 +1,25 @@
-import { withStyles } from '@material-ui/core/styles';
 import Tooltip from '@material-ui/core/Tooltip/Tooltip';
+import MAvatar from '@material-ui/core/Avatar';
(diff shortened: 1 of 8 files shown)


MichaelMure:
:+1: :+1:

MichaelMure:
@sandhose can we merge this so the webUI is less broken than before ?

sandhose:
image

The timeline now looks like this.

I haven't packed the assets to avoid conflicts.

MichaelMure:
Super cool, thanks !

PR #106, opened 2019-03-23
Select and Label command documentation

jfacchini:
Hey @MichaelMure ! Ça roule?

I was getting started with the tool and got a bit confused when I tried to add a label. I thought it was creating a list of labels... So I thought I would give a bit more clues on these commands by adding a bit more info.

- Informed that label command adds/removes a label to a given bug.
- Added suggestion to use select command when omitting to provide a bug id as argument.
- Added a bit more detail for the select command documentation, giving a hint to "deselect" command.


MichaelMure:
3a71

Issue #115, opened 2019-03-31
Use And Query in Label Search


sladyn98:
With respect to this comment https://github.com/MichaelMure/git-bug/issues/99#issuecomment-471210546.   If multiple lables are included in the query :
for eg: ` --label:easypick --label:hard` git-bug fetches those which have either `easypick` OR  `hard` whereas according to the github convention multiple labels specified should use AND and return bugs with both labels.

PR #116, opened 2019-03-31
cache/filter.go: Fix query match for --label 

sladyn98:
Fixed filter query for multiple labels from
`OR` to `AND` according to github conventions.

Fixed https://github.com/MichaelMure/git-bug/issues/115


MichaelMure:
:+1:

Issue #107, opened 2019-03-27
Have the list of participants easily accessible in the API


sandhose:
I'd like to add the list of participants in the webui, like in gitea:

![image](https://user-images.githubusercontent.com/1549952/55111837-8cb14b00-50db-11e9-844f-814d6cbfd26b.png)

For that, I need the list of people that interacted with the bug in the GraphQL API. Because it implies to go through the whole list of operations to lists authors, this might be something we want to have in the cache.

MichaelMure:
Good point !

We don't necessarely need to have the list of participants in the cache, because we load the full bug from git anyway, but it would still make sense if we want to be able to query bugs based on who interacted with them.

There is two type of participants that we might want to expose:
- participants: only users that wrote a comment or the initial bug
- actors: any user that interacted, including setting a label, changing the title ...

- [ ] add new functions on a Bug to compute both lists
- [ ] expose each lists in the Snapshot
- [ ] expose each lists in the GraphQL API
- [ ] add each list in the BugExcerpt
- [ ] implement new query selectors

A-Hilaly:
Hello !

This is a good starting point for me, can i work on it ?

MichaelMure:
Sure, go ahead ! Feel free to ask for clarifications as needed as well.

A-Hilaly:
Thanks ! I'll open a PR as soon as i have something working !

MichaelMure:
This has been dealt with by @A-Hilaly. @sandhose your turn ;)

PR #117, opened 2019-03-31
WIP: Implement participants and actors functionalities

A-Hilaly:
Dealing with #107 

- [x]  add `Participants` and `Actors` fields to Snapshot
- [x]  append operations authors to each list on `Apply()` call
- [x]  expose each lists in the GraphQL API
- [x]  add each list in the BugExcerpt
- [x]  implement new query filter + doc

### DIFF
--- a/bug/snapshot.go
+++ b/bug/snapshot.go
@@ -12,12 +12,14 @@ import (
 type Snapshot struct {
 	id string
 
-	Status    Status
-	Title     string
-	Comments  []Comment
-	Labels    []Label
-	Author    identity.Interface
-	CreatedAt time.Time
+	Status       Status
+	Title        string
+	Comments     []Comment
+	Labels       []Label
+	Author       identity.Interface
+	Actors       []identity.Interface
+	Participants []identity.Interface
+	CreatedAt    time.Time
 
 	Timeline []TimelineItem
 
@@ -61,4 +63,26 @@ func (snap *Snapshot) SearchTimelineItem(hash git.Hash) (TimelineItem, error) {
 	}
 
 	return nil, fmt.Errorf("timeline item not found")
+}
+
+// append the operation author to the actors list
+func (snap *Snapshot) addActor(actor identity.Interface) {
+	for _, a := range snap.Actors {
+		if actor.Id() == a.Id() {
+			return
+		}
+	}
+
+	snap.Actors = append(snap.Actors, actor)
+}
+
+// append the operation author to the participants list
+func (snap *Snapshot) addParticipant(participant identity.Interface) {
+	for _, p := range snap.Participants {
+		if participant.Id() == p.Id() {
+			return
+		}
+	}
+
+	snap.Participants = append(snap.Participants, participant)
 }
--- a/cache/bug_excerpt.go
+++ b/cache/bug_excerpt.go
@@ -23,10 +23,12 @@ type BugExcerpt struct {
 	CreateUnixTime    int64
 	EditUnixTime      int64
 
-	Status      bug.Status
-	Labels      []bug.Label
-	Title       string
-	LenComments int
+	Status       bug.Status
+	Labels       []bug.Label
+	Title        string
+	LenComments  int
+	Actors       []string
+	Participants []string
 
 	// If author is identity.Bare, LegacyAuthor is set
 	// If author is identity.Identity, AuthorId is set and data is deported
@@ -44,6 +46,16 @@ type LegacyAuthorExcerpt struct {
 }
 
 func NewBugExcerpt(b bug.Interface, snap *bug.Snapshot) *BugExcerpt {
+	participantsIds := make([]string, len(snap.Participants))
+	for i, participant := range snap.Participants {
+		participantsIds[i] = participant.Id()
+	}
+
+	actorsIds := make([]string, len(snap.Actors))
+	for i, actor := range snap.Actors {
+		actorsIds[i] = actor.Id()
+	}
+
 	e := &BugExcerpt{
 		Id:                b.Id(),
 		CreateLamportTime: b.CreateLamportTime(),
@@ -52,6 +64,8 @@ func NewBugExcerpt(b bug.Interface, snap *bug.Snapshot) *BugExcerpt {
 		EditUnixTime:      snap.LastEditUnix(),
 		Status:            snap.Status,
 		Labels:            snap.Labels,
+		Actors:            actorsIds,
+		Participants:      participantsIds,
 		Title:             snap.Title,
 		LenComments:       len(snap.Comments),
 		CreateMetadata:    b.FirstOp().AllMetadata(),
--- a/cache/filter.go
+++ b/cache/filter.go
@@ -56,35 +56,43 @@ func LabelFilter(label string) Filter {
 }
 
 // ActorFilter return a Filter that match a bug actor
-func ActorFilter(actor string) Filter {
+func ActorFilter(query string) Filter {
 	return func(repoCache *RepoCache, excerpt *BugExcerpt) bool {
-		for _, identityExcerpt := range repoCache.identitiesExcerpts {
-			if strings.Contains(strings.ToLower(identityExcerpt.Name), actor) ||
-				actor == identityExcerpt.Id || actor == identityExcerpt.Login {
-				for _, actorId := range excerpt.Actors {
-					if identityExcerpt.Id == actorId {
-						return true
-					}
-				}
+		query = strings.ToLower(query)
+
+		for _, id := range excerpt.Actors {
+			identityExcerpt, ok := repoCache.identitiesExcerpts[id]
+			if !ok {
+				panic("missing identity in the cache")
+			}
+
+			if strings.Contains(strings.ToLower(identityExcerpt.Name), query) ||
+				query == identityExcerpt.Id || query == strings.ToLower(identityExcerpt.Login) {
+				return true
 			}
 		}
+
 		return false
 	}
 }
 
 // ParticipantFilter return a Filter that match a bug participant
-func ParticipantFilter(participant string) Filter {
+func ParticipantFilter(query string) Filter {
 	return func(repoCache *RepoCache, excerpt *BugExcerpt) bool {
-		for _, identityExcerpt := range repoCache.identitiesExcerpts {
-			if strings.Contains(strings.ToLower(identityExcerpt.Name), participant) ||
-				participant == identityExcerpt.Id || participant == identityExcerpt.Login {
-				for _, participantId := range excerpt.Participants {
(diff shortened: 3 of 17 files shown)


bug/bug.go
```diff
@@ -667,8 +667,10 @@ func (bug *Bug) LastOp() Operation {
 // Compile a bug in a easily usable snapshot
 func (bug *Bug) Compile() Snapshot {
 	snap := Snapshot{
-		id:     bug.id,
-		Status: OpenStatus,
+		id:           bug.id,
+		Actors:       bug.computeActors(),
```
A-Hilaly: You think it's better to have one function that compute both actors and participants in order to loop only once on operations ?

bug/bug.go
```diff
@@ -681,3 +683,65 @@ func (bug *Bug) Compile() Snapshot {
 
 	return snap
 }
+
+// Participants return the participants list of the bug.
+// Participants are identities that added comments to a
+// bug or created it.
+func (bug *Bug) computeParticipants() []identity.Interface {
```
A-Hilaly: is the right approach to list participants/actors ?

bug/bug.go
```diff
@@ -681,3 +683,65 @@ func (bug *Bug) Compile() Snapshot {
...
+	participants := make([]identity.Interface, 0, 1)
+
+	it := NewOperationIterator(bug)
```
A-Hilaly: Should i loop only over packed operations instead of using `NewOperationIterator`?

MichaelMure:
@A-Hilaly if you do the GraphQL part, have a look at https://github.com/MichaelMure/git-bug/commit/15c258cdc4ba37820362a44dfc2636ed1ff92b4c, just now in master, that introduce a lazy loading Identity connection for GraphQL. You most likely can reuse it easily.

A-Hilaly:
That's useful thanks !
I just exposed the fields on bug excerpt. I'll start working on graphql part now

cache/bug_excerpt.go
```diff
@@ -52,6 +54,8 @@ func NewBugExcerpt(b bug.Interface, snap *bug.Snapshot) *BugExcerpt {
 		EditUnixTime:      snap.LastEditUnix(),
 		Status:            snap.Status,
 		Labels:            snap.Labels,
+		Actors:            snap.Actors(),
```
MichaelMure: You only want to store the id here, not the full identity. The id will need to be resolved in a full Identity later, wherever it's needed (in the query filter most likely).

MichaelMure: ^ this is to be efficient and store the minimal amount of data on disk, as it will be read and loaded in memory each time git-bug start.

A-Hilaly: Okkey I see why the tests failed now, i'll rebase and make the changes

sladyn98:
@gitmate-bot rebase

MichaelMure:
Hey! I'm [GitMate.io](https://gitmate.io)! This pull request is being rebased automatically. Please **DO NOT** push while rebase is in progress or your changes would be lost permanently :warning:

MichaelMure:
Automated rebase failed! Please rebase your pull request manually via the command line.

Reason:
```
Command: git rebase base/master

Exit Code: 128

Cause:

error: Failed to merge in the changes.


```

bug/bug_test.go
```diff
@@ -109,3 +109,63 @@ func equivalentBug(t *testing.T, expected, actual *Bug) {
...
+
+	bug.Append(createOp)
+	bug.Append(setTitleOp)
+	bug.Append(addCommentOp)
+	bug.Append(labelChangeOp)
+
+	repo := repository.NewMockRepoForTest()
+
+	err := bug.Commit(repo)
+	assert.Nil(t, err)
```
MichaelMure: you can use `assert.NoError(t, err)` that will give a better output if something break

graphql/graphql_test.go
```diff
@@ -123,13 +133,15 @@ func TestQueries(t *testing.T) {
...
 				PageInfo models.PageInfo
 				Nodes    []struct {
-					Author    Person
-					CreatedAt string `json:"createdAt"`
-					HumanId   string `json:"humanId"`
-					Id        string
-					LastEdit  string `json:"lastEdit"`
-					Status    string
-					Title     string
+					Author       Person
```
MichaelMure: that's a place I forgot to update when I renamed Person to Identity. Would you mind updating it here as well ?

A-Hilaly: I'll make that in my next changes

A-Hilaly: using `identity.Interface` instead of `Person` causes graphql tests to fail. I get some like 
```bash
'DefaultRepository.AllBugs.Nodes[1].Author' expected type 'identity.Interface', got 'map[string]interface {}'
```
maybe we should use a special model struct for identity ?

bug/snapshot.go
```diff
@@ -62,3 +64,27 @@ func (snap *Snapshot) SearchTimelineItem(hash git.Hash) (TimelineItem, error) {
 
 	return nil, fmt.Errorf("timeline item not found")
 }
+
+// append the operation author to the actors list
+func (snap *Snapshot) addActor(actor identity.Interface) {
+	for _, a := range snap.Actors {
```
A-Hilaly: Maybe should i just store actors and participants in a `map[string]identity.Interface` to avoid looping each time `addActor` ?

bug/op_set_metadata.go
```diff
@@ -25,6 +25,8 @@ func (op *SetMetadataOperation) Hash() (git.Hash, error) {
 }
 
 func (op *SetMetadataOperation) Apply(snapshot *Snapshot) {
+	snapshot.addActor(op.Author)
```
MichaelMure: Adding metadata to an operation is typically done by the importers/exporters. I'd say it doesn't qualify as being an actor.

bug/snapshot.go
MichaelMure: As the Snapshot API would stay the same, this is really a question of performance. Is it better to use more CPU, or more memory ? I'd say it's better this way, as it avoid any memory allocation, and no GC as well (so less CPU !).

bug/snapshot.go
```diff
@@ -62,3 +64,27 @@ func (snap *Snapshot) SearchTimelineItem(hash git.Hash) (TimelineItem, error) {
...
+// append the operation author to the participants and actors list
+func (snap *Snapshot) addParticipant(participant identity.Interface) {
+	for _, p := range snap.Participants {
+		if participant.Id() == p.Id() {
+			return
+		}
+	}
+
+	// a participant is an actor too
+	snap.Actors = append(snap.Actors, participant)
```
MichaelMure: Yes, a participant is also an actor. But what is the name of your function ? ;)

For readability, it's much better to have the calling code do both `addParticipant` and `addActor` instead of obfuscating what's really happening.

graphql/graphql_test.go
MichaelMure: Hoo, I got what you are doing. `Person` here is just a dumb struct (a few lines up) to store the graphql response. I just wanted to rename this struct to `Identity`, not using `identity.Interface`

bug/snapshot.go
A-Hilaly: Agreed !

A-Hilaly:
@gitmate-bot rebase

cache/filter.go
```diff
@@ -55,6 +55,40 @@ func LabelFilter(label string) Filter {
 	}
 }
 
+// ActorFilter return a Filter that match a bug actor
+func ActorFilter(actor string) Filter {
+	return func(repoCache *RepoCache, excerpt *BugExcerpt) bool {
+		for _, identityExcerpt := range repoCache.identitiesExcerpts {
```
MichaelMure: Two things:
1. you need to force `actor` to lower case to be fully case insensitive. Btw, let's call it `query`, that's what it is ;)
1. you are iterating over *all* the known identity in the repository, checking if they match the query and then checking if that's one of the actors of the bug. You can simply iterate over the actorIds of the bug, get the corresponding IdentityExcerpt (it's a map) and check if it match the query. Much less work to do !

graphql/schema/bug.graphql
```diff
@@ -36,6 +36,8 @@ type Bug {
   title: String!
   labels: [Label!]!
   author: Identity!
+  actors: [Identity]!
```
MichaelMure: After talking with @sandhose, we'll go for the connection here. But we can merge like this and I'll add it. It's a bit tricky.

A-Hilaly: Alright ! Ping me when it's ready. I'd like to see how that's implemented 👍

cache/filter.go
A-Hilaly: > You can simply iterate over the actorIds of the bug get the corresponding IdentityExcerpt

Yes totally agree. That's a better approach

doc/queries.md
```diff
@@ -33,6 +33,27 @@ You can filter based on the person who opened the bug.
 | `author:QUERY` | `author:descartes` matches bugs opened by `René Descartes` or `Robert Descartes` |
 |                | `author:"rené descartes"` matches bugs opened by `René Descartes`                |
 
+### Filtering by participant
+
+You can filter based on the person who participated in the bug (Opened the bug or added a comment).
```
sladyn98: Could this be made `You can filter based on the person who participated in any activity related to the bug (Opened bug or added a comment`

cache/filter.go
```diff
@@ -55,6 +55,48 @@ func LabelFilter(label string) Filter {
...
+			if !ok {
+				panic("missing identity in the cache")
+			}
+
+			if strings.Contains(strings.ToLower(identityExcerpt.Name), query) ||
+				query == identityExcerpt.Id || query == strings.ToLower(identityExcerpt.Login) {
+				return true
+			}
+		}
+
```
sladyn98: You could use `gofmt` to reduce unnecessary white spaces :P I mean its not a necessacity but a good coding practice :/

A-Hilaly: it is `gofmt`-ed. Sometimes it's good to leave space to simplify reading. But i agree last space before return isn't necessary

cache/query.go
```diff
@@ -56,13 +56,21 @@ func ParseQuery(query string) (*Query, error) {
...
+			result.Participant = append(result.Participant, f)
+
 		case "label":
 			f := LabelFilter(qualifierQuery)
 			result.Label = append(result.Label, f)
 
 		case "title":
 			f := TitleFilter(qualifierQuery)
-			result.Label = append(result.Title, f)
+			result.Title = append(result.Title, f)
```
MichaelMure: good catch !

MichaelMure:
:+1: :+1: :+1:

Issue #108, opened 2019-03-28
comment edition with the command line


MichaelMure:
Comment needs to be editable from the command line.

This warrant a new command: `git bug comment edit <id>`

Comment's id can be retrieved using `git bug comment`

sladyn98:
This is a feature that is required since it is already present in the TermUI.
I will try and send in a PR for this issue :+1:

sladyn98:
@MichaelMure one more thing we would need to take care of would be comment identities.We cannot allow edits to users to whom the comment does not belong.

MichaelMure:
Yes, but to do that properly, we will need cryptography and signature, as well as a ACL system. I don't want to add a "soft" constraint for now (only allow the author to edit), because that would make the current version forward incompatible once more than the author can edit (on github, repo admin can edit a comment as an example).

sladyn98:
But in the case when the exporter is developed this system becomes paramount. For now we could allow edits to comments by any author. And in the last case you mentioned we would have to first determine the repo admin to allow him the edit... beacuse if not we wouldn't be able to export the bug.

MichaelMure:
This has been working for long

PR #128, opened 2019-04-16
comment_edit.go:Added comment edit feature

sladyn98:
Gives users the option to edit a comment
by specifying the comment Id

Closes https://github.com/MichaelMure/git-bug/issues/108

### DIFF
--- a/commands/comment_edit.go
+++ b/commands/comment_edit.go
@@ -1,0 +1,65 @@
+package commands
+
+import (
+	"fmt"
+
+	"github.com/MichaelMure/git-bug/cache"
+	"github.com/MichaelMure/git-bug/input"
+	"github.com/MichaelMure/git-bug/util/git"
+	"github.com/MichaelMure/git-bug/util/interrupt"
+	"github.com/spf13/cobra"
+)
+
+func runCommentEdit(cmd *cobra.Command, args []string) error {
+
+	backend, err := cache.NewRepoCache(repo)
+	if err != nil {
+		return err
+	}
+	defer backend.Close()
+	interrupt.RegisterCleaner(backend.Close)
+
+	for _, id := range backend.AllBugsIds() {
+		b, err := backend.ResolveBugPrefix(id)
+		if err != nil {
+			return err
+		}
+		snap := b.Snapshot()
+
+		for _, comment := range snap.Comments {
+
+			if args[0] == comment.HumanId() {
+
+				commentEdit, err := input.BugCommentEditorInput(repo, comment.Message)
+				if err == input.ErrEmptyMessage {
+					fmt.Println("Empty message, aborting.")
+					return nil
+				}
+				if commentEdit == comment.Message {
+					fmt.Println("No changes found, aborting.")
+					return nil
+				}
+				_, err = b.EditComment(git.Hash(comment.Id()), commentEdit)
+				if err != nil {
+					return err
+				}
+				fmt.Println("Comment edited successfully")
+				return b.Commit()
+			}
+		}
+	}
+	return fmt.Errorf("no bug found with matching Id")
+
+}
+
+var commentEditCmd = &cobra.Command{
+	Use:     "edit [<id>]",
+	Short:   "Edit a comment of a bug.",
+	PreRunE: loadRepo,
+	RunE:    runCommentEdit,
+}
+
+func init() {
+	commentCmd.AddCommand(commentEditCmd)
+	commentEditCmd.Flags().SortFlags = false
+}
--- a/doc/man/git-bug-comment.1
+++ b/doc/man/git-bug-comment.1
@@ -26,4 +26,4 @@ Display or add comments to a bug.
 
 .SH SEE ALSO
 .PP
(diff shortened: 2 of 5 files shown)


commands/comment_edit.go
```diff
@@ -0,0 +1,72 @@
...
+
+	"github.com/MichaelMure/git-bug/cache"
+	"github.com/MichaelMure/git-bug/input"
+	"github.com/MichaelMure/git-bug/util/git"
+	"github.com/MichaelMure/git-bug/util/interrupt"
+	"github.com/spf13/cobra"
+)
+
+var (
+	commentEditComment string
```
golangcibot: `commentEditComment` is unused (from `deadcode`)

commands/comment_edit.go
```diff
@@ -0,0 +1,72 @@
...
+	for _, id := range backend.AllBugsIds() {
+		b, err := backend.ResolveBugPrefix(id)
+		if err != nil {
+			return err
+		}
+		snap := b.Snapshot()
+
+		for _, comment := range snap.Comments {
+
+			if err != nil {
```
golangcibot: nilness: impossible condition: nil != nil (from `govet`)

MichaelMure:
Hey, I had a look at this.

It works, but there is a problem of performance. I have only 59 bugs locally, and this command take almost 3 seconds to find the comment to edit it. This is because every single bug is fully loaded from the disk until the correct one is found. That wont do.

We need a different approach.

I was thinking that maybe we should, instead of using just the hash of the operation to represent the comment, use a concatenation of both the bug hash and the operation hash.

Say we have a bug with a hash/id of `16839f3b0c9f89e87fd2c0fab71ef2d13cb1564b`. The human readable version is the truncation to 7 characters: `16839f3`.

Currently, a comment has a hash/id that is the hash of the operation that created it. Say for the example, `a681c01fdafe9f125348984024bab60cccbfb255`. The human readable version is again, the truncation to 7 characters: `a681c01`.

We could say that we assign to the comment the id `16839f3b0c9f89e87fd2c0fab71ef2a681c01fda`, composed of 30 characters of the bug id, and 10 of the operation. Humand version would be the concat of both: `16839f3a681c01`

This would leave a key space of 36^30 for the bug (~5 * 10^46) and 36^10 for the comment (~3 * 10^15).

Using the maths from, https://www.ilikebigbits.com/2018_10_20_estimating_hash_collisions.html, with a maximum probability of one in a million, that give space for 10^20 bugs and 58000 comments in one bug. That seems reasonable (?).

With that you could parse the id and directly fetch the bug, then the comment.

That needs some more thinking though.

sladyn98:
@MichaelMure  Yeah the math is perfect in your case, I was thinking more on the lines of a caching mechanism where we store the `bugId` as a key and all the corresponding comments in the corresponding bucket of that key as values.In practice the number of comments for a particular bug would not be that large.The lookup in a hashtable would be `O(1)` and the search in the value space would be `O(n)` where n is the number of comments. So in worst case we would have `O(n)`

MichaelMure:
I just had a literal shower thought about this.

Instead of concatenating a part of the bug ID with a part of the comment ID, we could *interleave* them.

That is, instead of doing `BBBBBBBBBBBBBBBBBBBBBBBBBBBBBBCCCCCCCCCC` with B being a character of the bug's ID and C being a characters of the comment's ID, we could do `BBBCBBBCBBBCBBBCBBBCBBBCBBBCBBBC`.

This way, the longer the prefix of this ID is, the more information we get about *both* the bug and the comment.

We could even do an irregular pattern if we want to be able to have meaningful 4+3, 6+3 or 6+4 prefix, while still end up with a nice 30+10 for the full ID:

| | |
| ----- | ---- |
| 4+3 |     `BCBCBCB`|
| 6+3 |     `BCBCBCBBB`|
| 6+4 |      `BCBCBCBBBC`|
| 30+10 |  `BCBCBCBBBCBBBBCBBBBCBBBBCBBBBCBBBBCBBBBC` |

sladyn98:
> I just had a literal shower thought about this.
> 
> Instead of concatenating a part of the bug ID with a part of the comment ID, we could _interleave_ them.
> 
> That is, instead of doing `BBBBBBBBBBBBBBBBBBBBBBBBBBBBBBCCCCCCCCCC` with B being a character of the bug's ID and C being a characters of the comment's ID, we could do `BBBCBBBCBBBCBBBCBBBCBBBCBBBCBBBC`.
> 
> This way, the longer the prefix of this ID is, the more information we get about _both_ the bug and the comment.
> 
> We could even do an irregular pattern if we want to be able to have meaningful 4+3, 6+3 or 6+4 prefix, while still end up with a nice 30+10 for the full ID:
> 
> 4+3	`BCBCBCB`
> 6+3	`BCBCBCBBB`
> 6+4	`BCBCBCBBBC`
> 30+10	`BCBCBCBBBCBBBBCBBBBCBBBBCBBBBCBBBBCBBBBC`

This seems like a nice plan.This would give us the freedom of not only adding more information about the bug but also the comment.

sladyn98:
@MichaelMure I still have some time to work on this, is the previous comment the way you want the comment feature edited.

MichaelMure:
Closing in favor or #424

PR #131, opened 2019-05-03
Github bridge and importer changes

### LINKED ISSUE #74: Bug: github import missing comments.
Here's a particular example: https://github.com/AladW/aurutils/issues/363.
From my brief conversation with @MichaelMure on Gitter, this seems to be caused by not handling deleted edits properly.

A-Hilaly:
- implement a Github iterator to separate the import logic from graphql queries
- change the bridge interfaces to support imports and exports after a known date
- add importer tests
- add ForceChangeLabels functionalities to `bugCache`
- fix #74


bridge/github/import_test.go
```diff
@@ -0,0 +1,220 @@
...
+	backend, err := cache.NewRepoCache(repo)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	defer backend.Close()
+	interrupt.RegisterCleaner(backend.Close)
+
+	importer := &githubImporter{}
+	importer.Init(core.Configuration{
```
golangcibot: Error return value of `importer.Init` is not checked (from `errcheck`)

bridge/github/import_test.go
```diff
@@ -0,0 +1,223 @@
...
+	}
+
+	defer backend.Close()
+	interrupt.RegisterCleaner(backend.Close)
+
+	importer := &githubImporter{}
+	err = importer.Init(core.Configuration{
+		"user":    "MichaelMure",
+		"project": "git-but-test-github-bridge",
+		"token":   os.Getenv("GITHUB_TOKEN"),
```
MichaelMure: A check with an explicit error (`Env var GITHUB_TOKEN missing`) would be nice instead of an auth failure later

bridge/github/import_test.go
```diff
@@ -0,0 +1,223 @@
...
+						}
+						t.Errorf("bad operation type index = %d expected = AddCommentOperation", i)
+					case *bug.EditCommentOperation:
+						if op2, ok := ops[i].(*bug.EditCommentOperation); ok {
+							assert.Equal(t, op.(*bug.EditCommentOperation).Message, op2.Message)
+							continue
+						}
+						t.Errorf("bad operation type index = %d expected = EditCommentOperation", i)
+					default:
+
```
MichaelMure: default case should panic because we are not checking anything (that's a dev error, let's crash violently)

bridge/github/import_test.go
```diff
@@ -0,0 +1,223 @@
...
+				for i, op := range tt.bug.Operations {
+					switch op.(type) {
+					case *bug.CreateOperation:
+						if op2, ok := ops[i].(*bug.CreateOperation); ok {
+							assert.Equal(t, op2.Title, op.(*bug.CreateOperation).Title)
+							assert.Equal(t, op2.Message, op.(*bug.CreateOperation).Message)
+							continue
+						}
+						t.Errorf("bad operation type index = %d expected = CreationOperation", i)
```
MichaelMure: you can probably do that once in the loop, and then remove all the type checks and `continue`

bridge/github/import_test.go
```diff
@@ -0,0 +1,223 @@
...
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			b, err := backend.ResolveBugCreateMetadata(keyGithubUrl, tt.url)
+			if err != nil {
+				t.Fatal(err)
+			}
+
+			ops := b.Snapshot().Operations
+			if tt.exist {
```
MichaelMure: you can deal with this bad case early and keep a good [line of sight](https://medium.com/@matryer/line-of-sight-in-code-186dd7cdea88):

```
if !tt.exist {
    assert.Equal(t, b, nil)
    continue
}

...
```

MichaelMure: but actually, do you need that test ? `tt.exist` doesn't seems to be false anywhere.

bridge/github/iterator.go
```diff
@@ -0,0 +1,390 @@
...
+	// number of timelines/userEditcontent/issueEdit to query
+	// at a time, more capacity = more used memory = less queries
+	// to make
+	capacity int
+
+	// sticky error
+	err error
+
+	// count to keep track of the number of imported issues
+	count int
```
MichaelMure: maybe a more explicit naming ? `importedIssues` ?

bridge/github/iterator.go
```diff
@@ -0,0 +1,390 @@
...
+	i.commentEdit.variables["timelineFirst"] = githubv4.Int(1)
+	i.commentEdit.variables["timelineAfter"] = (*githubv4.String)(nil)
+	i.commentEdit.variables["commentEditLast"] = githubv4.Int(i.capacity)
+	i.commentEdit.variables["commentEditBefore"] = (*githubv4.String)(nil)
+}
+
+// reverse UserContentEdits arrays in both of the issue and
+// comment timelines
+func (i *iterator) reverseTimelineEditNodes() {
+	reverseEdits(i.timeline.query.Repository.Issues.Nodes[0].UserContentEdits.Nodes)
```
MichaelMure: creating an temporary variable would greatly improve the readability:
`node := i.timeline.query.Repository.Issues.Nodes[0]`

bridge/github/iterator.go
```diff
@@ -0,0 +1,390 @@
...
+}
+
+// Next issue
+func (i *iterator) NextIssue() bool {
+	// we make the first move
+	if i.count == 0 {
+
+		// init variables and goto queryIssue block
+		i.initTimelineQueryVariables()
+		goto queryIssue
```
MichaelMure: if you create `func (i *iterator) queryIssue() bool` you don't need to use `goto`:

```
if i.count == 0 {
		i.initTimelineQueryVariables()
		return i.queryIssue()
}
```

You can call that new function in the bottom part of this function as well.

bridge/github/import.go
```diff
@@ -13,248 +14,133 @@ import (
...
-		if err != nil {
-			return err
-		}
-
-		if len(q.Repository.Issues.Nodes) == 0 {
-			return nil
-		}
+// ImportAll .
+func (gi *githubImporter) ImportAll(repo *cache.RepoCache, since time.Time) error {
+	gi.iterator.since = since
```
MichaelMure: While this work, I feel like is could easily leads to bugs later when the code is updated, because you are changing internal variables. Your iterator should works with only the public interface (that is, it should be possible to move it in another package without breaking anything).

Two options:
- create a new local iterator for each call of `ImportAll`: this has the added benefit of making your importer stateless and thread safe
- add a `Reset(since time.Time)` function where the iterator deal with its internals to reset.

bridge/github/import.go
```diff
@@ -587,8 +389,10 @@ func (gi *githubImporter) ensureCommentEdit(repo *cache.RepoCache, b *cache.BugC
 	switch {
 	case edit.DeletedAt != nil:
 		// comment deletion, not supported yet
+		fmt.Println("comment deletion ....")
```
MichaelMure: if you want to print something, you should say explicitely that it didn't work: `comment deletion is not supported yet`

bridge/github/import.go
```diff
@@ -670,7 +475,7 @@ func (gi *githubImporter) getGhost(repo *cache.RepoCache) (*cache.IdentityCache,
 		"login": githubv4.String("ghost"),
 	}
 
-	err = gi.client.Query(context.TODO(), &q, variables)
+	err = gi.iterator.gc.Query(context.TODO(), &q, variables)
```
MichaelMure: it works, but it does tangle the importer and the iterator together. It would be better if the importer has it's own client. As it's a one shot call, you can probably create it just for this function.

bridge/github/iterator.go
```diff
@@ -0,0 +1,390 @@
...
+	timeline timelineIterator
+
+	// issue edit iterator
+	issueEdit issueEditIterator
+
+	// comment edit iterator
+	commentEdit commentEditIterator
+}
+
+func newIterator(conf core.Configuration) *iterator {
```
MichaelMure: to untangle the iterator from the importer, this should:
- be public
- not use the importer config to source the useful values, and take them directly instead.

bridge/github/import_test.go
A-Hilaly: Its true `tt.exist` is useless now. Before writing the tests i thought it can be useful for some test cases.

bridge/github/iterator.go
A-Hilaly: Yep `importedIssues` is clearer !

util/text/transform.go
```diff
@@ -0,0 +1,34 @@
...
+	text = strings.Replace(text, "\r\n", "\n", -1)
+
+	// remove all unicode control characters except
+	// '\n', '\r' and '\t'
+	t := runes.Remove(runes.Predicate(func(r rune) bool {
+		switch r {
+		case '\r', '\n', '\t':
+			return false
+		}
+		if unicode.IsControl(r) {
```
golangcibot: S1008: should use 'return <expr>' instead of 'if <expr> { return <bool> }; return <bool>' (from `gosimple`)

PR #135, opened 2019-05-17
Enhancement to Github iterator and importer

A-Hilaly:
- Iterator now ignore nil and empty edits
- Fix imported bugs count
- Add imported identities count

### DIFF
--- a/bridge/github/import.go
+++ b/bridge/github/import.go
@@ -54,22 +54,15 @@ func (gi *githubImporter) ImportAll(repo *cache.RepoCache, since time.Time) erro
 		fmt.Printf("importing issue: %v %v\n", gi.importedIssues+1, issue.Title)
 		gi.importedIssues++
 
-		// get issue edits
-		issueEdits := []userContentEdit{}
-		for gi.iterator.NextIssueEdit() {
-			issueEdits = append(issueEdits, gi.iterator.IssueEditValue())
-		}
-
 		// create issue
-		b, err := gi.ensureIssue(repo, issue, issueEdits)
+		b, err := gi.ensureIssue(repo, issue)
 		if err != nil {
 			return fmt.Errorf("issue creation: %v", err)
 		}
 
 		// loop over timeline items
 		for gi.iterator.NextTimeline() {
-			item := gi.iterator.TimelineValue()
-			if err := gi.ensureTimelineItem(repo, b, item); err != nil {
+			if err := gi.ensureTimelineItem(repo, b, gi.iterator.TimelineValue()); err != nil {
 				return fmt.Errorf("timeline event creation: %v", err)
 			}
 		}
@@ -90,7 +83,7 @@ func (gi *githubImporter) ImportAll(repo *cache.RepoCache, since time.Time) erro
 	return nil
 }
 
-func (gi *githubImporter) ensureIssue(repo *cache.RepoCache, issue issueTimeline, issueEdits []userContentEdit) (*cache.BugCache, error) {
+func (gi *githubImporter) ensureIssue(repo *cache.RepoCache, issue issueTimeline) (*cache.BugCache, error) {
 	// ensure issue author
 	author, err := gi.ensurePerson(repo, issue.Author)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


bridge/github/import.go
```diff
@@ -23,61 +23,46 @@ const (
...
 
+func (gi *githubImporter) Reset() {
+	gi.importedIssues = 0
+	gi.importedIdentities = 0
+}
+
 // ImportAll .
 func (gi *githubImporter) ImportAll(repo *cache.RepoCache, since time.Time) error {
-	iterator := NewIterator(gi.conf[keyUser], gi.conf[keyProject], gi.conf[keyToken], since)
+	gi.iterator = NewIterator(gi.conf[keyUser], gi.conf[keyProject], gi.conf[keyToken], since)
```
A-Hilaly: Was forced to do this since the `ImportAll`, `ensureIssue` and `ensurePerson`  functions need to have access to `importedIssues`/`importedIdentities` variables

bridge/github/iterator.go
```diff
@@ -49,8 +49,8 @@ type iterator struct {
 	// sticky error
 	err error
 
-	// number of imported issues
-	importedIssues int
+	// started
+	started bool
```
A-Hilaly: Maybe we can use a better name for this. For now it is only used to know if the iterator has already started querying or not.

bridge/github/import.go
```diff
@@ -23,61 +23,46 @@ const (
...
+	importedIdentities int
 }
 
 func (gi *githubImporter) Init(conf core.Configuration) error {
 	gi.conf = conf
 	return nil
 }
 
+func (gi *githubImporter) Reset() {
+	gi.importedIssues = 0
```
MichaelMure: This doesn't fully do what the function's name imply: you still have the iterator laying around with the current iteration state.

Pro-tip: write the documentation (especially for public functions). This will force you to phrase clearly what the function do or should do, and you will see what is missing or wrong, if any.

bridge/github/import.go
```diff
@@ -87,16 +72,17 @@ func (gi *githubImporter) ImportAll(repo *cache.RepoCache, since time.Time) erro
...
 	}
 
-	if err := iterator.Error(); err != nil {
+	if err := gi.iterator.Error(); err != nil {
 		fmt.Printf("import error: %v\n", err)
 		return err
 	}
 
-	fmt.Printf("Successfully imported %v issues from Github\n", iterator.ImportedIssues())
+	fmt.Printf("Successfully imported %d issues from Github\n", gi.importedIssues)
```
MichaelMure: What about a unified way to phrase that ? `Successfully imported %d issues and %s identities from Github` ?

bridge/github/import.go
```diff
@@ -23,61 +23,46 @@ const (
...
-				err := gi.ensureTimelineComment(repo, b, item.IssueComment, commentEdits)
-				if err != nil {
-					return fmt.Errorf("timeline comment creation: %v", err)
-				}
-
-			} else {
-				if err := gi.ensureTimelineItem(repo, b, item); err != nil {
-					return fmt.Errorf("timeline event creation: %v", err)
-				}
+		for gi.iterator.NextTimeline() {
```
MichaelMure: Much cleaner !

What about `NextTimelineItem()` ?

bridge/github/iterator.go
MichaelMure: What about dropping entirely this variable and initializing your variables in the constructor ?

A-Hilaly: Much more better ! I didn't think about this at first sight

bridge/github/import.go
A-Hilaly: That's true, i changed the name to `ResetCounters`. Also added some comments to all the non commented public functions.

bridge/github/import.go
A-Hilaly: Done !

bridge/github/import.go
A-Hilaly: Actually i don't know if we really need the `ResetCounters` function here, the only use case i can imagine would be calling `ImportAll` multiple times (with different repositories ?)

MichaelMure: `ResetCounters` doesn't make sense if you can't also reset the whole iterator and start from zero. But it's easier and cheap to just create a new one, so why bother ?

A-Hilaly: Agreed, It would be better to just recreate a new one

PR #153, opened 2019-05-25
[Breaking] Bridge configuration enhancements

A-Hilaly:
Changes to the bridge configuration process:
- Introduce new flags to simplify the configuration: `--owner`, `--project`, `--url` and `--token`
- Verify the project after configuration (Launchpad)
- Verify the owner, project and token after configuration (Github)
- Support token generation for private repositories (Github)
- Support providing a token directly in terminal prompt (Github)
- Use `git remote -v` to suggest repositories URLs (Github)
- Add some tests for configuration process (Launchpad and Github) 
- Update documentation, usage, examples and man pages

Github configuration example:

```bash
git bug bridge configure \
    --name=default \
    --target=github \
    --owner=$(OWNER) \
    --project=$(PROJECT) \
    --token=$(TOKEN)

# You can also pass --url instead of --owner and --project flags
# If the 3 are provided --owner and --project will be used.
```

Launchpad bridge configuration:
```bash
git bug bridge configure \
    --name=default \
    --target=launchpad-preview \
    --url=https://bugs.launchpad.net/ubuntu/
```

Configuration using terminal prompt and a token:
```bash
[1]: github
[2]: launchpad-preview
target: 1
name [default]: default

Detected projects:
[1]: github.com/a-hilaly/git-bug
[2]: github.com/MichaelMure/git-bug

[0]: Another project

Select option: 1

[1]: user provided token
[2]: interactive token creation
Select option: 0

You can generate a new token by visiting https://github.com/settings/tokens.
Choose 'Generate new token' and set the necessary access scope for your repository.

The access scope depend on the type of repository.
Public:
	- 'public_repo': to be able to read public repositories
Private:
	- 'repo'       : to be able to read private repositories

Enter token: 
Successfully configured bridge: default
```


bridge/github/config.go
```diff
@@ -184,6 +192,130 @@ func randomFingerprint() string {
...
+			return string(byteToken), nil
+		}
+
+		fmt.Println("token is empty")
+	}
+}
+
+func loginAndRequestToken(owner, project string) (string, error) {
+	fmt.Println("git-bug will now generate an access token in your Github profile. Your credential are not stored and are only used to generate the token. The token is stored in the repository git config.")
+	fmt.Println()
```
MichaelMure:
```
git-bug will now generate an access token in your Github profile. Your credential are not stored and are only used to generate the token. The token is stored in the repository git config.

The access scope depend on the type of repository.
Public:
  - 'user:email': to be able to read public-only users email
Private:
  - 'user:email': to be able to read public-only users email
  - 'repo'      : to be able to read private repositories

```

bridge/github/config.go
```diff
@@ -11,100 +11,108 @@ import (
...
 	githubV3Url = "https://api.github.com"
-	keyUser     = "user"
+	keyOwner    = "owner"
 	keyProject  = "project"
 	keyToken    = "token"
-)
 
-func (*Github) Configure(repo repository.RepoCommon) (core.Configuration, error) {
-	conf := make(core.Configuration)
+	defaultTimeout = 5 * time.Second
```
MichaelMure: Not everyone has a broadband access. Let's make it 60seconds.

bridge/github/config.go
```diff
@@ -184,6 +192,130 @@ func randomFingerprint() string {
...
+		resp, err = requestTokenWith2FA(note, username, password, otpCode, scope)
+		if err != nil {
+			return "", err
+		}
+
+		defer resp.Body.Close()
+	}
+
+	if resp.StatusCode == http.StatusCreated {
+		token, err := decodeBody(resp.Body)
```
MichaelMure: I know I wrote that, but `return decodeBody(resp.Body)` is enough here ;)

bridge/github/config.go
```diff
@@ -184,6 +192,130 @@ func randomFingerprint() string {
...
+		return "", err
+	}
+
+	password, err := promptPassword()
+	if err != nil {
+		return "", err
+	}
+
+	var scope string
+	if isPublic {
```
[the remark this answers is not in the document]
MichaelMure: also, don't you need `scope = "user:email"` for private repo as well ?

bridge/github/config.go
```diff
@@ -217,36 +387,42 @@ func promptURL() (string, string, error) {
...
-		return projectUser, projectName, nil
+		return owner, project, nil
 	}
 }
 
-func splitURL(url string) (string, string, error) {
-	re, err := regexp.Compile(`github\.com\/([^\/]*)\/([^\/]*)`)
-	if err != nil {
-		panic(err)
+func splitURL(url string) (string, string, string, error) {
```
MichaelMure: With that much return values, it might be time to use golang's named return value

bridge/github/config.go
```diff
@@ -217,36 +387,42 @@ func promptURL() (string, string, error) {
...
+		return "", "", "", fmt.Errorf("bad github project url")
 	}
 
-	res := re.FindStringSubmatch(url)
+	return res[0], res[1], res[2], nil
+}
 
-	if res == nil {
-		return "", "", fmt.Errorf("bad github project url")
+func valideGithubURLRemotes(remotes map[string]string) []string {
```
MichaelMure: a `validateXXX` function would return a bool or an error, but you are returning an array. A better name is needed!

bridge/github/config.go
```diff
@@ -207,7 +339,45 @@ func promptUsername() (string, error) {
...
-func promptURL() (string, string, error) {
+func promptURL(remotes map[string]string) (string, string, error) {
+	validRemotes := valideGithubURLRemotes(remotes)
+	if len(validRemotes) > 0 {
+		for {
+			fmt.Println("\nDetected projects:")
+
+			// print valid remote github urls
+			for i, remote := range validRemotes {
+				fmt.Printf("[%d]: %v\n", i+1, remote)
```
MichaelMure: a nice touch would be to print the remote as the familiar github URL. That is, instead of printing `github.com:MichaelMure/git-bug.git`, having `github.com/MichaelMure/git-bug`

commands/bridge_configure.go
```diff
@@ -98,4 +113,10 @@ var bridgeConfigureCmd = &cobra.Command{
```
MichaelMure: Ideally, the `bridge configure` command would have a `Long` description that explain shortly the process (flags or automated process, token or login, ..) and which flags are valid for which bridge. Kinda what you wrote to present the PR.

commands/bridge_configure.go
```diff
@@ -98,4 +113,10 @@ var bridgeConfigureCmd = &cobra.Command{
 
 func init() {
 	bridgeCmd.AddCommand(bridgeConfigureCmd)
+	bridgeConfigureCmd.Flags().StringVarP(&name, "name", "n", "", "Bridge name")
```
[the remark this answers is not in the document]
A-Hilaly: Agreed ! Working on this

bridge/github/config.go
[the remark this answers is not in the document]
A-Hilaly: the `scope = "user:email"` isn't needed to read private repositories.
I will investigate the scopes needed for making comments/editions

bridge/github/iterator.go
```diff
@@ -70,22 +70,22 @@ func NewIterator(user, project, token string, since time.Time) *iterator {
 			issueEdit:   indexer{-1},
 			commentEdit: indexer{-1},
 			variables: map[string]interface{}{
-				"owner": githubv4.String(user),
-				"name":  githubv4.String(project),
+				keyOwner: githubv4.String(owner),
```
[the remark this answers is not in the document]
A-Hilaly: That is true ! I changed it.

bridge/github/config.go
```diff
@@ -291,14 +495,48 @@ func prompt2FA() (string, error) {
...
 
 		byte2fa, err := terminal.ReadPassword(int(syscall.Stdin))
+		fmt.Println()
 		if err != nil {
 			return "", err
 		}
 
-		if len(byte2fa) > 0 {
-			return string(byte2fa), nil
+		if len(byte2fa) != 6 {
```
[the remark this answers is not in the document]
A-Hilaly: I did this because when i tried to test the changes i miss typed the 2FA code and had to do everything again 😆. ( I thought that all 2FA methods are officially 6digits, at least for SMS/and Google authenticator apps) 
Do you think we should give users a second and third chance to type the 2FA code ?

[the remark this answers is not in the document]
MichaelMure: Depending on what make sense in the code, you can either repeat the 2FA process or the whole auth process. That said, it's a nice thing to have but remember that it's a process that will be used about once per user. It's acceptable if it's not perfect :-)

commands/bridge_configure.go
```diff
@@ -98,4 +113,10 @@ var bridgeConfigureCmd = &cobra.Command{
...
+	bridgeConfigureCmd.Flags().StringVarP(&target, "target", "t", "", "Bridge target name. Valid values are [github,gitlab,gitea,launchpad-preview]")
+	bridgeConfigureCmd.Flags().StringVarP(&bridgeParams.URL, "url", "u", "", "Repository url")
+	bridgeConfigureCmd.Flags().StringVarP(&bridgeParams.Owner, "owner", "o", "", "Repository owner")
+	bridgeConfigureCmd.Flags().StringVarP(&bridgeParams.Token, "token", "T", "", "Authentication token")
+	bridgeConfigureCmd.Flags().StringVarP(&bridgeParams.Project, "project", "p", "", "Repository name")
 }
```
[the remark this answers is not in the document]
A-Hilaly: Awesome 👍 didn't know about this one

bridge/github/config.go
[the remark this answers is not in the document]
A-Hilaly: It is true users will have to do this only once. Let's keep it this way then 👍

bridge/github/config.go
A-Hilaly: Changed the name to `getValidateGithubRemoteURLs`

bridge/github/config.go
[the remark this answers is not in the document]
A-Hilaly: To add/edit comment to a public repository `scope = "repo:public_repo"` is enough (no need to add the `scope = "user:email"` )
For private repositories the full `scope = "repo"` is needed

bridge/github/config.go
```diff
@@ -184,6 +192,130 @@ func randomFingerprint() string {
...
+
+		line = strings.TrimRight(line, "\n")
+
+		index, err := strconv.Atoi(line)
+		if err != nil || (index != 0 && index != 1) {
+			fmt.Println("invalid input")
+			continue
+		}
+
+		if index == 0 {
```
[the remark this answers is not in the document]
A-Hilaly: Done !

bridge/github/config.go
A-Hilaly: Done. I changed the regex and used `strings.TrimSuffix` to remove the `.git` part

commands/bridge_configure.go
A-Hilaly: Done

commands/bridge_configure.go
```diff
@@ -90,12 +105,38 @@ func promptName() (string, error) {
...
 
 var bridgeConfigureCmd = &cobra.Command{
-	Use:     "configure",
-	Short:   "Configure a new bridge.",
+	Use:   "configure",
+	Short: "Configure a new bridge.",
+	Long: `	Configure a new bridge by passing flags or/and using interactive terminal prompts. You can avoid all the terminal prompts by passing all the necessary flags to configure your bridge.
+	Repository configuration can be made by passing or the --url flag or the --project and/or --owner flags. If the three flags are provided git-bug will use --project and --owner flags.
```
MichaelMure:
```suggestion
	Repository configuration can be made by passing either the --url flag or the --project and --owner flags. If the three flags are provided git-bug will use --project and --owner flags.
```
Both --project and --owner are required if --url is not given, correct ?

commands/bridge_configure.go
```diff
@@ -90,12 +105,38 @@ func promptName() (string, error) {
...
+	Token configuration can be made by passing it in the --token flag or in the terminal prompt. If you don't already have one you can use terminal prompt to login and generate it directly.
```
MichaelMure:
```suggestion
	Token configuration can be directly passed with the --token flag or in the terminal prompt. If you don't already have one you can use the interactive procedure to generate one.
```

commands/bridge_configure.go
```diff
@@ -90,12 +105,38 @@ func promptName() (string, error) {
...
+	For Github and Gitlab bridges, git-bug need a token to export and import issues, comments and editions for public and private repositories.
+	For Launchpad bridges, git-bug for now supports only public repositories and you only need --project or --url flag to configure it.`,
+	Example: `# For Github
```
MichaelMure: Let's add at the beginning the simple `git bug bridge configure` for the interactive process.

commands/bridge_configure.go
```diff
@@ -90,12 +105,38 @@ func promptName() (string, error) {
...
 
 var bridgeConfigureCmd = &cobra.Command{
-	Use:     "configure",
-	Short:   "Configure a new bridge.",
+	Use:   "configure",
+	Short: "Configure a new bridge.",
+	Long: `	Configure a new bridge by passing flags or/and using interactive terminal prompts. You can avoid all the terminal prompts by passing all the necessary flags to configure your bridge.
+	Repository configuration can be made by passing or the --url flag or the --project and/or --owner flags. If the three flags are provided git-bug will use --project and --owner flags.
+	Token configuration can be made by passing it in the --token flag or in the terminal prompt. If you don't already have one you can use terminal prompt to login and generate it directly.
+	For Github and Gitlab bridges, git-bug need a token to export and import issues, comments and editions for public and private repositories.
```
MichaelMure:
```suggestion
```

Actually, the examples should be enough to explain what flags are valid/useful.

MichaelMure: And there is no Gitlab bridge ... yet ;)

commands/bridge_configure.go
```diff
@@ -90,12 +105,38 @@ func promptName() (string, error) {
...
+	For Launchpad bridges, git-bug for now supports only public repositories and you only need --project or --url flag to configure it.`,
```
MichaelMure:
```suggestion
```

MichaelMure:
Almost there. You should think a bit about what part of your code is the more likely to have a bug or to break in the future, and try to have a test to prevent that. There is a few easy win that would go a long way. As an example, the github API is out of our control, it would be nice to know if it changed somehow. You could have a short test to run ` validateProject("MichaelMure/git-bug")` and see if that fail.

commands/bridge_configure.go
```diff
@@ -7,12 +7,24 @@ import (
...
 	"github.com/MichaelMure/git-bug/util/interrupt"
 	"github.com/spf13/cobra"
 )
 
+const (
+	defaultName = "default"
+)
+
+var (
+	name         string
```
MichaelMure: those variables are in the same namespace as all commands, they need to be prefixed with the command's name

bridge/github/config.go
```diff
@@ -217,36 +393,43 @@ func promptURL() (string, string, error) {
...
-		panic(err)
+func splitURL(url string) (shortURL string, owner string, project string, err error) {
+	cleanURL := strings.TrimSuffix(url, ".git")
+	res := rxGithubURL.FindStringSubmatch(cleanURL)
+	if res == nil {
+		return "", "", "", fmt.Errorf("bad github project url")
 	}
 
-	res := re.FindStringSubmatch(url)
+	return res[0], res[1], res[2], nil
```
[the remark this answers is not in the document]
A-Hilaly: Done !

commands/bridge_configure.go
A-Hilaly: Done.

commands/bridge_configure.go
A-Hilaly: Agreed ! changed the variables to `bridgeConfigureName` and `bridgeConfigureTarget`

bridge/github/config_test.go
```diff
@@ -0,0 +1,146 @@
...
+			owner, project, err := splitURL(tt.args.url)
+			assert.Equal(t, tt.want.err, err)
+			assert.Equal(t, tt.want.owner, owner)
+			assert.Equal(t, tt.want.project, project)
+		})
+	}
+}
+
+func TestValidateProject(t *testing.T) {
+	tokenPrivateScope := os.Getenv("GITHUB_TOKEN_PRIVATE")
```
A-Hilaly: I'm not sure about these environment variable names. We already use one with name `GITHUB_TOKEN`in `import_test.go`, should we rename it to `GITHUB_TOKEN_PRIVATE` too ?

bridge/github/config.go
```diff
@@ -11,100 +11,108 @@ import (
...
 
-	fmt.Println()
-	fmt.Println("git-bug will now generate an access token in your Github profile. Your credential are not stored and are only used to generate the token. The token is stored in the repository git config.")
-	fmt.Println()
-	fmt.Println("The token will have the following scopes:")
-	fmt.Println("  - user:email: to be able to read public-only users email")
-	// fmt.Println("The token will have the \"repo\" permission, giving it read/write access to your repositories and issues. There is no narrower scope available, sorry :-|")
-	fmt.Println()
+var (
+	rxGithubURL = regexp.MustCompile(`github\.com[\/:]([a-zA-Z0-9\-\_]+)\/([a-zA-Z0-9\-\_\.]+)`)
```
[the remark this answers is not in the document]
A-Hilaly: Done ! I verified this. It is true it will compile the regex at every `git bug` related or not the bridge.

bridge/launchpad/config.go
```diff
@@ -3,25 +3,65 @@ package launchpad
...
 )
 
-const keyProject = "project"
+const (
+	keyProject     = "project"
+	defaultTimeout = 60 * time.Second
+)
+
+var (
+	rxLaunchpadURL = regexp.MustCompile(`launchpad\.net[\/:]([^\/]*[a-z]+)`)
```
MichaelMure: :-O

A-Hilaly: Oops ! will fix that and add necessary tests for launchpad bridge config

A-Hilaly: Done !

bridge/launchpad/config.go
```diff
@@ -52,3 +91,33 @@ func promptProjectName() (string, error) {
...
+	}
+
+	return resp.StatusCode == http.StatusOK, nil
+}
+
+// extract project name from url
+func splitURL(url string) (string, error) {
+	re, err := regexp.Compile(`launchpad\.net[\/:]([^\/]*[a-z]+)`)
+	if err != nil {
+		panic("regexp compile:" + err.Error())
```
A-Hilaly: Since this error should never happen (clearly a development error) i thought maybe it is better to panic here

commands/bridge_configure.go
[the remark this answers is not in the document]
A-Hilaly: Opened #157 make changes in another PR

A-Hilaly:
Something really weird happen with travis in this PR. since i added tests for `validateUsername` in the `github` package tests fail from time to time (i juste keep restarting them until they pass). In my local environment tests never fail 🤔

bridge/github/config.go
```diff
@@ -184,6 +194,135 @@ func randomFingerprint() string {
...
+	fmt.Println("Public:")
+	fmt.Println("  - 'public_repo': to be able to read public repositories")
+	fmt.Println("Private:")
+	fmt.Println("  - 'repo'       : to be able to read private repositories")
+	fmt.Println()
+
+	for {
+		fmt.Print("Enter token: ")
+
+		byteToken, err := terminal.ReadPassword(int(syscall.Stdin))
```
MichaelMure: Trying to use it, I don't think masking the input is the way to go for the token. It's not a password that the user is used to type, it's not even something the user will type. What you want to to be able to check that it was pasted properly, that there is no extra spaces or weird character. I think it needs to be visible.

commands/bridge_configure.go
```diff
@@ -90,12 +105,66 @@ func promptName() (string, error) {
...
 }
 
 func init() {
 	bridgeCmd.AddCommand(bridgeConfigureCmd)
+	bridgeConfigureCmd.Flags().StringVarP(&bridgeConfigureName, "name", "n", "", "A distinctive name to identify the bridge")
+	bridgeConfigureCmd.Flags().StringVarP(&bridgeConfigureTarget, "target", "t", "",
+		fmt.Sprintf("The target of the bridge. Valid values are [%s]", strings.Join(bridge.Targets(), ",")))
+	bridgeConfigureCmd.Flags().StringVarP(&bridgeParams.URL, "url", "u", "", "The URL of the target repository")
+	bridgeConfigureCmd.Flags().StringVarP(&bridgeParams.Owner, "owner", "o", "", "The owner of the target repository")
+	bridgeConfigureCmd.Flags().StringVarP(&bridgeParams.Token, "token", "T", "", "The authentication token for the API")
```
MichaelMure: The problem in having a flag to pass the token is that it will stay in the shell history, which is a pretty bad idea. We at least need second a `--token-stdin` flag, if not entirely removing the `--token` flag.

bridge/github/config.go
```diff
@@ -184,6 +194,135 @@ func randomFingerprint() string {
 	return string(b)
 }
 
+func promptTokenOptions(owner, project string) (string, error) {
+	for {
+		fmt.Println()
+		fmt.Println("[0]: user provided token")
```
MichaelMure: All the other prompt starts at 1, but this one start at 0.

commands/bridge_configure.go
```diff
@@ -21,26 +33,31 @@ func runBridgeConfigure(cmd *cobra.Command, args []string) error {
 	defer backend.Close()
 	interrupt.RegisterCleaner(backend.Close)
```
MichaelMure:
```suggestion

	termState, err := terminal.GetState(int(syscall.Stdin))
	if err != nil {
		return err
	}
	interrupt.RegisterCleaner(func() error {
		return terminal.Restore(int(syscall.Stdin), termState)
	})
```

I did a ctrl+c when the configure was asking for my token (invisible input) and it broke my terminal. This fix the problem.

bridge/github/config.go
```diff
@@ -217,42 +394,93 @@ func promptURL() (string, string, error) {
...
-
+	res := re.FindStringSubmatch(cleanURL)
 	if res == nil {
-		return "", "", fmt.Errorf("bad github project url")
+		return "", "", ErrBadProjectURL
+	}
+
+	owner = res[1]
+	project = res[2]
+	return owner, project, nil
```
MichaelMure:
```suggestion
	return
```

MichaelMure: or remove the two lines above

commands/bridge_configure.go
```diff
@@ -21,26 +33,31 @@ func runBridgeConfigure(cmd *cobra.Command, args []string) error {
...
 
-	target, err := promptTarget()
-	if err != nil {
-		return err
+	if bridgeConfigureTarget == "" {
+		bridgeConfigureTarget, err = promptTarget()
```
MichaelMure: not your code, but I just found out that there is an annoying thing here: each bridge implementation are registered with a `init()` function and the order is not deterministic :(
One run will have the github one first, another will have the launchpad one first.
We need a way to make that consistent.

MichaelMure: this can be dealt with in another PR btw

commands/bridge_configure.go
A-Hilaly: 👍  Didn't know about this !

bridge/github/config.go
A-Hilaly: I hesitated before writing this, agreed checking the token and making it visible is more suitable for this situation

bridge/github/config.go
A-Hilaly: Done

commands/bridge_configure.go
A-Hilaly: Agreed having a different order every time we want to configure is a little weird. Will be dealing with this ASAP ;)

commands/bridge_configure.go
A-Hilaly: Agreed this is a great feature to implement

bridge/github/config_test.go
MichaelMure: yes, let keep all that consistent, we only need two kind of tokens

A-Hilaly: Done !

commands/bridge_configure.go
A-Hilaly: Opened #164 for this one

PR #156, opened 2019-05-30
Fix bridge rm command

A-Hilaly:
- Add `RmConfigSection` to `RepoCommon` interface
- Add tests cases for `RmConfigSection`
- Fix `git bug bridge rm` expected behaviour by using `RmConfigSection`

### DIFF
--- a/commands/bridge_rm.go
+++ b/commands/bridge_rm.go
@@ -23,7 +23,7 @@ func runBridgeRm(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	fmt.Printf("Successfully deleted bridge %v", args[0])
+	fmt.Printf("Successfully removed bridge configuration %v", args[0])
 	return nil
 }
 
--- a/commands/bridge_rm.go
+++ b/commands/bridge_rm.go
@@ -1,10 +1,13 @@
 package commands
 
 import (
+	"fmt"
+
+	"github.com/spf13/cobra"
+
 	"github.com/MichaelMure/git-bug/bridge"
 	"github.com/MichaelMure/git-bug/cache"
 	"github.com/MichaelMure/git-bug/util/interrupt"
-	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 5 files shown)


repository/mock_repo.go
```diff
@@ -96,15 +96,22 @@ func (r *mockRepoForTest) ReadConfigString(key string) (string, error) {
 	return val, nil
 }
 
-func (r *mockRepoForTest) RmConfigs(keyPrefix string) error {
+// RmConfig remove all key/value pair matching the key prefix
+func (r *mockRepoForTest) RmConfig(k string) error {
 	for key := range r.config {
```
A-Hilaly: Would it be better to mock `mockRepoForTest.config` as `map[string]map[string]string`. That's the only way i see to create mocks for `RmConfigSection`

commands/bridge_rm.go
```diff
@@ -20,11 +23,12 @@ func runBridgeRm(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	fmt.Printf("Successfully deleted bridge %v", args[0])
```
MichaelMure:
```suggestion
	fmt.Printf("Successfully removed bridge configuration %v", args[0])
```

repository/mock_repo.go
MichaelMure: yes !

repository/mock_repo.go
```diff
@@ -103,6 +103,7 @@ func (r *mockRepoForTest) ReadConfigString(key string) (string, error) {
 	return val, nil
 }
 
+// RmConfig remove all key/value pair matching the key prefix
```
MichaelMure:
```suggestion
// RmConfigs remove all key/value pair matching the key prefix
```

MichaelMure:
Are we good with this one or do you want to make more changes ?

A-Hilaly:
Yep this PR is ready ! I rebased and added more test cases for `RmConfigs`

Issue #161, opened 2019-06-05
Consistent bridge implementations list


A-Hilaly:
As [discussed](https://github.com/MichaelMure/git-bug/pull/153#discussion_r290089526) in #153 , for now bridges are not listed in a ordered way and we need a way to make  consistent. 
at each run, `git bug bridge configure` should suggest the supported bridges in the same order.

PR #162, opened 2019-06-05
Consistent bridge implementations list

A-Hilaly:
Dealing with #161


MichaelMure:
Yeah, that's probably the best option for now.

Issue #157, opened 2019-05-30
Store bridge type alongside the other params in the config


A-Hilaly:
[As discussed](https://github.com/MichaelMure/git-bug/pull/153#discussion_r287577822) in #153

Instead of adding the bridge type as prefix to the final name of the bridge, use the real given name and store the bridge type alongside the other bridge params in the config.

example:
```
[git-bug "bridge.default"]
        target = github
        user = MichaelMure
        project = git-bug
        token = MYTOKEN
```

A-Hilaly:
Fixed in #170

PR #170, opened 2019-06-15
[bridge] Change bridge name format in the config

A-Hilaly:
dealing with #157

### DIFF
--- a/bridge/core/bridge.go
+++ b/bridge/core/bridge.go
@@ -236,7 +236,6 @@ func (b *Bridge) loadConfig() (Configuration, error) {
 	result := make(Configuration, len(pairs))
 	for key, value := range pairs {
 		key := strings.TrimPrefix(key, keyPrefix)
-		fmt.Println(key, value)
 		result[key] = value
 	}
 
--- a/bridge/core/bridge.go
+++ b/bridge/core/bridge.go
@@ -87,30 +87,24 @@ func NewBridge(repo *cache.RepoCache, target string, name string) (*Bridge, erro
 
 // LoadBridge instantiate a new bridge from a repo configuration
 func LoadBridge(repo *cache.RepoCache, name string) (*Bridge, error) {
-	bridge := &Bridge{
-		Name: name,
-		repo: repo,
-	}
-
-	conf, err := bridge.loadConfig()
+	conf, err := loadConfig(repo, name)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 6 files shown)


bridge/core/bridge.go
```diff
@@ -225,14 +236,10 @@ func (b *Bridge) loadConfig() (Configuration, error) {
 	result := make(Configuration, len(pairs))
 	for key, value := range pairs {
 		key := strings.TrimPrefix(key, keyPrefix)
+		fmt.Println(key, value)
```
MichaelMure:
```suggestion
```

bridge/core/bridge.go
```diff
@@ -81,15 +85,33 @@ func NewBridge(repo *cache.RepoCache, target string, name string) (*Bridge, erro
...
+		repo: repo,
+	}
+
+	conf, err := bridge.loadConfig()
 	if err != nil {
 		return nil, err
 	}
+	bridge.conf = conf
+
+	target := bridge.conf[keyTarget]
```
MichaelMure: That's pretty much the implementation of `NewBridge()`, so why the need to copy it instead of using it ? If you want to change the API to have LoadBridge instead, you should remove NewBridge() or at least deprecate it.

bridge/github/config.go
```diff
@@ -109,6 +111,10 @@ func (*Github) Configure(repo repository.RepoCommon, params core.BridgeParams) (
 }
 
 func (*Github) ValidateConfig(conf core.Configuration) error {
+	if _, ok := conf[keyTarget]; !ok {
```
MichaelMure: you can check that the target is indeed `github`

bridge/core/bridge.go
A-Hilaly: The problem i encountered while writing this was that `LoadBridge` need to read the configuration before instantiating a bridge object, and since `loadConfig()` is a method of `Bridge` i was forced to do this.
`NewBridge` is used by `git bug bridge configure` who already knows the target name and want to instantiate a bridge directly. 

What do you think of making `loadConfig` a more general function: something like `loadConfig(repo *cache.Repo, name string)` and then call `NewBridge` from  `LoadBridge` to avoid code repetition ?

MichaelMure: sounds good to me

A-Hilaly: Done !

PR #186, opened 2019-07-26
bridge/gitlab: handle all known notes types

### LINKED ISSUE #185: Gitlab importation - unhandled note type
When trying to import issues from Gitlab, it starts and then:

```
panic: unhandled note type

goroutine 1 [running]:
github.com/MichaelMure/git-bug/bridge/gitlab.(*gitlabImporter).ensureNote(0xc420290080, 0xc4201edfc0, 0xc4201a8aa0, 0xc42058a180, 0x0, 0x0)
```

It still successfully imported some bugs. :)

The failing bug may be titled with some accents, quotes and parentheses, I don't know if this could be the cause of the crash.

A-Hilaly:
fix #185

### DIFF
--- a/bridge/gitlab/import.go
+++ b/bridge/gitlab/import.go
@@ -248,7 +248,15 @@ func (gi *gitlabImporter) ensureNote(repo *cache.RepoCache, b *cache.BugCache, n
 
 		return err
 
-	case NOTE_UNKNOWN:
+	case NOTE_UNKNOWN,
+		NOTE_ASSIGNED,
+		NOTE_UNASSIGNED,
+		NOTE_CHANGED_MILESTONE,
+		NOTE_REMOVED_MILESTONE,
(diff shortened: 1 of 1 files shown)


bridge/gitlab/import.go
```diff
@@ -248,7 +248,14 @@ func (gi *gitlabImporter) ensureNote(repo *cache.RepoCache, b *cache.BugCache, n
 
 		return err
 
-	case NOTE_UNKNOWN:
```
MichaelMure: `NOTE_UNKNOWN` would lead to a panic

A-Hilaly: Yes agreed that's not a reason to panic. Added `NOTE_UNKNOWN` to the cases

cgivord:
This patch got me through a successful importation of all my bugs ! :)
Thank you both for this, awesome job !!

PR #190, opened 2019-07-31
Bridge: pull/push enhancements

### LINKED ISSUE #175: feat: graceful kill for the bridge export and import functionalities
for now interrupting git-bug while exporting or importing issues can damage local or remote and cause problems if when we re-execute these commands.
To avoid this we can implement a global context to gracefully shutdown these functions ! for this we need:

- [ ] To use a global context for the bridge importer and exporter
- [ ] use the context to cancel http requests
- [ ] Add cancel functions to Interrupt handler
- [ ] Commit operations before exiting the functions (SetMetadataOperation in the exporter for example `)

(this bug was created using the new exporter)

A-Hilaly:
- [x] changes for the `Importer` and `Exporter` interfaces
- [x] Add ImportEvents functionalities to `core` package
- [x] Use a global context for the bridge importer and exporter
- [x] gracefull kill on bridge pull/push commands 
- [x] improve bridge pull/push events logging
- [x] analyse the events to print statistics at the end of pull/push 
fix #175

### DIFF
--- a/bridge/core/import.go
+++ b/bridge/core/import.go
@@ -49,9 +49,9 @@ func (er ImportResult) String() string {
 		return fmt.Sprintf("new identity: %s", er.ID)
 	case ImportEventNothing:
 		if er.ID != "" {
-			return fmt.Sprintf("ignoring import event %s: %s", er.ID, er.Reason)
+			return fmt.Sprintf("no action taken for event %s: %s", er.ID, er.Reason)
 		}
-		return fmt.Sprintf("ignoring event: %s", er.Reason)
+		return fmt.Sprintf("no action taken: %s", er.Reason)
 	case ImportEventError:
 		if er.ID != "" {
 			return fmt.Sprintf("import error at id %s: %s", er.ID, er.Err.Error())
--- a/bridge/core/import.go
+++ b/bridge/core/import.go
@@ -1,0 +1,128 @@
+package core
+
+import (
+	"fmt"
+
+	"github.com/MichaelMure/git-bug/entity"
+)
+
+type ImportEvent int
+
+const (
+	_ ImportEvent = iota
+	ImportEventBug
+	ImportEventComment
+	ImportEventCommentEdition
+	ImportEventStatusChange
+	ImportEventTitleEdition
+	ImportEventLabelChange
+	ImportEventIdentity
+	ImportEventNothing
+	ImportEventError
+)
+
+// ImportResult is an event that is emitted during the import process, to
+// allow calling code to report on what is happening, collect metrics or
+// display meaningful errors if something went wrong.
+type ImportResult struct {
+	Err    error
+	Event  ImportEvent
+	ID     entity.Id
+	Reason string
+}
+
+func (er ImportResult) String() string {
+	switch er.Event {
+	case ImportEventBug:
+		return fmt.Sprintf("new issue: %s", er.ID)
+	case ImportEventComment:
+		return fmt.Sprintf("new comment: %s", er.ID)
+	case ImportEventCommentEdition:
+		return fmt.Sprintf("updated comment: %s", er.ID)
+	case ImportEventStatusChange:
+		return fmt.Sprintf("changed status: %s", er.ID)
+	case ImportEventTitleEdition:
+		return fmt.Sprintf("changed title: %s", er.ID)
+	case ImportEventLabelChange:
+		return fmt.Sprintf("changed label: %s", er.ID)
+	case ImportEventIdentity:
+		return fmt.Sprintf("new identity: %s", er.ID)
+	case ImportEventNothing:
+		if er.ID != "" {
+			return fmt.Sprintf("ignoring import event %s: %s", er.ID, er.Reason)
+		}
+		return fmt.Sprintf("ignoring event: %s", er.Reason)
+	case ImportEventError:
+		if er.ID != "" {
+			return fmt.Sprintf("import error at id %s: %s", er.ID, er.Err.Error())
+		}
+		return fmt.Sprintf("import error: %s", er.Err.Error())
+	default:
+		panic("unknown import result")
+	}
+}
+
+func NewImportError(err error, id entity.Id) ImportResult {
+	return ImportResult{
+		Err:   err,
+		ID:    id,
+		Event: ImportEventError,
+	}
+}
+
+func NewImportNothing(id entity.Id, reason string) ImportResult {
+	return ImportResult{
+		ID:     id,
+		Reason: reason,
+		Event:  ImportEventNothing,
+	}
+}
+
+func NewImportBug(id entity.Id) ImportResult {
+	return ImportResult{
+		ID:    id,
+		Event: ImportEventBug,
+	}
+}
+
+func NewImportComment(id entity.Id) ImportResult {
+	return ImportResult{
+		ID:    id,
+		Event: ImportEventComment,
+	}
+}
+
+func NewImportCommentEdition(id entity.Id) ImportResult {
+	return ImportResult{
+		ID:    id,
+		Event: ImportEventCommentEdition,
+	}
+}
+
+func NewImportStatusChange(id entity.Id) ImportResult {
+	return ImportResult{
+		ID:    id,
+		Event: ImportEventStatusChange,
+	}
+}
+
+func NewImportLabelChange(id entity.Id) ImportResult {
+	return ImportResult{
+		ID:    id,
+		Event: ImportEventLabelChange,
+	}
+}
+
+func NewImportTitleEdition(id entity.Id) ImportResult {
+	return ImportResult{
+		ID:    id,
+		Event: ImportEventTitleEdition,
+	}
+}
+
+func NewImportIdentity(id entity.Id) ImportResult {
+	return ImportResult{
+		ID:    id,
+		Event: ImportEventIdentity,
+	}
+}
--- a/bridge/core/interfaces.go
+++ b/bridge/core/interfaces.go
@@ -1,6 +1,7 @@
 package core
 
 import (
+	"context"
 	"time"
 
 	"github.com/MichaelMure/git-bug/cache"
@@ -29,10 +30,10 @@ type BridgeImpl interface {
 
 type Importer interface {
 	Init(conf Configuration) error
-	ImportAll(repo *cache.RepoCache, since time.Time) error
+	ImportAll(ctx context.Context, repo *cache.RepoCache, since time.Time) (<-chan ImportResult, error)
 }
 
 type Exporter interface {
 	Init(conf Configuration) error
-	ExportAll(repo *cache.RepoCache, since time.Time) (<-chan ExportResult, error)
+	ExportAll(ctx context.Context, repo *cache.RepoCache, since time.Time) (<-chan ExportResult, error)
 }
--- a/bridge/github/export.go
+++ b/bridge/github/export.go
@@ -79,7 +79,7 @@ func (ge *githubExporter) getIdentityClient(id entity.Id) (*githubv4.Client, err
 }
 
 // ExportAll export all event made by the current user to Github
-func (ge *githubExporter) ExportAll(repo *cache.RepoCache, since time.Time) (<-chan core.ExportResult, error) {
+func (ge *githubExporter) ExportAll(ctx context.Context, repo *cache.RepoCache, since time.Time) (<-chan core.ExportResult, error) {
 	out := make(chan core.ExportResult)
 
 	user, err := repo.GetUserIdentity()
@@ -91,6 +91,7 @@ func (ge *githubExporter) ExportAll(repo *cache.RepoCache, since time.Time) (<-c
 
 	// get repository node id
 	ge.repositoryID, err = getRepositoryNodeID(
+		ctx,
 		ge.conf[keyOwner],
 		ge.conf[keyProject],
 		ge.conf[keyToken],
@@ -117,20 +118,28 @@ func (ge *githubExporter) ExportAll(repo *cache.RepoCache, since time.Time) (<-c
 				return
 			}
 
-			snapshot := b.Snapshot()
+			select {
 
-			// ignore issues created before since date
-			// TODO: compare the Lamport time instead of using the unix time
-			if snapshot.CreatedAt.Before(since) {
-				out <- core.NewExportNothing(b.Id(), "bug created before the since date")
-				continue
-			}
+			case <-ctx.Done():
+				// stop iterating if context cancel function is called
+				return
 
-			if snapshot.HasAnyActor(allIdentitiesIds...) {
-				// try to export the bug and it associated events
-				ge.exportBug(b, since, out)
-			} else {
-				out <- core.NewExportNothing(id, "not an actor")
+			default:
+				snapshot := b.Snapshot()
+
+				// ignore issues created before since date
+				// TODO: compare the Lamport time instead of using the unix time
+				if snapshot.CreatedAt.Before(since) {
+					out <- core.NewExportNothing(b.Id(), "bug created before the since date")
+					continue
+				}
+
+				if snapshot.HasAnyActor(allIdentitiesIds...) {
+					// try to export the bug and it associated events
+					ge.exportBug(ctx, b, since, out)
+				} else {
+					out <- core.NewExportNothing(id, "not an actor")
+				}
 			}
 		}
 	}()
@@ -139,7 +148,7 @@ func (ge *githubExporter) ExportAll(repo *cache.RepoCache, since time.Time) (<-c
 }
 
 // exportBug publish bugs and related events
-func (ge *githubExporter) exportBug(b *cache.BugCache, since time.Time, out chan<- core.ExportResult) {
+func (ge *githubExporter) exportBug(ctx context.Context, b *cache.BugCache, since time.Time, out chan<- core.ExportResult) {
 	snapshot := b.Snapshot()
 
 	var bugGithubID string
@@ -199,7 +208,7 @@ func (ge *githubExporter) exportBug(b *cache.BugCache, since time.Time, out chan
 		}
 
 		// create bug
-		id, url, err := createGithubIssue(client, ge.repositoryID, createOp.Title, createOp.Message)
+		id, url, err := createGithubIssue(ctx, client, ge.repositoryID, createOp.Title, createOp.Message)
 		if err != nil {
 			err := errors.Wrap(err, "exporting github issue")
 			out <- core.NewExportError(err, b.Id())
@@ -257,7 +266,7 @@ func (ge *githubExporter) exportBug(b *cache.BugCache, since time.Time, out chan
 			opr := op.(*bug.AddCommentOperation)
 
 			// send operation to github
-			id, url, err = addCommentGithubIssue(client, bugGithubID, opr.Message)
+			id, url, err = addCommentGithubIssue(ctx, client, bugGithubID, opr.Message)
 			if err != nil {
 				err := errors.Wrap(err, "adding comment")
 				out <- core.NewExportError(err, b.Id())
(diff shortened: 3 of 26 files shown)


A-Hilaly:
This PR need to be rebased after #189

bridge/github/export_test.go
```diff
@@ -183,13 +184,21 @@ func TestPushPull(t *testing.T) {
 	})
 	require.NoError(t, err)
 
+	parentCtx := context.Background()
+	ctx, cancel := context.WithCancel(parentCtx)
+	defer cancel()
+	interrupt.RegisterCleaner(func() error {
```
MichaelMure: Not necessarily a problem, but the repo is *not* removed id the test is killed.

bridge/core/import.go
```diff
@@ -0,0 +1,114 @@
...
+	case ImportEventIdentity:
+		return "new identity"
+	case ImportEventNothing:
+		return fmt.Sprintf("no event: %v", er.Reason)
+	default:
+		panic("unknown import result")
+	}
+}
+
+func NewImportError(err error, reason string) ImportResult {
```
MichaelMure: This is used with an id, while here it's a reason

bridge/github/export.go
```diff
@@ -111,26 +111,33 @@ func (ge *githubExporter) ExportAll(repo *cache.RepoCache, since time.Time) (<-c
 		allBugsIds := repo.AllBugsIds()
 
 		for _, id := range allBugsIds {
-			b, err := repo.ResolveBug(id)
-			if err != nil {
-				out <- core.NewExportError(err, id)
+			select {
+			case <-ctx.Done():
+				out <- core.NewExportError(ctx.Err(), "")
```
MichaelMure: Do as you prefer, but maybe you need a new result type for an interrupted process ?

bridge/github/import.go
```diff
@@ -43,40 +46,60 @@ func (gi *githubImporter) Init(conf core.Configuration) error {
...
 
-		// create issue
-		b, err := gi.ensureIssue(repo, issue)
-		if err != nil {
-			return fmt.Errorf("issue creation: %v", err)
-		}
+		// Loop over all matching issues
+		for gi.iterator.NextIssue() {
+			select {
+			case <-ctx.Done():
```
MichaelMure: So essentially, you allow to stop the process between each bugs. That's a clear improvement but I think it can be done better. 

What you could do is:
- pass the context down up to the functions doing the actual network calls (which means those will actually be cancelled instead of waiting). When that happen, everything will return early with (I suppose) an error (the one given by the context).
- in this function, check the error. If it's a context cancelled error, you can safely commit the bug to keep what was successfully imported before the cancel.
- in this function,

MichaelMure: Note: this is because you don't need to cancel anything done locally. It's fast and will take a few milliseconds worst case. What you really need to cancel is everything where you need to really wait or that can hang (so in this case, the network calls).

bridge/core/import.go
```diff
@@ -0,0 +1,114 @@
...
+	case ImportEventTitleEdition:
+		return "changed title"
+	case ImportEventLabelChange:
+		return "changed label"
+	case ImportEventIdentity:
+		return "new identity"
+	case ImportEventNothing:
+		return fmt.Sprintf("no event: %v", er.Reason)
+	default:
+		panic("unknown import result")
```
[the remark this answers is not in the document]
A-Hilaly: > ExportResult actually has the same problem, String() will panic with the result of NewImportError() (that is, a perfectly legal value).

I saw this issue when i tried to use the import functionalities. Handling this case will be so usefull i think !

> `MergeResult` also 

I will make changes in the `entity` package also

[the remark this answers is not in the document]
A-Hilaly: Should i add a new type `ImportEventError` and same for other event types ? 

Since the `NewError` implementation doesn't have any specific type

```golang
func NewImportError(err error, reason string) ImportResult {
	return ImportResult{
                // ImportType is 0 here
		Err:    err,
		Reason: reason,
	}
}
```

bridge/github/export.go
A-Hilaly: Yes that would be nice !

bridge/core/interfaces.go
```diff
@@ -29,10 +30,10 @@ type BridgeImpl interface {
 
 type Importer interface {
 	Init(conf Configuration) error
-	ImportAll(repo *cache.RepoCache, since time.Time) error
```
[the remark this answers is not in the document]
A-Hilaly: Done !

bridge/github/import.go
A-Hilaly: > pass the context down up to the functions doing the actual network calls (which means those will actually be cancelled instead of waiting). When that happen, everything will return early with (I suppose) an error (the one given by the context).

Is it OK to change the `iterator` methods signatures to use `context`  ? We'll have something like `NextIssue(ctx context.Context)`, `NextComment(ctx context.Context)` , `NextTimelineItem(ctx context.Context)`  ...

What if one of these functions doesn't have to make an api call (already cached some items), should they be checking the `ctx.Err()` at the start of each call ?

bridge/core/import.go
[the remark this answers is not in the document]
MichaelMure: Yeah, a new type of event make sense

bridge/github/import.go
MichaelMure: Maybe pass the context when creating the iterator ?

bridge/core/import.go
A-Hilaly: Yes normally it should be an id

commands/bridge_pull.go
```diff
@@ -31,12 +35,46 @@ func runBridgePull(cmd *cobra.Command, args []string) error {
...
+	done := make(chan struct{}, 1)
+
+	interrupt.RegisterCleaner(func() error {
+		fmt.Println("Received interrupt signal, stopping the import...\n(Hit ctrl-c again to kill the process.)")
+		// send signal to stop the importer
+		cancel()
+
+		go func() {
+			c := make(chan os.Signal, 1)
+			signal.Notify(c, os.Kill)
```
golangcibot: SA1016: os.Kill cannot be trapped (did you mean syscall.SIGTERM?) (from `staticcheck`)

bridge/github/import.go
A-Hilaly: Done 👍

bridge/core/import.go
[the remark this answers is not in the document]
A-Hilaly: Done for all export, import and merge EventTypes

bridge/github/export.go
```diff
@@ -117,20 +118,38 @@ func (ge *githubExporter) ExportAll(repo *cache.RepoCache, since time.Time) (<-c
...
 
-			// ignore issues created before since date
-			// TODO: compare the Lamport time instead of using the unix time
-			if snapshot.CreatedAt.Before(since) {
-				out <- core.NewExportNothing(b.Id(), "bug created before the since date")
-				continue
-			}
+			case <-ctx.Done():
+				if ctx.Err() == nil {
+					return
```
MichaelMure: hmmm, doesn't that need to return an error of some sort ? Could you try with a https://golang.org/pkg/context/#WithTimeout to see what happen ?

bridge/github/export.go
```diff
@@ -599,18 +617,14 @@ func (ge *githubExporter) getLabelsIDs(gc *githubv4.Client, repositoryID string,
...
-func createGithubIssue(gc *githubv4.Client, repositoryID, title, body string) (string, string, error) {
+func createGithubIssue(ctx context.Context, gc *githubv4.Client, repositoryID, title, body string) (string, string, error) {
 	m := &createIssueMutation{}
 	input := githubv4.CreateIssueInput{
 		RepositoryID: repositoryID,
 		Title:        githubv4.String(title),
 		Body:         (*githubv4.String)(&body),
 	}
 
-	parentCtx := context.Background()
```
MichaelMure: Unless I'm missing something, you still need the timeouts,

bridge/github/import.go
```diff
@@ -123,14 +152,19 @@ func (gi *githubImporter) ensureIssue(repo *cache.RepoCache, issue issueTimeline
 			}
 
 			// importing a new bug
-			gi.importedIssues++
+			gi.out <- core.NewImportBug(b.Id())
+		} else {
+			reason := fmt.Sprintf("bug already imported")
```
MichaelMure: do you really need `Sprintf` here ? ;)

bridge/core/import.go
```diff
@@ -0,0 +1,118 @@
...
+		return fmt.Sprintf("updated comment: %s", er.ID)
+	case ImportEventStatusChange:
+		return fmt.Sprintf("changed status: %s", er.ID)
+	case ImportEventTitleEdition:
+		return fmt.Sprintf("changed title: %s", er.ID)
+	case ImportEventLabelChange:
+		return fmt.Sprintf("changed label: %s", er.ID)
+	case ImportEventIdentity:
+		return fmt.Sprintf("new identity: %s", er.ID)
+	case ImportEventNothing:
```
MichaelMure: maybe a conditional formatting for when you have an ID as well ?

commands/bridge_pull.go
```diff
@@ -31,12 +35,61 @@ func runBridgePull(cmd *cobra.Command, args []string) error {
...
+	importedIdentities := 0
+	for result := range events {
+		fmt.Println(result.String())
+
+		switch result.Event {
+		case core.ImportEventBug:
+			importedIssues++
+		case core.ImportEventIdentity:
+			importedIdentities++
+		default:
```
MichaelMure:
```suggestion
```

commands/bridge_push.go
```diff
@@ -32,20 +33,46 @@ func runBridgePush(cmd *cobra.Command, args []string) error {
...
 	}
 
+	parentCtx := context.Background()
+	ctx, cancel := context.WithCancel(parentCtx)
+	defer cancel()
+
+	done := make(chan struct{}, 1)
+
+	interrupt.RegisterCleaner(func() error {
+		fmt.Println("Received interrupt signal, stopping the import...\n(Hit ctrl-c again to kill the process.)")
```
MichaelMure: you need to implement the kill here as well

bridge/github/export.go
A-Hilaly: it is true that these errors should be handled properly (i didn't pay attention to that at first sight)
I tried to run some tests locally. Two types of errors are catched when context is cancelled (timeout or manual cancel):
 - `context deadline exceeded` for context timeout
- `context cancelled` when cancel() is called

bridge/github/export.go
A-Hilaly: Agreed, i thought that cancel is enough for this case, which is not correct in case

bridge/github/import.go
A-Hilaly: 10* 🤦‍♂

bridge/core/import.go
A-Hilaly: Done !

commands/bridge_push.go
A-Hilaly: done !

bridge/github/export.go
MichaelMure: Is it really useful to test and return if `ctx.Err() == nil` ? I'm not entirely familiar with `context` but if `ctx.Done()` happen, does it may or have to be an error ?

bridge/github/export.go
MichaelMure: You still need the timeouts.

bridge/github/import.go
MichaelMure: hmmm, if you handle `<-ctx.Done()` in the iterator, you can't have it here as well. Unless I'm mistaken, there is a single value sent to this channel so you essentially have a race condition.

bridge/github/export.go
A-Hilaly: I just checked this. There will be always an error when the context is done. 
there are two possible errors:
- `context deadline exceeded`
- `context cancelled`

Should we consider `context cancelled` as soft error ?

bridge/github/import.go
A-Hilaly: Correct ! it is called twice in the same function

commands/bridge_pull.go
```diff
@@ -31,12 +35,60 @@ func runBridgePull(cmd *cobra.Command, args []string) error {
...
+		mu.Unlock()
+
+		fmt.Println("Received interrupt signal, stopping the import...\n(Hit ctrl-c again to kill the process.)")
+
+		// send signal to stop the importer
+		cancel()
+
+		// block until importer gracefully shutdown
+		<-done
+		close(done)
```
A-Hilaly: Do you think it is safe to close a channel directly after sending in it ? What if this block execute before the main goroutine try to receive ?

bridge/github/export.go
MichaelMure: Both are basically the same thing, the context is no longer valid. They should have the same consequences.

commands/bridge_pull.go
MichaelMure: Any values in the channel is still available for read even if the channel has been closed. It's safe.

A-Hilaly: Nice ! Didn't know about this

bridge/gitlab/import.go
```diff
@@ -35,49 +34,83 @@ func (gi *gitlabImporter) Init(conf core.Configuration) error {
...
+						return
+					}
+				}
+
+				// Loop over all label events
+				for gi.iterator.NextLabelEvent() {
+					labelEvent := gi.iterator.LabelEventValue()
+					if err := gi.ensureLabelEvent(repo, b, labelEvent); err != nil {
+						err := fmt.Errorf("label event creation: %v", err)
+						out <- core.NewImportError(err, entity.Id(strconv.Itoa(labelEvent.ID)))
```
A-Hilaly: I have a lot of these transformation in PR (`string` to `entity.Id` and `int` to `entity.Id`) would it be better to add two new functions to the package `entity` ( `IdFromString` and `IdFromInt` ) ?

bridge/gitlab/import.go
```diff
@@ -269,6 +323,9 @@ func (gi *gitlabImporter) ensureNote(repo *cache.RepoCache, b *cache.BugCache, n
 		NOTE_UNLOCKED,
 		NOTE_MENTIONED_IN_ISSUE,
 		NOTE_MENTIONED_IN_MERGE_REQUEST:
+
+		reason := fmt.Sprintf("unsupported note type: %v", noteType)
```
A-Hilaly: Maybe better readable reasons instead of just passing the `noteType` which is an integer here

bridge/gitlab/iterator.go
```diff
@@ -52,12 +56,13 @@ type iterator struct {
 }
 
 // NewIterator create a new iterator
-func NewIterator(projectID, token string, since time.Time) *iterator {
+func NewIterator(ctx context.Context, capacity int, projectID, token string, since time.Time) *iterator {
```
A-Hilaly: Now that capacity is a parameter in `gitlab.NewIterator` and `github.NewIterator` should we change the `core.Importer` and `core.Exporter` interfaces so we can adjust this variable from the bridge commande line ?

bridge/github/import.go
```diff
@@ -42,40 +40,67 @@ func (gi *githubImporter) Init(conf core.Configuration) error {
...
+	out := make(chan core.ImportResult)
+	gi.out = out
+
+	go func() {
+		defer close(gi.out)
+
+		// Loop over all matching issues
+		for gi.iterator.NextIssue() {
+			issue := gi.iterator.IssueValue()
+			var b *cache.BugCache
```
A-Hilaly: if the context is cancelled and this variable isn't stil initialised we'll have a `panic` if we try to commit the bug in `select` block. maybe something like this can be better
```golang
if b != nil {
    if err := b.CommitAsNeeded(); err != nil {
	err := errors.Wrap(err, "bug commit")
	out <- core.NewImportError(err, b.Id())
    }
}
```

bridge/gitlab/iterator.go
MichaelMure: Probably not. The good value for capacity will likely be different for github and gitlab, you can't really generalize.

bridge/github/import.go
A-Hilaly: done

bridge/gitlab/import.go
A-Hilaly: done

bridge/github/import.go
```diff
@@ -42,40 +40,68 @@ func (gi *githubImporter) Init(conf core.Configuration) error {
...
+			var err error
+
+			select {
+			case <-ctx.Done():
+				if b != nil {
```
golangcibot: nilness: impossible condition: nil != nil (from `govet`)

bridge/gitlab/import.go
```diff
@@ -35,49 +34,84 @@ func (gi *gitlabImporter) Init(conf core.Configuration) error {
...
+
+		// Loop over all matching issues
+		for gi.iterator.NextIssue() {
+			issue := gi.iterator.IssueValue()
+			var b *cache.BugCache
+			var err error
+
+			select {
+			case <-ctx.Done():
+				if b != nil {
```
golangcibot: nilness: impossible condition: nil != nil (from `govet`)

bridge/launchpad/launchpad_api.go
```diff
@@ -182,6 +148,7 @@ func (lapi *launchpadAPI) queryBug(url string) (LPBug, error) {
 	if err != nil {
 		return bug, err
 	}
+	req = req.WithContext(ctx)
```
golangcibot: SA4006: this value of `req` is never used (from `staticcheck`)

bridge/gitlab/import.go
```diff
@@ -89,13 +98,14 @@ func (gi *gitlabImporter) ensureIssue(repo *cache.RepoCache, issue *gitlab.Issue
 
 	// resolve bug
 	b, err := repo.ResolveBugCreateMetadata(keyGitlabUrl, issue.WebURL)
-	if err != nil && err != bug.ErrBugNotExist {
-		return nil, err
-	}
-
 	if err == nil {
+		reason := fmt.Sprintf("bug already imported")
```
MichaelMure: image

bridge/github/export.go
```diff
@@ -117,20 +117,28 @@ func (ge *githubExporter) ExportAll(repo *cache.RepoCache, since time.Time) (<-c
 				return
 			}
 
-			snapshot := b.Snapshot()
+			select {
```
MichaelMure: I'm not sure you really need that, but okay

bridge/gitlab/import.go
A-Hilaly: Aiiiiee

bridge/github/export.go
A-Hilaly: `exportBug` handle context cancellation only it's making http requests. If all the operations are already exported the importer will not stop iterating and never catch the context cancel

commands/bridge_push.go
```diff
@@ -32,20 +35,55 @@ func runBridgePush(cmd *cobra.Command, args []string) error {
...
 	}
 
-	for result := range out {
-		if result.Err != nil {
-			fmt.Println(result.Err, result.Reason)
-		} else {
-			fmt.Printf("%s: %s\n", result.String(), result.ID)
+	exportedIssues := 0
+	for result := range events {
+		fmt.Println(result.String())
```
MichaelMure: could you not print the "nothing" events ? same for the import.

bridge/launchpad/import.go
```diff
@@ -45,98 +44,116 @@ func (li *launchpadImporter) ensurePerson(repo *cache.RepoCache, owner LPPerson)
...
-				map[string]string{
-					keyLaunchpadID: lpBugID,
-				},
-			)
-			if err != nil {
-				return errors.Wrapf(err, "failed to add bug id #%s", lpBugID)
+	go func() {
+		for _, lpBug := range lpBugs {
+			select {
+			case <-ctx.Done():
```
MichaelMure: same, do you really need that select ?

A-Hilaly: `lpAPI.SearchTasks` query multiple tasks a time, we need to check context cancellation at each iteration

commands/bridge_push.go
A-Hilaly: done

PR #192, opened 2019-08-05
bridge: Github exporter support case sensitive labels

### LINKED ISSUE #176: bug: name case sensitivity in retrieving and creating labels using github graphql api
When exporting new events to Github, creating and retrieving bugs using the graphql API can lead to some unexpected behaviour (only in some special cases):
- creating a label `bug` then trying to create a label `Bug` will return an error `Name has already been taken`
- having only one label named `bug` and trying to retrieve a label with name `Bug` will return a null ID.

For now to avoid this problem, try to not use two labels with the same letters like `core` and `CorE`.

I wrote a message to the Github support, i'll keep this issue up to date once i have a response.

A-Hilaly:
- import all existing labels at Github before start export task
- fix #176

### DIFF
--- a/bridge/github/export.go
+++ b/bridge/github/export.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	"strings"
 	"time"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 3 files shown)


A-Hilaly:
it is not necessary but i think it would be better to rebase and merge after #190

bridge/github/export.go
```diff
@@ -100,6 +100,17 @@ func (ge *githubExporter) ExportAll(repo *cache.RepoCache, since time.Time) (<-c
 		return nil, err
 	}
 
+	client, err := ge.getIdentityClient(user.Id())
+	if err != nil {
+		return nil, err
+	}
+
+	// query all labels
+	err = ge.cacheGithubLabels(client)
```
MichaelMure: you will need to have pass the context here as well after #190

A-Hilaly: Exactly why i suggested a rebase before :)

PR #193, opened 2019-08-07
Future proof the operation's ID

MichaelMure:
- Fix a potential problem where en operation's ID would change if go's json code changed
- rename Hash to ID to be consistent
- ID() doesn't return an error anymore

I'm still pondering if a custom type should be used instead of `string` (kind of like how `git.Hash` was used, but without any relation to git).

### DIFF
--- a/bridge/gitlab/import.go
+++ b/bridge/gitlab/import.go
@@ -132,7 +132,7 @@ func (gi *gitlabImporter) ensureNote(repo *cache.RepoCache, b *cache.BugCache, n
 	gitlabID := parseID(note.ID)
 
 	id, errResolve := b.ResolveOperationWithMetadata(keyGitlabId, gitlabID)
-	if errResolve != cache.ErrNoMatchingOp {
+	if errResolve != nil && errResolve != cache.ErrNoMatchingOp {
 		return errResolve
 	}
 
--- a/bridge/gitlab/import.go
+++ b/bridge/gitlab/import.go
@@ -11,7 +11,6 @@ import (
 	"github.com/MichaelMure/git-bug/bug"
 	"github.com/MichaelMure/git-bug/cache"
 	"github.com/MichaelMure/git-bug/identity"
-	"github.com/MichaelMure/git-bug/util/git"
 	"github.com/MichaelMure/git-bug/util/text"
 )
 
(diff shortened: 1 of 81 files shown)


MichaelMure:
Also note: this is not a breaking change :tada:

bridge/gitlab/import.go
```diff
@@ -185,10 +184,10 @@ func (gi *gitlabImporter) ensureNote(repo *cache.RepoCache, b *cache.BugCache, n
 			_, err = b.EditCommentRaw(
```
A-Hilaly: `errResolve` here can never equal `cache.ErrNoMatchingOp` you think we should drop the first error check ?

A-Hilaly:
> I'm still pondering if a custom type should be used instead of string (kind of like how git.Hash was used, but without any relation to git).

I think it is good to differentiate between normal ID's and git hashes, also`git.Hash` got `String()` method. This make sense for me

bridge/gitlab/import.go
```diff
@@ -130,10 +129,10 @@ func (gi *gitlabImporter) ensureIssue(repo *cache.RepoCache, issue *gitlab.Issue
 }
 
 func (gi *gitlabImporter) ensureNote(repo *cache.RepoCache, b *cache.BugCache, note *gitlab.Note) error {
-	id := parseID(note.ID)
+	gitlabID := parseID(note.ID)
 
-	hash, errResolve := b.ResolveOperationWithMetadata(keyGitlabId, id)
-	if errResolve != nil && errResolve != cache.ErrNoMatchingOp {
```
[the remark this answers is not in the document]
MichaelMure: huu, actually I'm not sure why I made this change anymore. Does that make sense to you ?

graphql/resolvers/operations.go
```diff
@@ -6,39 +6,74 @@ import (
...
+var _ graph.EditCommentOperationResolver = editCommentOperationResolver{}
+
 type editCommentOperationResolver struct{}
 
+func (editCommentOperationResolver) ID(ctx context.Context, obj *bug.EditCommentOperation) (string, error) {
+	return obj.Id().String(), nil
+}
+
+func (editCommentOperationResolver) Target(ctx context.Context, obj *bug.EditCommentOperation) (string, error) {
+	panic("implement me")
```
MichaelMure: @A-Hilaly  :-O

bridge/gitlab/import.go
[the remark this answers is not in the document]
A-Hilaly: Thinking a a little bit now i don't think it's correct to do this. We shouldn't trust `nil` errors and return them like in Github importer. like a i said above having an imported operation doesn't mean we have the latest error

graphql/resolvers/operations.go
A-Hilaly: Who did this x) ?? Is it complex to implement ?

PR #205, opened 2019-08-26
issue 178: fetch the repo dir with rev-parse --git-dir

seeduvax:
Since is returns the .git dir directly, it is not more needed to
concatenate .git.

### DIFF
--- a/cache/repo_cache.go
+++ b/cache/repo_cache.go
@@ -350,11 +350,11 @@ func (c *RepoCache) writeIdentityCache() error {
 }
 
 func bugCacheFilePath(repo repository.Repo) string {
-	return path.Join(repo.GetPath(), ".git", "git-bug", bugCacheFile)
+	return path.Join(repo.GetPath(), "git-bug", bugCacheFile)
 }
 
 func identityCacheFilePath(repo repository.Repo) string {
-	return path.Join(repo.GetPath(), ".git", "git-bug", identityCacheFile)
+	return path.Join(repo.GetPath(), "git-bug", identityCacheFile)
 }
 
 func (c *RepoCache) buildCache() error {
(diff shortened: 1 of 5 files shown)


MichaelMure:
There is an issue with `InitGitRepo` (https://github.com/MichaelMure/git-bug/blob/master/repository/git.go#L117). The path should be adjusted in this function (`/foo` --> `/foo/.git`).

MichaelMure:
`CleanupTestRepos` is tricky as well, because it has to know if the repo is bare or not to remove the correct directory. I guess the simplest way would be to check the path, if it end with `.git` --> remove the parent dir, if not, remove the direct directory.

seeduvax:
Just for my understanding: what's the purpose of InitGitRepo? Only testing?

seeduvax:
Even last commit makes everything OK for you, wait a while before merging. I observed a strange behavior with one of my hooks calling git bug. I need to check whether it is my changes or because I'm playing with a repository containing bugs created with a git-bug version from January.

seeduvax:
I confirm I've broken something, `git bug show -f humanId`no more works when done from my `prepare-commit-msg` hook.

MichaelMure:
> Just for my understanding: what's the purpose of InitGitRepo? Only testing?

Only testing ... for now.

MichaelMure:
> I confirm I've broken something, `git bug show -f humanId`no more works when done from my `prepare-commit-msg` hook.

You shouldn't need to do that, but maybe try to delete all files in `.git/git-bug/` ?

seeduvax:
With this last commit I do no more see failures while using git-bug with various kind of repo, and situations (including call from hooks).
I really don't understand why the `git rev-list`performed to retrieved the selected bug content was failing when git-bug is called from a hook. Doing the exact same call within the hook seems to work well.
@MichaleMure with this additional hack I hope it will OK for you. But I'll feel better with some explanation of this unexpected behavior of git.

seeduvax:
OK for `;`. I removed many myself and will surely missed some in future because of may too many year of coding with c/c++/java. I'll take care of it.

MichaelMure:
No worries, I did it already.

PR #210, opened 2019-08-30
interrupt: allow to cancel a cleaner

### LINKED ISSUE #209: feat: support cleaner cancellation or replacement in util/interrupt
Instead of storing cleaners in an array we can use maps the help identify and cancel/replace them in the other parts of git-bug code. See https://github.com/MichaelMure/git-bug/blob/master/util/interrupt/cleaner.go#L13

MichaelMure:
fix #209

### DIFF
--- a/util/interrupt/cleaner.go
+++ b/util/interrupt/cleaner.go
@@ -4,45 +4,75 @@ import (
 	"fmt"
 	"os"
 	"os/signal"
+	"sync"
 	"syscall"
 )
 
-// Cleaner type refers to a function with no inputs that returns an error
-type Cleaner func() error
+// CleanerFunc is a function to be executed when an interrupt trigger
+type CleanerFunc func() error
 
-var cleaners []Cleaner
-var active = false
+// CancelFunc, if called, will disable the associated cleaner.
+// This allow to create temporary cleaner. Be mindful though to not
+// create too much of them as they are just disabled, not removed from
+// memory.
+type CancelFunc func()
 
-// RegisterCleaner is responsible for registering a cleaner function. When a function is registered, the Signal watcher is started in a goroutine.
-func RegisterCleaner(f ...Cleaner) {
-	for _, fn := range f {
-		cleaners = append([]Cleaner{fn}, cleaners...)
-		if !active {
-			active = true
-			go func() {
-				ch := make(chan os.Signal, 1)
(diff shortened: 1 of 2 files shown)


util/interrupt/cleaner.go
```diff
@@ -4,45 +4,75 @@ import (
...
+// CleanerFunc is a function to be executed when an interrupt trigger
+type CleanerFunc func() error
+
+// CancelFunc, if called, will disable the associated cleaner.
+// This allow to create temporary cleaner. Be mindful though to not
+// create too much of them as they are just disabled, not removed from
+// memory.
+type CancelFunc func()
+
+type wrapper struct {
```
A-Hilaly: Should wrapper contain a sticky `sync.Mutex` ? In case concurrent read and write operations happens : `cancel()` and `clean()`

util/interrupt/cleaner.go
```diff
@@ -4,45 +4,75 @@ import (
...
+var handlerCreated = false
+
+// RegisterCleaner is responsible for registering a cleaner function.
+// When a function is registered, the Signal watcher is started in a goroutine.
+func RegisterCleaner(cleaner CleanerFunc) CancelFunc {
+	mu.Lock()
+	defer mu.Unlock()
+
+	w := &wrapper{f: cleaner}
+	cancel := func() { w.disabled = true }
```
A-Hilaly: Or maybe just use the global mutex 
```go
cancel := func() {
    mu.Lock()
    defer mu.Unlock()
    w.disabled = true
}
```

util/interrupt/cleaner.go
```diff
@@ -4,45 +4,75 @@ import (
...
+			continue
+		}
+		err := cleaner.f()
 		if err != nil {
-			errorlist = append(errorlist, err)
+			errorList = append(errorList, err)
 		}
 	}
-	cleaners = []Cleaner{}
+	cleaners = []*wrapper{}
```
A-Hilaly: i understand that cleaners list should be cleared after `clean()` call. Is there anyway we can reuse the list after cleaning up everything ?

MichaelMure: This is fine, no point in trying to reuse a list. Also, Exit(1) is called right after.

util/interrupt/cleaner.go
MichaelMure: Done

PR #211, opened 2019-08-31
Update webui

MichaelMure:
@sandhose it was OK until I tried to migrate to material-UI v4. Now it compiles and run but there is some runtime error and the UI is funky. Halp plz :)


MichaelMure:
Found the error, but the but list is still weird.
image

PR #219, opened 2019-09-17
Global config, global bridge token

MichaelMure:


repository/repo.go
```diff
@@ -30,24 +30,33 @@ type RepoCommon interface {
...
 	GetRemotes() (map[string]string, error)
 
-	// StoreConfig store a single key/value pair in the config of the repo
+	RepoConfig
+}
+
+type RepoConfig interface {
+	// StoreConfig store a single key/value pair in the config of the repos
 	StoreConfig(key string, value string) error
+	StoreGlobalConfig(key string, value string) error
```
MichaelMure: It's very much readable for a human reader, but the godoc police will complain if there is no doc for those exported functions.

A-Hilaly: Just commented every public function/interface in https://github.com/MichaelMure/git-bug/pull/219/commits/d0ea12f8601671233866754d100fe616c08b7e6a

cache/repo_cache.go
```diff
@@ -129,26 +129,46 @@ func (c *RepoCache) StoreConfig(key string, value string) error {
 	return c.repo.StoreConfig(key, value)
 }
 
+func (c *RepoCache) StoreGlobalConfig(key string, value string) error {
```
MichaelMure: doc over here as well please :)

commands/bridge_token.go
```diff
@@ -0,0 +1,64 @@
...
+	Use:     "token",
+	Short:   "Configure and use bridge tokens.",
+	PreRunE: loadRepo,
+	RunE:    runTokenBridge,
+	Args:    cobra.NoArgs,
+}
+
+func init() {
+	bridgeCmd.AddCommand(bridgeTokenCmd)
+	bridgeTokenCmd.Flags().BoolVarP(&bridgeTokenAll, "all", "a", false, "")
```
MichaelMure: not sure this "all" flag is really needed, isn't that simply the default ?

cache/repo_cache.go
A-Hilaly: Aiie forgot about this part x|

commands/bridge_token.go
A-Hilaly: Yep it should be the default normally .. it's true we don't need all these tree flags together it's a little confusing

A-Hilaly: Maybe just use a "filter" flag instead ? something like `--filter=kind:local`, but this also mean we should change `token.Global bool` to `token.Kind string`

MichaelMure: just remove the "all" flag. List everything by default, have the "local" and "global" flag be filters.

cache/repo_cache.go
```diff
@@ -129,26 +129,51 @@ func (c *RepoCache) StoreConfig(key string, value string) error {
...
+	return c.repo.StoreGlobalConfig(key, value)
+}
+
+// ReadConfigs read all key/value pair matching the key prefix in the repo config
 func (c *RepoCache) ReadConfigs(keyPrefix string) (map[string]string, error) {
 	return c.repo.ReadConfigs(keyPrefix)
 }
 
-// ReadConfigBool read a single boolean value from the config
+// ReadGlobalConfigs read a single boolean value from the global config
```
MichaelMure: that's incorrect

MichaelMure:
I made some cleanups while reading your code, some being just simplifications, some being actual change (ex: token validation before storing).

A few things that still need to be addressed:
- due to go preventing cyclic package import, the bridges implementation register is done in `bridge/bridges.go`. This means that from the commands, you should not use static functions like `core.StoreToken`. Those should be mapped in `bridge/bridges.go` to ensure that the implementations are properly registered.  
- with a command like `git config -l`, git will list **both** local and global values
- in term of UX, most of the time, the user should not have to care if the token is global or local. If a token is available from one or the other, it should work. Pretty much the only case where the user should care is when storing a new one, should it be just for this repo or global ?
- bridges configurator should ask that question ^

bridge/core/token.go
```diff
@@ -0,0 +1,246 @@
...
+	tokenScopesKey       = "scopes"
+)
+
+// Token holds an API access token data
+type Token struct {
+	id     entity.Id
+	Value  string
+	Target string
+	Global bool
+	Scopes []string
```
MichaelMure: is storing the scopes really useful ?

commands/bridge_token_add.go
```diff
@@ -0,0 +1,39 @@
...
+
+	"github.com/MichaelMure/git-bug/bridge/core"
+)
+
+var (
+	bridgeToken core.Token
+)
+
+func runBridgeTokenAdd(cmd *cobra.Command, args []string) error {
+	_ = bridgeToken.Id() // TODO: a better design to avoid doing this
```
MichaelMure: it's better to have individual variables for the parameters and to use the constructor

commands/bridge_token_rm.go
```diff
@@ -0,0 +1,29 @@
+package commands
+
+import (
+	"github.com/spf13/cobra"
+
+	"github.com/MichaelMure/git-bug/bridge/core"
+)
+
+func runBridgeTokenRm(cmd *cobra.Command, args []string) error {
+	err := core.RemoveToken(repo, args[0])
```
MichaelMure: Here again, a single function in the core would simplify this. As it is, you also have a bug: if attempting to remove an non-existing local token, `RemoveToken()` should return an error.

repository/repo.go
```diff
@@ -29,25 +33,47 @@ type RepoCommon interface {
 
 	// GetRemotes returns the configured remotes repositories.
 	GetRemotes() (map[string]string, error)
+}
 
-	// StoreConfig store a single key/value pair in the config of the repo
+// RepoConfig is a repository able to store configurations key/value pairs.
+type RepoConfig interface {
+	// StoreConfig store a single key/value pair in the config of the repos
 	StoreConfig(key string, value string) error
```
MichaelMure: there is now a misleading naming here: if `StoreConfig` and friends operate on the local config only, they should be named `StoreLocalConfig`.

bridge/core/token.go
A-Hilaly: No not really, i thought we could use this just as a reminder to prompt when we suggest a token for a new bridge creation

bridge/core/token.go
```diff
@@ -0,0 +1,246 @@
...
+	return token, nil
+}
+
+// GetToken loads a token from repo config
+func GetToken(repo repository.RepoConfig, id string) (*Token, error) {
+	return loadToken(repo, id, false)
+}
+
+// GetGlobalToken loads a token from the global config
+func GetGlobalToken(repo repository.RepoConfig, id string) (*Token, error) {
```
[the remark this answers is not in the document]
A-Hilaly: Yes that is totally possible. 
The more i think about this, i wonder why not just have Global tokens only ? supporting both local and global tokens add so much runtime complexity to the code.

repository/repo.go
A-Hilaly: Agreed since all global config functions have "Global" in their names

A-Hilaly: I was thinking a lot about these two modes of interaction with the configuration (local/global) and thought maybe we can make them both implement the same interface.
Maybe using something like
```go
type RepoConfig interface {
	StoreConfig(key string, value string) error
	ReadConfigs(keyPrefix string) (map[string]string, error)
	ReadConfigBool(key string) (bool, error)
	ReadConfigString(key string) (string, error)
	RmConfigs(keyPrefix string) error
}

type Repo interface {
	GlobalConfig() RepoConfig
	LocalConfig() RepoConfig
	...
}
```
This can allow using easily using different storage or libraries if needed, but on the other hand it will bring more structs into the code trying to implement the `GlobalConfig()` and `LocalConfig()`, but can be easily implemented by one struct since in our case the only different between storing configuration git locally and globally is the `--global` flag 
Any thought about this design ?

MichaelMure: hooo, I like that !

Just name it `Config` though

bridge/core/token.go
MichaelMure: do you mind if we remove that ? It's not super useful and as scopes can be changed on the bugtracker side without git-bug knowing I'd like to avoid having git-bug displaying something incorrect.

MichaelMure: What might be useful is a creation date though, to quickly see the old token if you want to rotate them regularly, or just as a helper to remember where a token comes from.

A-Hilaly: I like the idea of create date, i'll implement this and remove the scopes field

bridge/core/token.go
```diff
@@ -0,0 +1,182 @@
...
+	// trim key prefix
+	configs := make(map[string]string)
+	for key, value := range rawconfigs {
+		newKey := strings.TrimPrefix(key, keyPrefix)
+		configs[newKey] = value
+	}
+
+	token := &Token{}
+
+	token.Value, _ = configs[tokenValueKey]
```
golangcibot: S1005: should write `token.Value = configs[tokenValueKey]` instead of `token.Value, _ = configs[tokenValueKey]` (from `gosimple`)

bridge/core/token.go
```diff
@@ -0,0 +1,182 @@
...
+	token.Target, _ = configs[tokenTargetKey]
```
golangcibot: S1005: should write `token.Target = configs[tokenTargetKey]` instead of `token.Target, _ = configs[tokenTargetKey]` (from `gosimple`)

MichaelMure:
@cheshirekow please have a look as it's the new way for bridges to store credentials. It's also possible to extend it for new kind of credential if that's needed.

cheshirekow:
> @cheshirekow please have a look as it's the new way for bridges to store credentials. It's also possible to extend it for new kind of credential if that's needed.

Ok, I read through. Thanks for the heads up!

Issue #99, opened 2019-02-25
feat: search by inverse label


njthomas:
In both termui and on the command line, I can search for certain labels, but I can't search for bugs that don't have the label. eg:

```
    git bug  ls -l foo  # lists all bugs with label "foo"
```

It would be nice if there was an inverse label search. For example:

```
    git bug  ls -L foo  # lists all bugs that don't have label "foo"
```

this would be useful for many things.

MichaelMure:
I like this idea.

Now, due to the way things are done internally and by consistency, this feature also needs to exist in the [query language](https://github.com/MichaelMure/git-bug/blob/master/doc/queries.md). This query language is heavily inspired by the [Github one](https://help.github.com/en/articles/understanding-the-search-syntax). In [this document](https://help.github.com/en/articles/understanding-the-search-syntax#exclude-certain-results) they mention that you can prefix a term with a `-` to invert it's meaning.

So with you example, the plain-text query command would be `git bug ls -label:foo`. Problem is, this definitely looks like a command line flags, right ? That will confuse the CLI parser and potentially users as well. Maybe we could use `~` instead (used as "not" in logic notation)? Not sure that's less confusing though, as it looks very similar. Any better idea ?

Using the capital letter for the shorthand to invert the filter would works great though. For the long flags, maybe we can prefix them with `not-` ? So we would have `--not-label foo` or `--not-status open`.

Thoughts ?

njthomas:
Yup, I'm not married to the syntax, so "gut bug ls ~label:foo" is fine with me.

You may also want to consider how find does it, they use a bang (aka exclamation mark) there, eg:

     find /etc ! -type f

For the the long options --not-label is totally fine.

njthomas:
I didn't realize this until recently, but searches for multiple labels seem to be doing a **logical OR** between the various labels, ie:

    git bug ls -l foo1 -l foo2  # returns bugs that are labeled foo1 OR foo2

I was under the assumption that searching for multiple labels would filter the result set even further down.

Was this a deliberate choice?

The reason I ask is that I have some code written to implement an inverse label search. It works properly when given a single label, but the results are somewhat broken at the moment when given multiple labels. I wanted to make sure it was doing  the right thing when given multiple labels.

MichaelMure:
[...]

PR #220, opened 2019-09-18
search by inverse label

njthomas:
- this is for issue #99
- adds command line switch -L to "git bug ls" to search by inverse label
- adds query "inverselabel" in termui for the same

### DIFF
--- a/cache/filter.go
+++ b/cache/filter.go
@@ -42,16 +42,34 @@ func AuthorFilter(query string) Filter {
 	}
 }
 
-// LabelFilter return a Filter that match a label
-func LabelFilter(label string) Filter {
+// LabelFilterGeneric return a Filter that matches/not matches a label
+// depending on inverse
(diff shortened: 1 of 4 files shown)


MichaelMure:
Hey @njthomas, very sorry to not have addressed this earlier. I've been traveling for 5 weeks, not so much brain power to spare ...

So this issue has been a bit worked on by someone else without reaching an actual PR. The US consensus at that time was the following (say, for status):
- `--not-status=open` for the CLI flag
- no short flag like `-L`
- `~status:open` for the full string query

In addition, I wrote at that time two little helpers to easily invert filters: https://gist.github.com/MichaelMure/80c795d0c3b0c95ce8763e006dd719a9

Could you update your PR accordingly ?

MichaelMure:
This is still a valid feature request but the query code has moved so much since that this would have to be implemented again. Sorry :-|

PR #223, opened 2019-10-01
bridge/gitlab: fix integration tests

A-Hilaly:


bridge/gitlab/iterator.go
```diff
@@ -240,6 +255,8 @@ func (i *iterator) getNextLabelEvents() bool {
 	i.labelEvent.cache = labelEvents
 	i.labelEvent.page++
 	i.labelEvent.index = 0
+
+	sort.Sort(i.labelEvent)
```
MichaelMure: hmmm, this sort only within one page of result, right ? Can you have the same ordering problem across multiple pages ?

MichaelMure:
gitlab :-1:

bridge/gitlab/iterator.go
A-Hilaly: Yes exactly each time a page is queried sort will be called. 
Good catch i didn't think about that case. Maybe we should just query everything and run a sort ?

MichaelMure: That might be a solution, but I suspect there is a limit to how much values you can ask ?

Another way would be to add some king of streamed sorting logic in `NextLabelEvent()`: the function would sort the N next values, waiting for the next query to complete if there isn't enough available locally. This way, locally bad sorting would be fixed. As long as gitlab doesn't do something too stupid, that might work.

A-Hilaly: Yes "per_page" parameter possible maximum is 100. I checked the Gitlab API normally we won't have this ordering problem across multiple pages.

By streaming logic, do you mean: we query and sort all the pages one by one and then do try to do a merge sort ?

MichaelMure: What I mean is: if you take your series of result pages (each holding N values) and make them into a single stream of values, you simplify a lot your problem. You can then fix the sorting of this stream on the fly, without having to think about the concept of pages.

A-Hilaly: Implemented in https://github.com/MichaelMure/git-bug/pull/223/commits/312bc58c540902901ca9ad7238b4cf85eefb9d24

bridge/gitlab/iterator.go
```diff
@@ -25,6 +26,18 @@ type labelEventIterator struct {
 	cache []*gitlab.LabelEvent
 }
 
+func (l *labelEventIterator) Len() int {
+	return len(l.cache)
+}
+
+func (l *labelEventIterator) Swap(i, j int) {
+	l.cache[i], l.cache[j] = l.cache[j], l.cache[i]
```
A-Hilaly: I didn't know that this is possible in Go. Pretty interesting. 
https://stackoverflow.com/questions/35707222/swap-two-numbers-golang/35707958

A-Hilaly:
@MichaelMure Done ! i think this is ready to ship :shipit:

MichaelMure:
\o/

PR #228, opened 2019-10-15
Display label colors in termui

ludovicm67:
This pull request displays label colors in the termui in the following locations: 
- bug table
image

- bug view
image

- label edition widget
image


termui/bug_table.go
```diff
@@ -290,10 +290,19 @@ func (bt *bugTable) render(v *gocui.View, maxX int) {
...
+		summaryTxt := fmt.Sprintf("%4d 💬", excerpt.LenComments)
+		labelsTxt := "" // fmt.Sprintf("L:%-2d", len(excerpt.Labels))
+
+		nbLabels := 0
+		for _, l := range excerpt.Labels {
+			lc := l.Color()
+			lc256 := lc.Term256()
+			labelsTxt += lc256.Escape() + " ◼" + lc256.Unescape()
+			nbLabels++
+			if nbLabels >= 5 {
```
MichaelMure: It would be nice to use `…` to signal that there is more than what is displayed.

termui/bug_table.go
```diff
@@ -290,10 +290,19 @@ func (bt *bugTable) render(v *gocui.View, maxX int) {
 	columnWidths := bt.getColumnWidths(maxX)
 
 	for _, excerpt := range bt.excerpts {
-		summaryTxt := fmt.Sprintf("C:%-2d L:%-2d",
-			excerpt.LenComments,
-			len(excerpt.Labels),
-		)
+		summaryTxt := fmt.Sprintf("%4d 💬", excerpt.LenComments)
```
MichaelMure: hmm, this will overflow if over 9999, right ? What about using `∞` after that ?

termui/bug_table.go
```diff
@@ -290,10 +290,19 @@ func (bt *bugTable) render(v *gocui.View, maxX int) {
...
+		labelsTxt := "" // fmt.Sprintf("L:%-2d", len(excerpt.Labels))
+
+		nbLabels := 0
+		for _, l := range excerpt.Labels {
+			lc := l.Color()
+			lc256 := lc.Term256()
+			labelsTxt += lc256.Escape() + " ◼" + lc256.Unescape()
+			nbLabels++
```
MichaelMure: you can use the first value of the `for` for that:
```
for i, l := range excerpt.Labels {
    if i > 5 {
        ....
```

ludovicm67: Yes, I could, but I need the value of `nbLabels` later in the code.
What's the best way to keep the value of the last `i` out of the `for` loop?

termui/bug_table.go
ludovicm67: Yes, that's right. It will be fixed the way you proposed in my next commit.

termui/bug_table.go
ludovicm67: Yep, it's fixed in my next commit!

termui/bug_table.go
MichaelMure: ha yes i didn't notice that. The way you do it is good. Another way would be to have the next part of the code based on the resulting string instead of a counter but either way is fine.

MichaelMure:
Hey, sorry for the delay, could you also have the same changes in `git bug ls`?

MichaelMure:
I fixed a bad rendering when the termui window was too smal, which involved switching entirely to `go-term-text` and make some addition there. And also an adventure in unicode land.

Finally \o/

Thanks for the help :-)

MichaelMure:
And also drop support for go 1.9 apparently.

PR #236, opened 2019-10-25
commands: support bridge imports after a given date

### LINKED ISSUE #167: bridge: importing events occurred after a known date
For now the importer can only iterate over all available issues/comments/events and import them if they don't exist locally . 

We can add two flags to the bridge `pull` and `push` subcommands in order to allow only importing and exporting new events:
##### 1 - `--no-resume` 

for example, by default the importer will have to import only new events. passing this flag will force the bridge to import all available issues and events.

**example:** `git bug bridge pull --no-resume`

##### 2 - `--since`
to import/export events after a known date
**example:** `git bug bridge push --since=2019-06-08T21:49:06Z`

### LINKED ISSUE #203: bridge: import/export only after the last import/export date
This would be an important optimization. 

> import and export only issues created or updated after a certain date and time. This would be very useful to improve the import and export time for repositories containing a lot of issues.

A-Hilaly:
- Support imports after a given date
- resume import since last successful one

fixes #167 
fixes #203


commands/bridge_pull.go
```diff
@@ -64,14 +74,37 @@ func runBridgePull(cmd *cobra.Command, args []string) error {
...
 	})
 
-	// TODO: by default import only new events
-	events, err := b.ImportAll(ctx, time.Time{})
+	// initially the importSince time is set the zero value
+	importSince := time.Time{}
+
+	// if noResumeImport is false try to read the last import time from the bridge configuration
+	if !noResumeImport {
+		lastImportTimeStr, err := repo.ReadConfigString(fmt.Sprintf("git-bug.bridge.%s.lastImportTime", b.Name))
```
MichaelMure: what about a specialized `ReadConfigDate` ? It would simplify the code here and we can easily support more format later (timestamp, data in various formal, duration ...)

commands/bridge_pull.go
```diff
@@ -88,9 +121,13 @@ func runBridgePull(cmd *cobra.Command, args []string) error {
 	// send done signal
 	close(done)
 
-	fmt.Printf("Successfully imported %d issues and %d identities with %s bridge\n", importedIssues, importedIdentities, b.Name)
+	fmt.Printf("imported %d issues and %d identities with %s bridge\n", importedIssues, importedIdentities, b.Name)
 
-	return nil
+	// do not store last import time if context was cancelled
+	if ctx.Err() != context.Canceled && ctx.Err() != context.DeadlineExceeded {
```
MichaelMure: I'd say don't store the last import time if *any* error happened.

commands/bridge_pull.go
A-Hilaly: Good idea that could be very useful

commands/bridge_pull.go
A-Hilaly: Agreed. This will need mores checks for the event types

A-Hilaly:
I hesitated to do this since the `core.Bridge` struct doesn't provide any mechanism to look into the configuration or rewrite it (and the `conf` isn't exported). This maybe will need to redesign the `core.BridgeImpl` interface and allow configuration manipulation

A-Hilaly:
I had two designs in mind the first time i thought about this but i'm not sure if one of them is appropriate for our use cases

The first one, adding extra methods the `bridgeImpl` to have something like

```go
type BridgeImpl interface {
	// Target return the target of the bridge (e.g.: "github")
	Target() string

	// Configure handle the user interaction and return a key/value configuration
	// for future use
	Configure(repo repository.RepoCommon, params BridgeParams) (Configuration, error)

	// ValidateConfig check the configuration for error
	ValidateConfig(conf Configuration) error

	// NewImporter return an Importer implementation if the import is supported
	NewImporter() Importer

	// NewExporter return an Exporter implementation if the export is supported
	NewExporter() Exporter
	
	SaveConfiguration(conf Configuration) error
	
	GetConfiguration() Configuration
}
```

This one will change the way we create bridges, but is not consistent with the way we load bridges (unless we add another method `LoadConfiguration()` but this will give a bridge the priviledge to load it self )

The second one is about adding two methods to the `bridge` package

```go
func SetBridgeMetadata(bridge, key, valye string) error {}
func ReadBridgeMetadata(bridge, key string) (string, error) {}
```

This is one will do the job but the problem is that it isn't aware of the bridges configuration since they can have different ones (`gitlab` and `github` for example)
I personally don't think this a very good design but could work...

MichaelMure:
Both designs still leak what they shouldn't. Why a command should be aware of the configuration and it's structure ? Or metadata ? Only core and the bridge implementation should touch that. 

What about:
```go
type Importer interface {
	Init(conf Configuration) error
	ImportAll(ctx context.Context, repo *cache.RepoCache) (<-chan ImportResult, error)
	ImportSince(ctx context.Context, repo *cache.RepoCache, since time.Time) (<-chan ImportResult, error)
}
```

MichaelMure:
Note: this means that Importers record the last import time on their own. It that leads to too much code duplication, maybe some helpers in the core can help.

A-Hilaly:
Looks like a pretty good design for me. The only thing i'm worried about is that importers are not aware of which bridge they are implementing (not aware of the bridge name for example). Should i add a pointer to bridge in the importers implementing `core.Importer` (were the `Name`field is Public ) or just copying the name field ?

MichaelMure:
Looking at the code, the good place to implement the read/write of the last import time would be in `core.Bridge.Import*(...)`. There is everything you need there and it would be written once for all the bridges.

commands/bridge_pull.go
```diff
@@ -64,8 +75,20 @@ func runBridgePull(cmd *cobra.Command, args []string) error {
...
+	var events <-chan core.ImportResult
+	switch {
+	case noResume:
+		events, err = b.ImportAllSince(ctx, time.Time{})
+	case importSince != "":
+		since, err := parseSince(importSince)
+		if err != nil {
+			return errors.Wrap(err, "import time parsing")
+		}
+		events, err = b.ImportAllSince(ctx, since)
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

PR #240, opened 2019-10-31
WIP: repository config interface and implementation rework

A-Hilaly:
As discussed in https://github.com/MichaelMure/git-bug/pull/219#discussion_r340809076, this is a first draft trying to redesign the config interface and it implementations


repository/config.go
```diff
@@ -0,0 +1,23 @@
...
+	// Store writes a single key/value pair in the config of the repo
+	Store(key string, value string) error
+
+	// ReadAll reads all key/value pair matching the key prefix
+	ReadAll(keyPrefix string) (map[string]string, error)
+
+	// ReadBool read a single boolean value from the config
+	// Return ErrNoConfigEntry or ErrMultipleConfigEntry if
+	// there is zero or more than one entry for this key
+	ReadBool(key string) (bool, error)
```
A-Hilaly: Should we really keep different Read functions in the interface ? we'll have `ReadBool`, `ReadTimestamp` and `ReadString` probably more in the future.
Or maybe have one `Read(key string) (string, error)` and move the conversion utility to another package ?

repository/config_git.go
```diff
@@ -0,0 +1,185 @@
...
+	"github.com/pkg/errors"
+)
+
+type gitConfig struct {
+	version *semver.Version
+	execFn  func(args ...string) (string, error)
+}
+
+func NewGitConfig(repo *GitRepo, global bool) *gitConfig {
+	version, _ := repo.GitVersion()
```
A-Hilaly: Ignored this error in this draft, but we should not. Normally we should either we panic or return an error, but that will change the `LocalConfig()` `GlobalConfig()` signatures to `LocalConfig() (Config, error)`

repository/config_runtime.go
```diff
@@ -0,0 +1,59 @@
+package repository
+
+import (
+	"strconv"
+	"strings"
+)
+
+type runtimeConfig struct {
```
A-Hilaly: something like `type runtimeConfig map[string]string` could also be correct and used directly in the `mockRepoForTest`

MichaelMure: Not so happy with the naming, what about `mockConfig` or `memConfig` ?

repository/config.go
MichaelMure: I think it's fine to have that in the interface. It allows the implementations to store the value the way they want if needed. You can still have shared internal helpers to do the conversion if you want to reduce the code duplication.

repository/config_git.go
MichaelMure: why not have the version checked at the same place as before ?

repository/git_testing.go
```diff
@@ -31,10 +31,11 @@ func CreateTestRepo(bare bool) *GitRepo {
 		log.Fatal(err)
 	}
 
-	if err := repo.StoreConfig("user.name", "testuser"); err != nil {
+	config := repo.LocalConfig()
```
MichaelMure: That's another reason not to check the version in the constructor. You should be able to feel ok to write `repo.LocalConfig().Store(...)` without performance impact or weird thing happening.

repository/config_git.go
A-Hilaly: Because `execFn` passes `config` arg by default and we can not use it to get the version :|, one other possible thing would be to get the version at the moment we create the `GitRepo` object and we passe it by parameter in `newGitConfig` function

A-Hilaly: If i don't passe the `config` as arg i should have a boolean field storing the type of the struct (global/local) and well finish with a if/else situation in all the implementations

A-Hilaly: Another way would to store the `repo.GitVersion` as a field in `GitConfig` struct

repository/config.go
MichaelMure: yes, most likely

repository/config_git.go
MichaelMure: what if you just store the flag as a string (`--local` or `--global`) ? You will then just have to insert it where needed.

repository/git_test.go
```diff
@@ -11,56 +11,57 @@ func TestConfig(t *testing.T) {
...
 	defer CleanupTestRepos(t, repo)
 
-	err := repo.StoreConfig("section.key", "value")
+	config := repo.LocalConfig()
+
+	err := config.StoreString("section.key", "value")
 	assert.NoError(t, err)
 
-	val, err := repo.ReadConfigString("section.key")
+	val, err := config.ReadString("section.key")
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

repository/git_test.go
```diff
@@ -11,56 +11,57 @@ func TestConfig(t *testing.T) {
...
 	assert.Equal(t, "value", val)
 
-	err = repo.StoreConfig("section.true", "true")
+	err = config.StoreString("section.true", "true")
 	assert.NoError(t, err)
 
-	val2, err := repo.ReadConfigBool("section.true")
+	val2, err := config.ReadBool("section.true")
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

repository/git_test.go
```diff
@@ -11,56 +11,55 @@ func TestConfig(t *testing.T) {
 	repo := CreateTestRepo(false)
 	defer CleanupTestRepos(t, repo)
 
-	err := repo.StoreConfig("section.key", "value")
+	err := repo.LocalConfig().StoreString("section.key", "value")
 	assert.NoError(t, err)
 
-	val, err := repo.ReadConfigString("section.key")
+	val, err := repo.LocalConfig().ReadString("section.key")
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

repository/git_test.go
```diff
@@ -11,56 +11,55 @@ func TestConfig(t *testing.T) {
...
 	assert.Equal(t, "value", val)
 
-	err = repo.StoreConfig("section.true", "true")
+	err = repo.LocalConfig().StoreString("section.true", "true")
 	assert.NoError(t, err)
 
-	val2, err := repo.ReadConfigBool("section.true")
+	val2, err := repo.LocalConfig().ReadBool("section.true")
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

MichaelMure:
:tada:

PR #247, opened 2019-11-10
bridge: move keyOrigin to core package

A-Hilaly:


bridge/core/bridge.go
```diff
@@ -20,8 +20,8 @@ var ErrImportNotSupported = errors.New("import is not supported")
 var ErrExportNotSupported = errors.New("export is not supported")
 
 const (
-	KeyTarget = "target"
-	KeyOrigin = "origin"
+	KeyTarget         = "target"
```
MichaelMure: when you do this kind of specialization in a variable name, it's good to also rename the other side: `ConfigKeyTarget`

A-Hilaly: Done !

MichaelMure:
cc @cheshirekow

Issue #231, opened 2019-10-22
import/export: the "nothing" event should be emitted only when nothing happen on a bug, not on an operation


MichaelMure:

MichaelMure:
So to be more explicite, the bridge should emit an event for each kind of imported thing (bug, comment, label ...) **or** the nothing event (meaning nothing happened on the bug) **or** an error event.

Issue #212, opened 2019-09-04
Reflections on the change needed in the data model


MichaelMure:
It appears to me that some changes are needed to fix some current problems or support more usecase.

To help structure the reflection, let's start with a ...

## Threat model

Level 1:
- all actors behave in good faith and run the normal software
- the goal is to prevent accidental tempering or corruption of the data, and in particular, changes or suppression in the semantic of a bug.

In essence, this threat model represent the normal workflow of git-bug where all the actors are identified and validated, either by being authorised to push to a repository, being a source repository being pulled, or by being a bot of some sort with the same level of credential.

Being resistant to this threat model is the minimal acceptable protection that git-bug should provide.

Level 2:
- bad actor(s) try to actively temper or corrupt the data, and in particular, try to change, inject or suppress some semantic information of a bug.
- they can run a custom version of the software

This threat model represent a fully decentralized workflow for git-bug where the actors are not fully identified and validated.

Being resistant to this threat model is **not** a goal of git-bug at the moment. This threat model exist to be able to identify where git-bug might fail in this scenario and possibly to make educated design decision towards supporting it at some point. In any case, being resistant to this thread model automatically imply being resistant to the first one and might lead to more elegant solution.

## Point of attention

- ordering of the operation inside a bug
    - reordering of existing operation inside an OperationPack
    - reordering of existing OperationPack
- insertion of an extra operation

## Problem#1: operation's ID collision

As of now, it's fairly easy to create collision of Operation's ID, especially for simple Operation like SetStatusOperation. Create two "close" operation within the same second and you have a collision. Essentially, there is either not enough entropy or no mechanism to ensure uniqueness (or really *enough* uniqueness as we are dealing with a distributed system).

Possible solutions:
- add some sort of arbitrary counter
    - not great as it doesn't give uniqueness guarantee (two actors can select the same next index as they don't have the complete knowledge of the system)
- increase the resolution of the time field
    - make accidental collision way less likely
- add a "random data" field like in `identity.Version`
- (?) add a Vector clock per bug with a clock value for each operation, making them unique, or another solution for Problem#2

## Problem#2: support decentralized workflow & operation ordering

As of now, git-bug enforce to have a single linear chain of operations. Those operations are interpreted in the order defined by this chain, which means that the oldest operation take precedence. If an operation's semantic doesn't make sense after the action of another, it is ignored.

[...]

PR #250, opened 2019-11-14
Implement jira bridge

cheshirekow:


cheshirekow:
Some notes:

I suggest you start with doc/jira_bridge.md as it describes several of the jira-specific issues that are dealt with by the bridge, as well as some (initial) design decisions. 

Note that I've tested the import quite a bit, but I've only tested the export with one bug that I created, exported, imported, modified, exported again. I don't want to use my work JIRA instance in case I messed up and export thousands duplicate bugs into our jira (the project I used to test has 6000 issues). One thing that would help a lot is if we could get in touch with Atlassian and request that they donate a license (or a cloud instance) to the `git-bug` project for testing . I think we really need a mock server for CI,but for dev cycles it would definitely be nice to work with the real thing. AlsoI think we should really test the export a bit before putting this out in the world. I would be happy to try and contact Atlassian, but I think as the project owner they  might appreciate more hearing from @MichaelMure. 

The bridge currently does nothing to import any of the JIRA fields that don'thave `git-bug` equivalents ("Assignee", "sprint", "story points", etc). I'm not sure what the right way to go about this is, but it would really be awesomeif `git-bug` could support all of these fields. I would love to hear thoughts on this for future development. 

I don't fully understand what the whole `getIdentityClient` code is for in the existing plugins. I have a suspicion it's there for future use (the ability to assume different identities and act on behalf of those identities). I tried to keep that stuff around in the JIRA bridge but that code isn't really doinganything right now.

bridge/jira/export.go
```diff
@@ -0,0 +1,453 @@
...
+					author.Id().String()))
+			return
+		}
+
+		// Load any custom fields required to create an issue from the git
+		// config file.
+		fields := make(map[string]interface{})
+		defaultFields, hasConf := self.conf[keyCreateDefaults]
+		if hasConf {
+			json.Unmarshal([]byte(defaultFields), &fields)
```
golangcibot: Error return value of `json.Unmarshal` is not checked (from `errcheck`)

bridge/jira/import.go
```diff
@@ -0,0 +1,598 @@
...
+				err := fmt.Errorf("issue creation: %v", err)
+				out <- core.NewImportError(err, "")
+				return
+			}
+
+			var commentIter *CommentIterator
+			for commentIter =
+				client.IterComments(issue.ID, defaultPageSize); commentIter.HasNext(); {
+				comment := commentIter.Next()
+				self.ensureComment(repo, bug, *comment)
```
golangcibot: Error return value of `self.ensureComment` is not checked (from `errcheck`)

bridge/jira/import.go
```diff
@@ -0,0 +1,598 @@
...
+					exportTime, err = http.ParseTime(exportTimeStr)
+					if err != nil {
+						continue
+					}
+					if !exportTime.Before(changelogEntry.Created.Time) {
+						break
+					}
+				}
+				if opIdx < len(snapshot.Operations) {
+					self.ensureChange(
```
golangcibot: Error return value of `self.ensureChange` is not checked (from `errcheck`)

bridge/jira/import.go
```diff
@@ -0,0 +1,598 @@
...
+						repo, bug, *changelogEntry, snapshot.Operations[opIdx])
+				} else {
+					self.ensureChange(repo, bug, *changelogEntry, nil)
```
golangcibot: Error return value of `self.ensureChange` is not checked (from `errcheck`)

bridge/jira/import.go
```diff
@@ -0,0 +1,598 @@
...
+			fromLabels := strings.Split(item.FromString, " ")
+			toLabels := strings.Split(item.ToString, " ")
+			removedLabels, addedLabels, _ := setSymmetricDifference(
+				fromLabels, toLabels)
+
+			opr, isRightType := potentialOp.(*bug.LabelChangeOperation)
+			if isRightType &&
+				labelSetsMatch(addedLabels, opr.Added) &&
+				labelSetsMatch(removedLabels, opr.Removed) {
+				b.SetMetadata(opr.Id(), map[string]string{
```
golangcibot: Error return value of `b.SetMetadata` is not checked (from `errcheck`)

bridge/jira/import.go
```diff
@@ -0,0 +1,598 @@
...
+					keyJiraOperationID: entry.ID,
+				})
+				self.out <- core.NewImportNothing("", "matched export")
+				return nil
+			}
+
+		case "status":
+			opr, isRightType := potentialOp.(*bug.SetStatusOperation)
+			if isRightType && statusMap[opr.Status.String()] == item.ToString {
+				b.SetMetadata(opr.Id(), map[string]string{
```
golangcibot: Error return value of `b.SetMetadata` is not checked (from `errcheck`)

bridge/jira/import.go
```diff
@@ -0,0 +1,598 @@
...
+				self.out <- core.NewImportNothing("", "matched export")
+				return nil
+			}
+
+		case "summary":
+			// NOTE(josh): JIRA calls it "summary", which sounds more like the body
+			// text, but it's the title
+			opr, isRightType := potentialOp.(*bug.SetTitleOperation)
+			if isRightType && opr.Title == item.ToString {
+				b.SetMetadata(opr.Id(), map[string]string{
```
golangcibot: Error return value of `b.SetMetadata` is not checked (from `errcheck`)

bridge/jira/export.go
```diff
@@ -0,0 +1,453 @@
...
+
+	"github.com/pkg/errors"
+
+	"github.com/MichaelMure/git-bug/bridge/core"
+	"github.com/MichaelMure/git-bug/bug"
+	"github.com/MichaelMure/git-bug/cache"
+	"github.com/MichaelMure/git-bug/entity"
+)
+
+var errDuplicateMatch = errors.New("Ambiguous match")
```
golangcibot: `errDuplicateMatch` is unused (from `varcheck`)

bridge/jira/client.go
```diff
@@ -0,0 +1,1453 @@
...
+	"time"
+
+	"github.com/MichaelMure/git-bug/bridge/core"
+	"github.com/MichaelMure/git-bug/bug"
+	"github.com/pkg/errors"
+)
+
+var errDone = errors.New("Iteration Done")
+var errTransitionNotFound = errors.New("Transition not found")
+var errTransitionNotAllowed = errors.New("Transition not allowed")
```
golangcibot: `errTransitionNotAllowed` is unused (from `varcheck`)

bridge/jira/client.go
```diff
@@ -0,0 +1,1453 @@
...
+
+	// TODO(josh)[767ee72]: Figure out a good way to "configure" the
+	// open/close state mapping. It would be *great* if we could actually
+	// *compute* the necessary transitions and prompt for missing metatdata...
+	// but that is complex
+	var buffer bytes.Buffer
+	fmt.Fprintf(&buffer,
+		`{"transition":{"id":"%s"}, "resolution": {"name": "Done"}}`,
+		transitionID)
+	request, err := http.NewRequest("POST", url, bytes.NewBuffer(buffer.Bytes()))
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

bridge/jira/config.go
```diff
@@ -0,0 +1,239 @@
...
+
+	jsonData, err := json.Marshal(
+		&SessionQuery{Username: username, Password: password})
+	if err != nil {
+		return nil, err
+	}
+
+	fmt.Printf("Attempting to login with credentials...\n")
+	client := NewClient(serverURL, nil)
+	err = client.RefreshTokenRaw(jsonData)
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

bridge/jira/import.go
```diff
@@ -0,0 +1,598 @@
...
+	if item.Updated == item.Created {
+		return nil
+	}
+
+	// If there has been an update to this comment, we try to find it in the
+	// database. We need a unique id so we'll concat the issue id with the update
+	// timestamp. Note that this must be consistent with the exporter during
+	// export of an EditCommentOperation
+	derivedID := getTimeDerivedID(item.ID, item.Updated)
+	targetOpID, err = b.ResolveOperationWithMetadata(
```
golangcibot: ineffectual assignment to `targetOpID` (from `ineffassign`)

bridge/jira/import.go
```diff
@@ -0,0 +1,598 @@
...
+	return fmt.Sprintf("%s-%d", jiraID, idx)
+}
+
+func labelSetsMatch(jiraSet []string, gitbugSet []bug.Label) bool {
+	if len(jiraSet) != len(gitbugSet) {
+		return false
+	}
+
+	sort.Strings(jiraSet)
+	gitbugStrSet := make([]string, len(gitbugSet), len(gitbugSet))
```
golangcibot: S1019: should use make([]string, len(gitbugSet)) instead (from `gosimple`)

bridge/jira/export.go
```diff
@@ -0,0 +1,453 @@
...
+			out <- core.NewExportNothing(
+				op.Id(), fmt.Sprintf(
+					"missing operation author credentials for user %.8s",
+					author.Id().String()))
+			continue
+		}
+
+		var id string
+		var exportTime time.Time
+		switch op.(type) {
```
golangcibot: S1034: assigning the result of this type assertion to a variable (switch op := op.(type)) could eliminate the following type assertions:
	bridge/jira/export.go:283:11
	bridge/jira/export.go:297:11
	bridge/jira/export.go:335:11
	bridge/jira/export.go:358:11
	bridge/jira/export.go:371:11 (from `gosimple`)

MichaelMure:
> I think we really need a mock server for CI,but for dev cycles it would definitely be nice to work with the real thing. AlsoI think we should really test the export a bit before putting this out in the world. I would be happy to try and contact Atlassian, but I think as the project owner they might appreciate more hearing from @MichaelMure.

Why not using a free plan, hosted by Atlassian ? It's up to 10 seat, "free forever".

> The bridge currently does nothing to import any of the JIRA fields that don'thave `git-bug` equivalents ("Assignee", "sprint", "story points", etc). I'm not sure what the right way to go about this is, but it would really be awesomeif `git-bug` could support all of these fields. I would love to hear thoughts on this for future development.

I thouched a bit on that here: https://github.com/MichaelMure/git-bug/pull/241#discussion_r343920996

> I don't fully understand what the whole `getIdentityClient` code is for in the existing plugins. I have a suspicion it's there for future use (the ability to assume different identities and act on behalf of those identities). I tried to keep that stuff around in the JIRA bridge but that code isn't really doing anything right now.

Yeah, it doesn't really do anything at the moment but will eventually. The goal is to be able to load as much credential as necessary and being able to use git-bug as a "real" bridge between two bug tracker for a migration. It's not really part of git-bug's end goal but it was a fairly simply architectural change to enable that use case.

MichaelMure:
Huu, sorry for asking that *after* you wrote that much code, but have you looked at https://github.com/andygrunwald/go-jira ? Wouldn't that simplify a bunch of things in here ?

cheshirekow:
> Why not using a free plan, hosted by Atlassian ? It's up to 10 seat, "free forever".

For real? Wow awesome. I only found the 30 day server trial. I'll look into this. 

> I thouched a bit on that here

Ok great, I haven't gotten back to that discussion yet. I've been focused on this bridge :). 

> Yeah, it doesn't really do anything at the moment but will eventually.

Ok good. I think I've respected the setup and the plumbing is in place for this bridge as well. 

> but have you looked at https://github.com/andygrunwald/go-jira 

Nope. It's news to me. 

> Wouldn't that simplify a bunch of things in here ?

I don't think so... I mean it would be a lot *less code* to use someone else's REST wrapper, if that's what you mean, but I don't think the rest of the bridge implementation changes. If that library looks good I would be happy to swap out for it. 

> sorry for asking that after you wrote that much code

..meh, the big time sink was studying the API and testing behaviors. Writing the REST wrapper was relatively quick.

bridge/jira/client.go
cheshirekow: Done

bridge/jira/client.go
cheshirekow: Done

bridge/jira/config.go
cheshirekow: DOne

bridge/jira/export.go
cheshirekow: Done

bridge/jira/export.go
cheshirekow: Done

bridge/jira/export.go
cheshirekow: Done

bridge/jira/import.go
cheshirekow: Done

bridge/jira/import.go
cheshirekow: Done

bridge/jira/import.go
cheshirekow: Done

bridge/jira/import.go
cheshirekow: Done

bridge/jira/import.go
cheshirekow: Done

bridge/jira/import.go
cheshirekow: Done

bridge/jira/import.go
cheshirekow: Done

bridge/jira/import.go
cheshirekow: Done

MichaelMure:
>  I mean it would be a lot less code to use someone else's REST wrapper

Having someone else do the maintenance is a big win, at least for me ;)

bridge/jira/import.go
```diff
@@ -0,0 +1,613 @@
...
+	// ensure person
+	author, err := self.ensurePerson(repo, item.Author)
+	if err != nil {
+		return err
+	}
+
+	targetOpID, err := b.ResolveOperationWithMetadata(
+		keyJiraID, item.ID)
+	if err == nil {
+		self.out <- core.NewImportNothing("", "comment already imported")
```
MichaelMure: Those Nothing event are not really useful the way they are done at the moment ... (for all the bridges, not just this PR).

I think that ideally, the bridge would emit an event for each kind of imported thing (bug, comment, label ...) **or** the nothing event (meaning nothing happened on the bug) **or** an error event.

MichaelMure: I updated https://github.com/MichaelMure/git-bug/issues/231 about this.

bridge/jira/export.go
```diff
@@ -0,0 +1,449 @@
...
+	ctx *context.Context, id entity.Id) (*Client, error) {
+	client, ok := self.identityClient[id]
+	if ok {
+		return client, nil
+	}
+
+	// TODO(josh)[]: The github exporter appears to contain code that will
+	// allow it to export bugs owned by other people as long as we have a token
+	// for that identity. I guess the equivalent for us will be as long as we
+	// have a credentials pair for that identity.
```
MichaelMure: correct

bridge/jira/export.go
```diff
@@ -0,0 +1,449 @@
...
+	user, err := repo.GetUserIdentity()
+	if err != nil {
+		return nil, err
+	}
+
+	// TODO(josh)[]: The github exporter appears to contain code that will
+	// allow it to export bugs owned by other people as long as we have a token
+	// for that identity. I guess the equivalent for us will be as long as we
+	// have a credentials pair for that identity.
+	client := NewClient(self.conf[keyServer], &ctx)
```
MichaelMure: just use getIdentityClient and build the client there ?

bridge/jira/export.go
```diff
@@ -0,0 +1,449 @@
...
+		}
+
+		// Load any custom fields required to create an issue from the git
+		// config file.
+		fields := make(map[string]interface{})
+		defaultFields, hasConf := self.conf[keyCreateDefaults]
+		if hasConf {
+			err = json.Unmarshal([]byte(defaultFields), &fields)
+			if err != nil {
+				panic("Invalid JSON in config")
```
MichaelMure: in go, the code should pretty much never panic. Just return an error.

bridge/jira/export.go
```diff
@@ -0,0 +1,449 @@
...
+			out <- core.NewExportComment(op.Id())
+
+			// cache comment id
+			self.cachedOperationIDs[op.Id()] = id
+
+		case *bug.EditCommentOperation:
+			if opr.Target == createOp.Id() {
+				// An EditCommentOpreation with the Target set to the create operation
+				// encodes a modification to the long-description/summary.
+				exportTime, err = client.UpdateIssueBody(bugJiraID, opr.Message)
```
MichaelMure: there is probably a way to pass the context up to the actual request to have a proper cancellation

bridge/jira/export.go
```diff
@@ -0,0 +1,449 @@
...
+					// possible that we just don't have enough information to make that
+					// update. In this case, just don't export the operation.
+					continue
+				}
+				out <- core.NewExportStatusChange(op.Id())
+				// TODO(josh)[c2c6767]: query changelog to get the changelog-id so that
+				// we don't re-import the same change.
+				id = bugJiraID
+			} else {
+				out <- core.NewExportNothing(
```
MichaelMure: I would think that if the bridge is unable to export a status, that would be an error

bridge/jira/export.go
```diff
@@ -0,0 +1,449 @@
...
+
+		default:
+			panic("unhandled operation type case")
+		}
+
+		// mark operation as exported
+		// TODO(josh)[c2c6767]: Should we query the changelog after we export?
+		// Some of the operations above don't record an ID... so we are bound to
+		// re-import them. It shouldn't cause too much of an issue but we will have
+		// duplicate edit entries for everything and it would be nice to avoid that.
```
MichaelMure: especially because it should be possible to run the bridge in a cron job or something to have the update. If the data increase each time for nothing ...

MichaelMure:
> For real? Wow awesome. I only found the 30 day server trial. I'll look into this.

I created an instance at https://git-bug.atlassian.net give me an email so I can invite you.

cheshirekow:
> > For real? Wow awesome. I only found the 30 day server trial. I'll look into this.
> 
> I created an instance at https://git-bug.atlassian.net give me an email so I can invite you.

Sent you an email. I'm glad you were able to get one. I've been trying for a few days now and I keep getting an error. Looks like it worked for me today though.

doc/jira_bridge.md
```diff
@@ -0,0 +1,365 @@
...
+
+JIRA does not support user/personal access tokens. They have experimental
+3-legged oauth support but that requires an API token for the app configured
+by the server administrator. The only reliable authentication mechanism then is
+the username/password and session-token mechanims. We can aquire a session
+token programatically from the username/password but these are very short lived
+(i.e. hours or less). As such the bridge currently requires an actual username
+and password as user credentials. It supports three options:
+
+1. Storing both username and password in a separate file referred to by
```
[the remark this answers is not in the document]
MichaelMure: If so it'd be nice to extend the core (with a credential type if that make sense, prompt functions ...) so other bridges can use it as well.

doc/jira_bridge.md
```diff
@@ -0,0 +1,365 @@
...
+we can match a changed field to the candidate `git-bug` operation then we have
+identified the match.
+
+### Unlogged Changes
+
+Comments (creation and edition) do not show up in the JIRA changelog. However
+JIRA reports both a `created` and `updated` date for each comment. If we
+import a comment which has an `updated` and `created` field which do not match,
+then we treat that as a new comment edition. If we do not already have the
+comment imported, then we import an empty comment followed by a comment edition.
```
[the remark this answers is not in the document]
MichaelMure: Operation order in the data-structure is meaningful and unless you manually alter it it's not possible to add an operation *before* an existing one. You wont be able to add a revision before another one.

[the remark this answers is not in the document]
MichaelMure: I think you are trying to solve the wrong problem here. The bridge should not import a comment that it already has. Maybe you can compute some sort of signature for your comment, maybe `hash(comment || date)` or something ?

doc/jira_bridge.md
```diff
@@ -0,0 +1,365 @@
...
+set during issue creation when exproting bugs. If you provide a value for this
+configuration option, it must include at least the `"issuetype"` field, or
+the bridge will not be able to export any new issues.
+
+Let's say that we want bugs exported to JIRA to have a default issue type of
+"Story" which is `issuetype` with id `10001`. Then we will add the following
+entry to our git-config:
+
+```
+create-issue-defaults = {"issuetype":"10001"}
```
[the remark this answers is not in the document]
MichaelMure: Alright, make sense to me.

bridge/jira/import.go
```diff
@@ -0,0 +1,613 @@
...
+	return nil
+}
+
+// ImportAll iterate over all the configured repository issues and ensure the
+// creation of the missing issues / timeline items / edits / label events ...
+func (self *jiraImporter) ImportAll(
+	ctx context.Context, repo *cache.RepoCache, since time.Time) (
+	<-chan core.ImportResult, error) {
+
+	sinceStr := since.Format("2006-01-02 15:04")
```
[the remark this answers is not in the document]
MichaelMure: hooo, forgot that one. So weird.

bridge/jira/export.go
MichaelMure: You are correct that nothing terrible can happen with this one, but it's also a question of UX. A stacktrace is not exactly user friendly ;)

bridge/jira/export.go
MichaelMure: Good point.

bridge/jira/export.go
MichaelMure: Although I didn't envision to have bridge return multiple error events, it's entirely possible. It'd be nice if we can stay within a single event per bug but if not that's ok.

doc/jira_bridge.md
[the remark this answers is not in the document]
cheshirekow: > Operation order in the data-structure is meaningful and unless you manually alter it it's not possible to add an operation before an existing one. You wont be able to add a revision before another one.

Wait, really? Maybe I misunderstood [this part][1]. I thought the whole idea of maintaining a list of edit operations was so that one could insert an interleaved operation. If not, that is fine. I guess it wont make much difference then whether we create the initial comment as empty or with the current text. I think it's important to actually record the edit operation, however, so that we know we don't have to do it on the next import of this issue. 

> The bridge should not import a comment that it already has.

It doesn't import a comment it already has. Sorry I wasn't clear: If the bridge already has a comment, but it observes a new revision, it does not import a whole new comment, it edits the existing comment.

[1]: https://github.com/MichaelMure/git-bug/blob/master/doc/model.md#you-cant-rely-on-the-time-provided-by-other-people-their-clock-might-by-off-for-anything-other-than-just-display

doc/jira_bridge.md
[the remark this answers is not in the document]
cheshirekow: That makes sense. We can look at the other bridges and see what all is common in them that can be shared.

bridge/jira/export.go
cheshirekow: Done.

doc/jira_bridge.md
```diff
@@ -0,0 +1,365 @@
...
+```
+create-issue-gitbug-id = "customfield_5678"
+```
+
+### Status Map
+
+You can specify the mapping between `git-bug` status and JIRA status id's using
+the following:
+```
+bug-open-id = 1
```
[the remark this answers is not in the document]
cheshirekow: Done.

doc/jira_bridge.md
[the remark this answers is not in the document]
MichaelMure: Lamport clock are used to order bugs with each other (creation and last edition), but within a bug the order of the operation are what matter. You can see it as a sort of blockchain where the more operation you add, the more you prevent tempering the history. That said, I'm considering at the moment to relax this purely linear chain constraint and have a DAG of Operation to get a more robust conflict merge (see https://github.com/MichaelMure/git-bug/issues/212 but it's quite messy). If that happen, Lamport clock or similar might have a bigger role to play.

In any case, bridges should consider the data structure as append only.

> If the bridge already has a comment, but it observes a new revision, it does not import a whole new comment, it edits the existing comment.

My comment still stand: if a comment edition is generated for each import while no change happened on the Jira side, it's going to be problematic.

bridge/jira/import.go
```diff
@@ -0,0 +1,613 @@
...
+
+// Return a unique string derived from a unique jira id and a timestamp
+func getTimeDerivedID(jiraID string, timestamp MyTime) string {
+	return fmt.Sprintf("%s-%d", jiraID, timestamp.Unix())
+}
+
+// Create a bug.Comment from a JIRA comment
+func (self *jiraImporter) ensureComment(
+	repo *cache.RepoCache, b *cache.BugCache, item Comment) error {
+	// ensure person
```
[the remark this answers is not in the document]
MichaelMure: Usually, go code would be wrapped like this:
```
// Create a bug.Comment from a JIRA comment
func (self *jiraImporter) ensureComment(repo *cache.RepoCache,
	b *cache.BugCache, item Comment) error {
```

But that is if you really want to. I don't think any other function call is wrapped like this in git-bug. Clearly nitpicking as well.

bridge/jira/import.go
MichaelMure: The goal is to have a stream of event that UIs can interpret, filter and display as they want. The "nothing" event say "we have this bug but nothing has changed, it's up to date". Emitting an event for "no title has been changed" would not be very useful.

Regarding your question, I'd say it depends on how sever the problem is. Does the user reasonably expect the import to work ? In that case that would be an error. If not ... discussing with @A-Hilaly we considered adding a warning event. Maybe that would be a good way to do that. The error would end the import process of a bug and optionally of the whole import, a warning would not.

bridge/jira/export.go
MichaelMure: haha, I see we went to the same direction. Let's do the warning then.

CC @A-Hilaly

MichaelMure:
So much teasing and no push ;)

cheshirekow:
> So much teasing and no push ;)

Ah, yeah sorry. I can push when I am home tonight. I wanted to test and then report on the test results with the next update, but now I have to deal with this:

image

On one hand, it's good they support user API tokens... On the other hand, we'll probably need both code paths because companies are slow to upgrade their JIRA deployments.

cheshirekow:
I did a little testing with my jira cloud instance and things are looking good. I did:

1. create a git-bug with a label, export to jira
2. add  a comment in jira, import to git-bug
3. edit the description in git-bug, push to jira

I had to hack in support for the new jira token auth. I'll need to review the new git-bug token management API to get that plumbed in correctly (not hacky) though... and do a little more comprehensive testing. Feel free to comment further on the current state of the patch though. I don't think too much will change to get in the token auth support.

MichaelMure:
> but now I have to deal with this

Ha bummer :( Github is doing that as well ... https://developer.github.com/changes/2019-11-05-deprecated-passwords-and-authorizations-api/

bridge/jira/export.go
```diff
@@ -0,0 +1,455 @@
...
+// UpdateIssueStatus attempts to change the "status" field by finding a
+// transition which achieves the desired state and then performing that
+// transition
+func UpdateIssueStatus(
+	client *Client, issueKeyOrID string, desiredStateNameOrID string) (
+	time.Time, error) {
+
+	var responseTime time.Time
+
+	tlist, err := client.GetTransitions(issueKeyOrID)
```
MichaelMure: really not an emergency but this could be cached at some point

bridge/jira/import.go
```diff
@@ -0,0 +1,600 @@
...
+
+		// add comment operation
+		op, err := b.AddCommentRaw(
+			author,
+			item.Created.Unix(),
+			cleanText,
+			nil,
+			map[string]string{
+				keyJiraID:      item.ID,
+				keyJiraProject: self.conf[keyProject],
```
MichaelMure: hmm, is it useful to tag operations with the project if the bug itself is already tagged ?

bridge/jira/export.go
```diff
@@ -0,0 +1,455 @@
...
+					out <- core.NewExportError(err, b.Id())
+					return err
+				}
+				out <- core.NewExportCommentEdition(op.Id())
+				// JIRA doesn't track all comment edits, they will only tell us about
+				// the most recent one. We must invent a consistent id for the operation
+				// so we use the comment ID plus the timestamp of the update, as
+				// reported by JIRA. Note that this must be consistent with the importer
+				// during ensureComment()
+				id = fmt.Sprintf("%s-%d", comment.ID, comment.Updated.Unix())
```
MichaelMure: I think you can use `getTimeDerivedID` here

MichaelMure:
I'm working on generalizing the credential code in the core so it's easy to add more kind.

bridge/jira/export.go
cheshirekow: Done

bridge/jira/export.go
cheshirekow: Indeed (bug `8acce9c`), but it's not *exactly* simple. What we can do is download the full state graph ("workflow") but then we'll need to include logic to figure out what transitions are actually currently valid for an issue. OR we can just ask JIRA which transitions are currently valid (that's what this API endpoint does). 

Not going round-trip through the REST API would be faster, but it is also *slightly* racy in that the cached workflow *might* be stale and it would require git-bug implementation of the validation logic that JIRA implements. In practice workflows don't change frequently so it's probably the better choice, but there is some actual design work to be done there so I'm holding off on this.

bridge/jira/import.go
```diff
@@ -0,0 +1,613 @@
...
+
+	"github.com/MichaelMure/git-bug/bridge/core"
+	"github.com/MichaelMure/git-bug/bug"
+	"github.com/MichaelMure/git-bug/cache"
+	"github.com/MichaelMure/git-bug/entity"
+	"github.com/MichaelMure/git-bug/util/text"
+)
+
+const (
+	keyOrigin          = "origin"
```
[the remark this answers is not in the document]
cheshirekow: Done

bridge/jira/import.go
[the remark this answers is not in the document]
cheshirekow: Oh interesting. In the languages I usually work with splitting arguments at different indention would be frowned upon. Good to know though, I'll go ahead and fix this one here and try to follow that moving forward.

bridge/jira/import.go
cheshirekow: I can't think of a reason... seems like it's not useful. I will remove the project tag here and in other operations.

doc/jira_bridge.md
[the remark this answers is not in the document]
cheshirekow: > My comment still stand: if a comment edition is generated for each import while no change happened on the Jira side, it's going to be problematic.

This is not the case. The logic is:

>  If the bridge already has a comment, but it observes a **new revision**, it does not import a whole new comment, it edits the existing comment.

The logic should not create a new comment edition for each import. It should create a new comment edition only if it sees an previously un-imported JIRA "revision" in the changelog... And in testing I believe this is accurately implemented.

bridge/jira/client.go
```diff
@@ -0,0 +1,1495 @@
...
+	if username == "" {
+		return fmt.Errorf(
+			"Invalid configuration lacks both a username and credentials sidecar " +
+				"path. At least one is required.")
+	}
+
+	password := conf[keyPassword]
+	if password == "" {
+		var err error
+		password, err = input.PromptPassword()
```
golangcibot: ineffectual assignment to `password` (from `ineffassign`)

doc/jira_bridge.md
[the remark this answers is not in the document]
MichaelMure: perfect then :+1:

bridge/jira/export.go
MichaelMure: yeah, let see if this is actually a performance problem first.

cheshirekow:
Did some testing last night and fixed a couple of bugs. I still haven't rebased on the new credential system, so I guess that is up next. Things are looking good so far though and I think will start using this at work to shake out more bugs and inspire a test plan. 

I'm going to resolve the open conversations on the review and make sure I didn't miss anything.

doc/jira_bridge.md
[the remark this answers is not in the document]
cheshirekow: Resolving this comment since you refactored the credential system. There will likely be more discussion after rebase.

bridge/jira/export.go
cheshirekow: Note to self: Also do this for un-implemented changelog events.

cheshirekow: This is done.

cheshirekow:
Update #5: I had to fix the git config parser in the core in order to correctly parse out a JSON map as a git config value. I also added the import/export warning events and used them. Lastly, I added a "reverse map" for bug ids, so that git-bug knows how to interpret JIRA statuses on import (if there are many).

MichaelMure:
Hey @cheshirekow what's up with this ? Do you need help ?

cheshirekow:
I rebased last week and resolved conflicts but it doesn't build due to interface requirements on the bridge. I guess you changed the bridge API? I haven't had time to dig in further.

As of the last commit all your comments have been addressed I think. It's fully functional and I've been using this branch as my daily driver for some weeks now. 

Some help adapting to whatever the upstream changes were would be great!

MichaelMure:
Alright, I'll finish https://github.com/MichaelMure/git-bug/pull/294 and rebase/merge this.

If you have the time, an automated test similar as the other bridges would be great.

MichaelMure:
First merge, use of the new prompts and some cleanup. This need more work especially on the credential part.

cheshirekow:
Thanks for the push. I built this morning and my pull/push from JIRA worked just fine.

MichaelMure:
@cheshirekow do you think it would be possible to detect automatically the credential (session or token) ? That would simplify both the code and the UX.

MichaelMure:
Also, there is a mismatch between the name and value of this constant, what's the good concept ? `metaKeyJiraOperationId = "jira-derived-id"`

cheshirekow:
> @cheshirekow do you think it would be possible to detect automatically the credential (session or token) ? That would simplify both the code and the UX.

Your idea in the other thread of trying some query with both credential types and just seeing what works makes sense. 

> Also, there is a mismatch between the name and value of this constant, what's the good concept ? metaKeyJiraOperationId = "jira-derived-id"

As I mentioned in the other thread, I'm not sure which is better to use, but I'm happy with either. Which do you prefer? I might suggest changing the name of the variable and keeping the metadata key just so I don't have to re-import my 1500-ish bugs that are already using "jira-derived-id" in their metadata. I'll go ahead and make that change.

cheshirekow:
Successful import from `git-bug.atlassian.net` and from my work jira so things look good with your credential changes. Thanks for pulling those in.

MichaelMure:
> But what if you need to use two different account for two different bridges?

So, that would only matter for importing private repo. For importing public repo, any credential would do. For exporting, the bridge will match based on the login, using whatever credential is needed for this specific piece of data.

So what could be done is for the bridges to store an additional information in their config: the login to use when importing. That would make more sense then using the first random one found.

MichaelMure:
So, does that means it's merge time ? What's left to do here ?

MichaelMure:
Ho, at least one thing would be to complete the Readme

cheshirekow:
> So, does that means it's merge time ? What's left to do here ?

Ready from my end. 

> Ho, at least one thing would be to complete the Readme

You mean add "Jira" to the tables? Ok, I just pushed a change to do that.

MichaelMure:
Weeeeeeeeeee :tada: :tada: :tada: :tada: :tada: :tada:

PR #254, opened 2019-11-18
Bridge: correctly emit NothingEvents

### LINKED ISSUE #231: import/export: the "nothing" event should be emitted only when nothing happen on a bug, not on an operation

A-Hilaly:
fix #231


MichaelMure:
CC @cheshirekow

cheshirekow:
> CC @cheshirekow

Ok so I see that a lot of the `NewExportNothing` events are dropped in this review. I was definitely copying their current placement so whats the semantics of the `NewExportNothing`? Should it only be used when a whole bug cannot be imported? Should it not be used when a particular operation cannot be imported?

Issue #235, opened 2019-10-24
termui shows stale bug status


cheshirekow:
I'm not sure how to provide relevant information for this issue, but if there's anything I can do to help debug let me know. The problem is that I have closed a few bugs from the command line using `git bug status close <id>`. After that, I started the termui, but it is displaying those bugs as "open". I'm using a filter `status:open label:json`. However, when I press enter to show details of that bug, it says "closed". Furthermore, when I go back to the list (press "q") that bug is no longer in the open list. I took a screencast:

![termui_stale](https://user-images.githubusercontent.com/1309820/67522325-6b104600-f661-11e9-8e63-7a32dba65e94.gif)

In the screencast note bugs `12fc493` and `71ac52e` at the bottom of the list. They are both listed as "open". Then I press "enter" to view the status. Note it says "closed". Then I press "q" and go back to the list. Note that the bug has disappeared from the list. 

This is observed on `master` from today. I deleted that and went back to the prebuilt binary from `v0.5.0` so I will update if the problem seems to go away or persists. 

I will note that these labels were possibly assigned "near" a termui crash (#233) so that *might* be related.

cheshirekow:
Yup, issue persists with `v0.5.0` (though maybe crossing the version boundary is the problem? I'm not sure how far master is from `0.5.0`). In the screencast above note bug `92c0d09` (also near the bottom). This bug is closed and when I open termui with `v0.5.0` it appears open, as it did with `master` (in the video above).

MichaelMure:
The termui list is only using the cache as a source of data. When you open a bug, the whole series of commit is loaded from git and the cache is updated (I think ... not sure if it's actually updated if you don't make any change to the bug). This suggest that your cache is somehow not reflecting the actual state of those bugs. A crash can very much do that. 

To check that, you can either simply delete your cache (`.git/git-bug/bug-cache`), or make a change to those bugs to force the update of the cache.

cheshirekow:
Ok great! I'll hold off on deleting the cache and try to spot any more bugs that incorrectly show up as open. Then I'll delete the cache and see if they show up as closed. I will report back.

cheshirekow:
Ok I got a situation where this happened again. Indeed deleting the cache seems to have fixed the stale status. However, I don't think the stale status was due to crash this time. I'm pretty sure that this happened after two pulls, with no local edits to the bug. Here is what I think the situation was:

Machine A: 
* Create bug `d3c0372`
* `git bug push`
[...]

PR #257, opened 2019-11-18
Core fixes

MichaelMure:
A single line fix for #235 and a bunch of esthetic fixes.


bug/bug.go
```diff
@@ -592,6 +587,8 @@ func (bug *Bug) Merge(repo repository.Repo, other Interface) (bool, error) {
 		bug.lastCommit = hash
 	}
 
+	bug.packs = newPacks
```
MichaelMure: fix is here

PR #260, opened 2019-11-23
Support bridge configuration with global tokens

A-Hilaly:
Support bridge configuration with global tokens for the following targets:

- [x] Github
- [x] Gitlab

### DIFF
--- a/bridge/core/bridge.go
+++ b/bridge/core/bridge.go
@@ -13,6 +13,7 @@ import (
 	"github.com/pkg/errors"
 
 	"github.com/MichaelMure/git-bug/cache"
+	"github.com/MichaelMure/git-bug/entity"
 	"github.com/MichaelMure/git-bug/repository"
 )
 
@@ -20,8 +21,10 @@ var ErrImportNotSupported = errors.New("import is not supported")
 var ErrExportNotSupported = errors.New("export is not supported")
 
 const (
-	ConfigKeyTarget = "target"
-	MetaKeyOrigin   = "origin"
+	ConfigKeyTarget  = "target"
+	ConfigKeyToken   = "token"
+	ConfigKeyTokenId = "token-id"
+	MetaKeyOrigin    = "origin"
 
 	bridgeConfigKeyPrefix = "git-bug.bridge"
 )
@@ -35,6 +38,7 @@ type BridgeParams struct {
 	Project    string
 	URL        string
 	Token      string
+	TokenId    string
 	TokenStdin bool
 }
 
@@ -275,6 +279,13 @@ func (b *Bridge) ensureInit() error {
 	if b.initDone {
 		return nil
 	}
+
+	token, err := LoadToken(b.repo, entity.Id(b.conf[ConfigKeyTokenId]))
+	if err != nil {
+		return err
+	}
+
+	b.conf[ConfigKeyToken] = token.Value
 
 	importer := b.getImporter()
 	if importer != nil {
--- a/bridge/github/config.go
+++ b/bridge/github/config.go
@@ -21,6 +21,7 @@ import (
 	"golang.org/x/crypto/ssh/terminal"
 
 	"github.com/MichaelMure/git-bug/bridge/core"
+	"github.com/MichaelMure/git-bug/entity"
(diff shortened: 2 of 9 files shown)


bridge/github/config.go
```diff
@@ -220,35 +262,77 @@ func randomFingerprint() string {
...
+		} else {
+			return githubTokens[index-2], nil
 		}
 
-		return loginAndRequestToken(owner, project)
+		return loadOrCreateToken(repo, token)
 	}
 }
 
+func tokenAlreadyExist(repo repository.RepoCommon, id string) (bool, error) {
```
golangcibot: func `tokenAlreadyExist` is unused (from `unused`)

bridge/github/config.go
```diff
@@ -87,27 +90,45 @@ func (g *Github) Configure(repo repository.RepoCommon, params core.BridgeParams)
...
 		if err != nil {
 			return nil, fmt.Errorf("reading from stdin: %v", err)
 		}
 		token = strings.TrimSuffix(token, "\n")
+	} else if params.TokenId != "" {
+		tokenId = entity.Id(params.TokenId)
 	} else {
-		token, err = promptTokenOptions(owner, project)
+		tokenObj, err = promptTokenOptions(repo, owner, project)
+		if err != nil {
```
MichaelMure: all those if/else are more complex than needed but that should work

bridge/github/config.go
```diff
@@ -128,15 +149,36 @@ func (g *Github) Configure(repo repository.RepoCommon, params core.BridgeParams)
 	return conf, nil
 }
 
+func loadOrCreateToken(repo repository.RepoCommon, tokenValue string) (*core.Token, error) {
```
MichaelMure: you could migrate that to the core

bridge/github/config.go
```diff
@@ -220,32 +262,59 @@ func randomFingerprint() string {
...
+		}
+
 		fmt.Println()
 		fmt.Println("[1]: user provided token")
 		fmt.Println("[2]: interactive token creation")
+		fmt.Println("known tokens for Github:")
+
+		var githubTokens []*core.Token
+		i := 0
+		for _, token := range tokens {
```
MichaelMure: i think you can use the index given by the range

bridge/github/config.go
```diff
@@ -220,32 +262,59 @@ func randomFingerprint() string {
...
 	for {
+		tokens, err := core.LoadTokens(repo)
+		if err != nil {
+			return nil, err
+		}
+
 		fmt.Println()
 		fmt.Println("[1]: user provided token")
 		fmt.Println("[2]: interactive token creation")
+		fmt.Println("known tokens for Github:")
```
MichaelMure: don't write this if you don't have any valid token

bridge/github/config.go
MichaelMure: also, it seems you need something like `LoadTokenWithTarget` in the core to simplify that

bridge/github/config.go
```diff
@@ -220,32 +262,59 @@ func randomFingerprint() string {
...
-		if err != nil || (index != 1 && index != 2) {
+		if err != nil || index < 1 || index > len(githubTokens)+2 {
 			fmt.Println("invalid input")
 			continue
 		}
 
+		var token string
 		if index == 1 {
-			return promptToken()
+			token, err = promptToken()
```
MichaelMure: use more switch :)

bridge/core/bridge.go
```diff
@@ -276,6 +280,13 @@ func (b *Bridge) ensureInit() error {
 		return nil
 	}
 
+	token, err := LoadToken(b.repo, entity.Id(b.conf[ConfigKeyTokenId]))
```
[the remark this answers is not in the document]
A-Hilaly: For the moment yes. Later we can have multiple auth kinds

bridge/github/config.go
A-Hilaly: Since the tokens list can contain tokens with different targets doesn't provide the correct index to show. `LoadTokenWithTarget` will definitely solve this :+1:

PR #271, opened 2019-12-08
bridge: huge refactor to accept multiple kind of credentials

MichaelMure:
This a (huge) refactor to be able to handle different kind of credentials (token, login/password ...) in the bridges. What started as a fairly normal generalization actually had an extensive impact, especially because:
- bridges don't store their credential anymore, they load it from the user ID (default or not)
- token now store the user ID
- the starting user workflow change from `config a bridge --> pull --> user adopt` to `create a user --> config a bridge --> pull`

@cheshirekow have a look, you should now be able to add a new kind of credential by copying `bridge/core/auth/token.go`. This include:
- deported file: the config would store the file path, with a helper to read the credential
- login/password
- OAuth


bridge/github/export.go
```diff
@@ -46,77 +52,95 @@ type githubExporter struct {
...
+	creds, err := auth.List(repo, auth.WithUserId(user.Id()), auth.WithTarget(target), auth.WithKind(auth.KindToken))
+	if err != nil {
+		return err
+	}
+
+	if len(creds) == 0 {
+		return ErrMissingIdentityToken
+	}
+
+	ge.defaultToken = creds[0].(*auth.Token)
```
A-Hilaly: Isn't defaultClient already loaded at the first place if err is nil ?

bridge/github/config.go
```diff
@@ -141,7 +124,6 @@ func (g *Github) Configure(repo repository.RepoCommon, params core.BridgeParams)
 	}
 
 	conf[core.ConfigKeyTarget] = target
-	conf[core.ConfigKeyTokenId] = tokenObj.ID().String()
```
A-Hilaly: If my understanding is correct bridges doesn't store the auth id anymore ? git-bug is responsible of finding the user token before the import and export operations

bridge/core/auth/credential.go
```diff
@@ -0,0 +1,232 @@
...
+		newKey := strings.TrimPrefix(key, keyPrefix)
+		configs[newKey] = value
+	}
+
+	var cred Credential
+
+	switch configs[configKeyKind] {
+	case KindToken:
+		cred = NewTokenFromConfig(configs)
+	case KindLoginPassword:
```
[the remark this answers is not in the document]
MichaelMure: it's not, left as an exercise to the reader

bridge/core/auth/credential.go
```diff
@@ -0,0 +1,232 @@
...
+	ID() entity.Id
+	UserId() entity.Id
+	Target() string
+	Kind() CredentialKind
+	CreateTime() time.Time
+	Validate() error
+
+	// Return all the specific properties of the credential that need to be saved into the configuration.
+	// This does not include Target, User, Kind and CreateTime.
+	ToConfig() map[string]string
```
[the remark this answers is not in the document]
MichaelMure: It's implemented once and for all in https://github.com/MichaelMure/git-bug/pull/271/files?file-filters%5B%5D=.go&hide-deleted-files=true#diff-f67b0a00aca1defb21f5187bfd593674R170-R208

Each kind of credential just have to deal with its custom values.

bridge/core/auth/credential.go
```diff
@@ -0,0 +1,232 @@
...
+
+// LoadWithPrefix load a credential from the repo config with a prefix
+func LoadWithPrefix(repo repository.RepoConfig, prefix string) (Credential, error) {
+	creds, err := List(repo)
+	if err != nil {
+		return nil, err
+	}
+
+	// preallocate but empty
+	matching := make([]Credential, 0, 5)
```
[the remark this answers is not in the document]
MichaelMure: It's just a "good enough for 99%" of the cases value. `append` will reallocate if its not enough. Sticking to a power of 2 would probably not change anything for such a small value.

bridge/core/auth/options.go
```diff
@@ -0,0 +1,62 @@
+package auth
+
+import (
+	"github.com/MichaelMure/git-bug/entity"
+	"github.com/MichaelMure/git-bug/identity"
+)
+
+type options struct {
```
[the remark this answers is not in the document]
MichaelMure: It's a relatively common pattern in go once there is too much variant of a function. Quite clean but also clutter the namespace ..

bridge/github/config.go
MichaelMure: correct

bridge/github/export.go
MichaelMure: It's needed because the github exporter use 2 kind of client: the graphQL one, and a bunch of manual HTTP requests which need this token directly

A-Hilaly: Oh i see, forgot about that 🤦‍♂

PR #280, opened 2019-12-26
bridge/gitlab: add base url configuration

A-Hilaly:
- [x] bridge/gitlab: add missing baseUrl prompt 
- [x] bridge/gitlab: fix api calls with self hosted Gitlab instances


bridge/gitlab/config.go
```diff
@@ -126,14 +134,74 @@ func (g *Gitlab) ValidateConfig(conf core.Configuration) error {
...
+			return defaultBaseURL, nil
+		case 1:
+			return promptBaseUrl()
+		}
+	}
+}
+
+func promptBaseUrl() (string, error) {
+	for {
+		fmt.Print("base url: ")
```
MichaelMure:
```suggestion
		fmt.Print("Base url: ")
```

Issue #284, opened 2020-01-03
Pushing to a bridge does not push comments added via git-bug.


activeshadow:
## To Reproduce

1. Pull from bug tracker (in my case, self-hosted GitLab) using bridge
1. Adopt an identity created when pulling via bridge
1. Comment on one of the existing issues locally using git-bug
1. Push to bug tracker using bridge

## Expected

Comment added to issue locally using git-bug should be present on issue in bug tracker (based on [this](https://github.com/MichaelMure/git-bug#exporter-implementations)).

## Actual

Comment not present on issue in bug tracker, but present when inspecting the issue locally using git-bug.

I've only tested with the GitLab bridge (and a self-hosted version of GitLab at that), so I'm not sure if the same thing happens with the other available bridges (or with the main gitlab.com).

MichaelMure:
Fix in https://github.com/MichaelMure/git-bug/pull/285

activeshadow:
@MichaelMure thank you for the quick turn-around! I'm assuming my issue here was due to the use of a self-hosted GitLab instance. Would this also affect issue labels? When I pulled from GitLab none of the labels were pulled, and when I labeled issues locally and pushed the labels didn't show up in GitLab.

MichaelMure:
That might be another problem. Could you try the `fix-gitlab` branch and see if that fix it as well ?

Thanks for dealing with the bleeding edge, that's very helpful.

activeshadow:
@MichaelMure OK so comments made locally are now being pushed to my self-hosted GitLab. Thank you!

It also looks like labels are being synced now... kinda. I say kinda because pulling labels from issues in GitLab seems to be hit or miss. Some issues locally will show labels added in GitLab, some won't. As an example, I had added some labels locally to issues that had labels in GitLab when they didn't get pulled initially (ie. before using the `fix-gitlab` branch). When i pushed to GitLab using the `fix-gitlab` branch, the local labels overwrote the remote labels completely, so I went and added the original labels back in GitLab. I then pulled from GitLab and only some of the original labels I added back got pulled. For other issues in GitLab that have labels already, they're not getting pulled down.

Also, I think the GitLab bridge is pulling deleted comments. I had a few comments locally that got pushed when I tried the `fix-gitlab` branch that I then deleted in GitLab since I had added them there too when they weren't pushed previously. I then pulled from GitLab and the deleted comments were pulled, meaning I now have two identical comments locally. This likely constitutes a separate issue.

[...]

PR #285, opened 2020-01-03
Fix gitlab

MichaelMure:


bridge/gitlab/export.go
```diff
@@ -159,7 +159,7 @@ func (ge *gitlabExporter) exportBug(ctx context.Context, b *cache.BugCache, sinc
 	gitlabID, ok := snapshot.GetCreateMetadata(metaKeyGitlabId)
 	if ok {
 		gitlabBaseUrl, ok := snapshot.GetCreateMetadata(metaKeyGitlabBaseUrl)
-		if ok && gitlabBaseUrl != ge.conf[gitlabBaseUrl] {
+		if ok && gitlabBaseUrl != ge.conf[keyGitlabBaseUrl] {
```
MichaelMure: fix is here

MichaelMure:
Reported as fixing the bug in https://github.com/MichaelMure/git-bug/issues/284#issuecomment-570645407

bridge/gitlab/export.go
A-Hilaly: Nice catch, this can definitely have stopped git-bug from exporting bugs

PR #287, opened 2020-01-04
Github import error

### LINKED ISSUE #286: import error: issue creation: no matching operation found
Looks like bugs/issues created on the webpage cannot be imported.

```bash
$ git bug bridge pull
import error: issue creation: no matching operation found
imported 0 issues and 0 identities with default bridge
```

Versions should be latest `master`

MichaelMure:
This PR is an attempt to fix #286 

My understanding is that when trying to import a comment edition, if the original comment is nowhere to be found locally the import will hard fail with an error. This kinda make sense because the comment really should be there but that also an assumption we can relax a bit I think. A warning might be enough.

@cheshirekow I cherry-picked from your PR the two commits introducing warning in the bridge's core (one completely, one while remove the Jira related changes). Rebasing your PR should go without too much though.

### DIFF
--- a/bridge/core/bridge.go
+++ b/bridge/core/bridge.go
@@ -347,7 +347,7 @@ func (b *Bridge) ImportAllSince(ctx context.Context, since time.Time) (<-chan Im
 
 		// relay all events while checking that everything went well
 		for event := range events {
-			if event.Err != nil {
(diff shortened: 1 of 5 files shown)


cheshirekow:
:+1:

PR #296, opened 2020-01-15
Changed 'close' to 'status' in the README file

omusaev:
Just a quick doc fix. I was looking for 'close' and didn't find it :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,7 +94,7 @@ Filter and sort bugs using a [query](doc/queries.md):
 git bug ls "status:open sort:edit"
 ```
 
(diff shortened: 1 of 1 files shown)


MichaelMure:
`close` is really the command I think should be showcased here. `status` is really meant for scripting and is redundant with `show`.

Julian:
I encountered the same issue here -- is there actually a command called close? I don't appear to have one at least.

```
⊙  g bug close 7c92fc9                                                                                                                                                                            julian@Air
Error: unknown command "close" for "git-bug"
```

Julian:
(Same as open -- my git bug has no such command, so I assumed what was meant was `add`).

PR #299, opened 2020-01-23
Fix column overflow in webui

ludovicm67:
If an image or a code block was too large the design of a bug message was broken.
This PR also adds support to open images in a new tab on a click.


MichaelMure:
:tada:

PR #301, opened 2020-01-25
Issue list improvements and filtering

sandhose:
This implements basic filters (right now only open/closed issues) and improves the issue list page.

Paging is now implemented with query params, so reloading the page displays the same issues.

I implemented the query filters parsing using regexs ; this is not exactly the same as in the backend but right now it’s Good Enough™.

Because the query parsing/stringifying is a bit complex, I wrote tests for it. Those are the first tests in the webui (🎉), although they are not run by the CI right now.

Please note that the label/author/sort filters are not working right now (mainly because I need to do some kind of drop downs for that).

Also check the `TODO`s in the code.

### DIFF
--- a/webui/package-lock.json
+++ b/webui/package-lock.json
@@ -2086,14 +2086,14 @@
       }
     },
     "@material-ui/core": {
-      "version": "4.8.3",
-      "resolved": "https://registry.npmjs.org/@material-ui/core/-/core-4.8.3.tgz",
-      "integrity": "sha512-ZJbfJQmkuZCSQTf0nzpfZwizmDdCq8ruZxnPNFnhoKDqgJpMvV8TJRi8vdI9ls1tMuTqxlhyhw8556fxOpWpFQ==",
+      "version": "4.9.0",
+      "resolved": "https://registry.npmjs.org/@material-ui/core/-/core-4.9.0.tgz",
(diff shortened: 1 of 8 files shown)


MichaelMure:
My take on that is that if/when full text search happen in git-bug, it will be a "simple" word search. The reason is that for performance we will need to build some kind of word index/count in the cache, which means loosing the context and order of each words. Note that it's also how Github works and most likely many others. It seems good enough for me.

Ultimately, that means we don't need extra complexity (multiple quote characters, escaping quotes) in the query itself. Having burnt myself with the feature creep recently for the same reason ("it's easy I'll just do it"), I would advise to go the simple route here.

sandhose:
> Ultimately, that means we don't need extra complexity (multiple quote characters, escaping quotes) in the query itself. Having burnt myself with the feature creep recently for the same reason ("it's easy I'll just do it"), I would advise to go the simple route here.

Well this query parsing is really just used to manipulate it and display the filters correctly ; it does not work exactly the same as in the backend. If a query is invalid, even if the webui parses it, the backend throws an error, which is then displayed in the UI. I don't think it is really an issue if the webui parses more thing than the backend ; the other way around would be more problematic.

---

Aside from that, I implemented sorting & issue count. I also commented out the `author` and `label` filters because they are not implemented for now. Right now it looks like this

Bug list view

I think it's ready to merge.

MichaelMure:
> If a query is invalid, even if the webui parses it, the backend throws an error, which is then displayed in the UI.

Sure but ideally that doesn't need a roundtrip with the server ;)

Awesome work, thanks a lot :)

PR #316, opened 2020-02-08
graphql: use the cache in priority for fast browsing in the WebUI

MichaelMure:
DO NOT MERGE, depend on bugs to be fixed in gqlgen first.

Browsing the webUI bug list goes from > 1s to ~20ms


graphql/models/lazy_bug.go
```diff
@@ -0,0 +1,214 @@
...
+	return lb.snap.Timeline, nil
+}
+
+func (lb *LazyBug) Operations() ([]bug.Operation, error) {
+	err := lb.load()
+	if err != nil {
+		return nil, err
+	}
+	result := make([]bug.Operation, len(lb.snap.Operations))
+	for i, operation := range lb.snap.Operations {
```
golangcibot: S1001: should use copy() instead of a loop (from `gosimple`)

PR #321, opened 2020-02-12
auth: refactor and introduce Login and LoginPassword, salt IDs

MichaelMure:
@cheshirekow this PR introduce two new kind of credentials: LoginPassword and Login

Bridges can use those to store either a tuple Login+Password or just the Login. Bridges have the liberty and the responsibility to exploit them as make sense. In particular, the Jira bridge will need to ask the user for the password when required if Login is used.

Sounds good to you ? Do you think it's really needed to store those in a different place than the git config as you have done in the Jira bridge ?

### DIFF
--- a/bridge/core/auth/login.go
+++ b/bridge/core/auth/login.go
@@ -1,0 +1,67 @@
+package auth
+
+import (
+	"crypto/sha256"
+	"fmt"
+
+	"github.com/MichaelMure/git-bug/entity"
+)
+
+const (
+	configKeyLoginLogin = "login"
+)
+
+var _ Credential = &Login{}
+
+type Login struct {
+	*credentialBase
+	Login string
+}
+
+func NewLogin(target, login string) *Login {
+	return &Login{
+		credentialBase: newCredentialBase(target),
+		Login:          login,
+	}
+}
+
+func NewLoginFromConfig(conf map[string]string) (*Login, error) {
+	base, err := newCredentialBaseFromConfig(conf)
+	if err != nil {
+		return nil, err
+	}
+
+	return &Login{
+		credentialBase: base,
+		Login:          conf[configKeyLoginLogin],
+	}, nil
+}
+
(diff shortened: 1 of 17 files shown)


bridge/core/auth/login.go
```diff
@@ -0,0 +1,67 @@
...
+
+	return &Login{
+		credentialBase: base,
+		Login:          conf[configKeyLoginLogin],
+	}, nil
+}
+
+func (lp *Login) ID() entity.Id {
+	h := sha256.New()
+	h.Write(lp.salt)
```
golangcibot: Error return value of `h.Write` is not checked (from `errcheck`)

bridge/core/auth/login.go
```diff
@@ -0,0 +1,67 @@
...
+	h.Write([]byte(lp.target))
```
golangcibot: Error return value of `h.Write` is not checked (from `errcheck`)

bridge/core/auth/login.go
```diff
@@ -0,0 +1,67 @@
...
+	h.Write([]byte(lp.Login))
```
golangcibot: Error return value of `h.Write` is not checked (from `errcheck`)

cheshirekow:
Yeah, I would say it's pretty necessary to be able to store passwords outside of the git config. The number of times I've copy pasted contents out of the config file is high enough that I would worry about keeping secrets in it. 


BTW, it looks like git itself has a solution to this problem:
https://git-scm.com/docs/gitcredentials

There even appears to be an implemented helper for gnome-keyring. Seems like tying into this helper protocol would be a nice generic way to support all the things one's local git does...

cheshirekow:
Please have a quick look at #324. Given that git comes with `git-credential-cache` and `git-credential-store` you may want to reconsider having a separate credential system for git-bug and just reuse the one that git has already. 

There may still be value in what you've implemented in this change though, as it may be easier for users to understand and configure.

PR #324, opened 2020-02-14
Add git-credential adapter

cheshirekow:
Provides a Credential implementation that defers to [git-credential][1].

[1]: https://git-scm.com/docs/git-credential#IOFMT

This is primarily meant to demonstrate the technique in order to inform #321. The change works (I tested with gnome-keyring which is awesome!) however I think it would be better to use the existing `GitRepo.runGitCommand` API. There is no repository object available where the credentials are created though so perhaps some additional plumbing can be added as part of #321. 

The following unit-test works but it can't be added because it depends on the environment:

```
func TestGitCredential(t *testing.T) {
	//repo := repository.NewMockRepoForTest()

	conf := map[string]string{
		"target": "fake.atlassian.net",
	}

	cred1, err := NewGitCredentialFromConfig(conf)
	assert.NoError(t, err)
	assert.Equal(t, cred1.Password, "foobar")

	cred2, err := NewGitCredentialFromConfig(conf)
	assert.NoError(t, err)
	assert.Equal(t, cred1.Password, cred2.Password)
}
```

### DIFF
--- a/bridge/core/auth/gitcred.go
+++ b/bridge/core/auth/gitcred.go
@@ -1,0 +1,143 @@
+package auth
+
+import (
+	"bytes"
+	"crypto/sha256"
+	"fmt"
+	"io"
+	"os/exec"
+	"strings"
+	"time"
+
+	"github.com/MichaelMure/git-bug/bridge/core"
+	"github.com/MichaelMure/git-bug/entity"
+	"github.com/MichaelMure/git-bug/repository"
+)
+
+var _ Credential = &GitCredential{}
+
+// GitCredential the credentials
+// see: https://git-scm.com/docs/git-credential#IOFMT
+type GitCredential struct {
+	target     string
+	createTime time.Time
+	Username   string
+	Password   string
+	meta       map[string]string
+}
+
+// NewGitCredential instantiate a new object
+func NewGitCredential(target, username, password string) *GitCredential {
+	return &GitCredential{
+		target:     target,
+		createTime: time.Now(),
+		Username:   username,
+		Password:   password,
+	}
+}
+
+// NewGitCredentialFromConfig execute "git credential fill"
+func NewGitCredentialFromConfig(conf map[string]string) (*GitCredential, error) {
+	cred := &GitCredential{}
+
+	cred.target = conf[configKeyTarget]
+	if createTime, ok := conf[configKeyCreateTime]; ok {
+		if t, err := repository.ParseTimestamp(createTime); err == nil {
+			cred.createTime = t
+		}
+	}
+	cred.meta = metaFromConfig(conf)
+
+	cmd := exec.Command("git", "credential", "fill")
+	stdin, err := cmd.StdinPipe()
+	if err != nil {
+		return nil, err
+	}
+
+	var stdout bytes.Buffer
+	cmd.Stdout = &stdout
+
+	go func() {
+		defer stdin.Close()
+		io.WriteString(stdin,
+			fmt.Sprintf("protocol=https\nhost=%s\n\n", cred.target))
+	}()
+
+	err = cmd.Run()
+	if err != nil {
+		return nil, err
+	}
+
+	lines := strings.Split(stdout.String(), "\n")
+	for _, line := range lines {
+		if strings.TrimSpace(line) == "" {
+			continue
+		}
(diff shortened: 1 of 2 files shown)


bridge/core/auth/gitcred.go
```diff
@@ -0,0 +1,143 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	var stdout bytes.Buffer
+	cmd.Stdout = &stdout
+
+	go func() {
+		defer stdin.Close()
+		io.WriteString(stdin,
```
golangcibot: Error return value of `io.WriteString` is not checked (from `errcheck`)

MichaelMure:
Nice find, I had no idea it existed. Using real secret store instead of the git config is definitely a nice thing to have.

Toying a bit with that thing, I found that:
1. it's a way to interact with user-specified secret storage
1. if no credential helper is configured (virtually 100% of users) it default to a one off prompt. Nothing is stored, even between git invocation.
1. a cache system exist (need to be configured) to store credential in memory in between git invocation
1. a storage system exist (need to be configured) to store credential in a clear-text file
1. the expressiveness of this credential system is highly coupled with what git know: URLs, login, password. Storing tokens or metadata is kinda hacky.

So, what I derive from that is:
- this system is deeply tied to what git understand but could be abused for our needs. A standardized way to talk to secret store is neat, especially as it could require no git-bug specific configuration.
- instead of being a new kind of credential (token, password, login/password), this would really be a new way to store credentials for the credential system
- credential's metadata can keep being stored in the git config. Shoehorning those into the git credential system seems a bit too much
- in term of UX, an explicit choice between the git config and this more secure system is not required each time a credential is stored. It seems ok to me to use that better system if available and default to the git config if not. It may cause problems though if this system become unavailable and credentials are left in an incoherent state.

Does that sounds reasonable to you ? If so I'll merge my branch and we can add this feature later as it's orthogonal to everything else. Only the storage functions will be impacted.

cheshirekow:
A couple of comments:

> if no credential helper is configured (virtually 100% of users) it default to a one off prompt. Nothing is stored, even between git invocation.

This is true, but I will note that `git-bug` has to write *SOMETHING* to the config when storing a credential. Whether it writes the secrets directly or:
```
[credentials]
    helper = store
```
seems like it would have no impact on the user experience so it's pretty lightweight. 

> Storing tokens or metadata is kinda hacky.

Yeah I don't think storing metadata in `git credential` is possible in any sensible way. I think `username=<token>,password=XXXXXX` or `username=XXXX,password=` are both fine ways of storing a token. I wouldn't really call them hacky. 

> but could be abused for our needs

`git-credentials` stores three things: a url, a username, a password... which is pretty much the same [three things][1] as [LoginPassword][2] (a target, a username, a password). So it's not much of an abuse. 

[1]: https://github.com/MichaelMure/git-bug/pull/321/files#diff-9f0fd1fb9b91d12b743791a91a230ce0R11
[2]: https://github.com/MichaelMure/git-bug/pull/321/files#diff-ad9e4077825811334ad729e2269f3464R17

> instead of being a new kind of credential (token, password, login/password), this would really be a new way to store credentials for the credential system

Interesting... so you propose that "storage model" be a separate concept and each credential would need to implement an interface to enable it to be used with the `git-credential` storage model. Yeah, that seems reasonable. 

> Does that sounds reasonable to you ? If so I'll merge my branch and we can add this feature later as it's orthogonal to everything else. Only the storage functions will be impacted.

Yeah it does. The only thing is just that I think we'll want the `GitRepo` object available in `loadFromConfig`. It might be easier to deal with that plumbing now before the credential system grows more.

MichaelMure:
> So it's not much of an abuse.

Yes, I meant that as "using it for something it's not entirely designed for"

> The only thing is just that I think we'll want the GitRepo object available in loadFromConfig. It might be easier to deal with that plumbing now before the credential system grows more.

True, but it's not too far away. Ideally, this git credential thing would be abstracted behind another kind of Repo interface. Also, each kind of credential would implement something like `Secret() string` to separate that from the rest of the metadata. Probably some more adjustment here and there.

MichaelMure:
Closing in favor of https://github.com/MichaelMure/git-bug/issues/327

PR #329, opened 2020-02-15
jira: rework to use the credential system + adapt to refactors

MichaelMure:
@cheshirekow this PR adapt the Jira bridge to the credential system and to all the refactor I did recently. Please have a look

Note: It compiles but I can't test myself (automated tests would be veeery useful).
Note: the base of this PR is your branch pushed on this remote, otherwise I can't PR on it

A few questions:
- do you think it would be possible to detect automatically the credential (session or token) ? That would simplify both the code and the UX
- there is a mismatch between the name and value of this constant, what's the good concept ? metaKeyJiraOperationId = "jira-derived-id"
- it looks it would be very easy to support the official hosted Jira instances, it should simply be a default value in the baseURL prompt (see gitlab). What do you think ?

### DIFF
--- a/bridge/jira/config.go
+++ b/bridge/jira/config.go
@@ -1,6 +1,7 @@
 package jira
 
 import (
+	"context"
 	"fmt"
 
 	"github.com/MichaelMure/git-bug/bridge/core"
@@ -104,7 +105,7 @@ func (j *Jira) Configure(repo *cache.RepoCache, params core.BridgeParams) (core.
 	}
 
 	fmt.Printf("Attempting to login with credentials...\n")
-	client, err := buildClient(nil, baseURL, credType, cred)
+	client, err := buildClient(context.TODO(), baseURL, credType, cred)
 	if err != nil {
 		return nil, err
 	}
--- a/bridge/jira/config.go
+++ b/bridge/jira/config.go
@@ -1,15 +1,13 @@
 package jira
 
 import (
-	"encoding/json"
 	"fmt"
-	"io/ioutil"
-
-	"github.com/pkg/errors"
 
 	"github.com/MichaelMure/git-bug/bridge/core"
+	"github.com/MichaelMure/git-bug/bridge/core/auth"
 	"github.com/MichaelMure/git-bug/cache"
 	"github.com/MichaelMure/git-bug/input"
+	"github.com/MichaelMure/git-bug/repository"
 )
 
 const moreConfigText = `
@@ -28,32 +26,27 @@ after October 1st 2019 must use "TOKEN" authentication. You must create a user
 API token and the client will provide this along with your username with each
 request.`
 
-// Configure sets up the bridge configuration
-func (g *Jira) Configure(repo *cache.RepoCache, params core.BridgeParams) (core.Configuration, error) {
-	conf := make(core.Configuration)
-	conf[core.ConfigKeyTarget] = target
+func (*Jira) ValidParams() map[string]interface{} {
+	return map[string]interface{}{
+		"BaseURL":    nil,
+		"Login":      nil,
+		"CredPrefix": nil,
+		"Project":    nil,
+	}
+}
 
+// Configure sets up the bridge configuration
+func (j *Jira) Configure(repo *cache.RepoCache, params core.BridgeParams) (core.Configuration, error) {
 	var err error
 
-	// if params.Token != "" || params.TokenStdin {
-	// 	return nil, fmt.Errorf(
-	// 		"JIRA session tokens are extremely short lived. We don't store them " +
-	// 			"in the configuration, so they are not valid for this bridge.")
-	// }
-
-	if params.Owner != "" {
-		fmt.Println("warning: --owner is ineffective for a Jira bridge")
-	}
-
-	serverURL := params.URL
-	if serverURL == "" {
+	baseURL := params.BaseURL
+	if baseURL == "" {
 		// terminal prompt
-		serverURL, err = input.Prompt("JIRA server URL", "URL", input.Required)
+		baseURL, err = input.Prompt("JIRA server URL", "URL", input.Required, input.IsURL)
 		if err != nil {
 			return nil, err
 		}
 	}
-	conf[keyServer] = serverURL
 
 	project := params.Project
 	if project == "" {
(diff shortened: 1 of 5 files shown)


bridge/jira/config.go
```diff
@@ -62,77 +55,56 @@ func (g *Jira) Configure(repo *cache.RepoCache, params core.BridgeParams) (core.
...
+
+	err = j.ValidateConfig(conf)
 	if err != nil {
 		return nil, err
 	}
 
 	fmt.Printf("Attempting to login with credentials...\n")
-	client := NewClient(serverURL, nil)
-	err = client.Login(conf)
+	client, err := buildClient(nil, baseURL, credType, cred)
```
golangcibot: SA1012: do not pass a nil Context, even if a function permits it; pass context.TODO if you are unsure about which Context to use (from `staticcheck`)

MichaelMure: image

MichaelMure:
Note: the bridge gained the ability to reuse the same credentials when configuring a new bridge, without having to enter them again.

cheshirekow:
I'll test it out, thanks! I agree automated tests would be very useful.

cheshirekow:
> do you think it would be possible to detect automatically the credential (session or token) ? That would simplify both the code and the UX

Um... the data required for each is the same (username, password). I don't think there's a good way to detect whether or not a password is a token. The current JIRA implementation appears to generate tokens which are 24-characters strictly alphanumeric (`[A-Za-z0-9]{24}`) but I don't think that is documented or guaranteed. 

One thing we can do is:
1. If the server is `https://xxx.atlassian.net` then default to TOKEN
2. If the server is anything else then default to SESSION

> there is a mismatch between the name and value of this constant, what's the good concept ? metaKeyJiraOperationId = "jira-derived-id"

Hm... I'm not sure which is better. It is semantically an "Operation Identifier", but functionally it is a "derived identifier" (composed of multiple pieces of semi-unique information which together are unique). 

> it looks it would be very easy to support the official hosted Jira instances, it should simply be a default value in the baseURL prompt (see gitlab). What do you think ?

I think you would need an extra prompt then:

```
Is your jira server hosted at atlassian? [y/n]
What is the subdomain of your jira server? git-bug
```
resolves to `server=https://git-bug.atlassian.net`

If you're concerned about someone not knowing what the server URL is because it's a hosted instance, maybe it would just be better to add a note?
```
JIRA server URL (note: if your server is hosted at Atlassian, the server URL is in the form of https://[subdomain].atlassian.net). 
```

cheshirekow:
Looks like the credential config didn't get saved during the bridge config. 
```
josh@cobbler:~/Codes/git-bug-test$ git bug bridge configure 
[1]: github
[2]: gitlab
[3]: jira
[4]: launchpad-preview
target: 3
name [default]: 
JIRA server URL: https://git-bug.atlassian.net
JIRA project key: GBTC

JIRA has recently altered it's authentication strategies. Servers deployed
prior to October 1st 2019 must use "SESSION" authentication, whereby the REST
client logs in with an actual username and password, is assigned a session, and
passes the session cookie with each request. JIRA Cloud and servers deployed
after October 1st 2019 must use "TOKEN" authentication. You must create a user
API token and the client will provide this along with your username with each
request.
[1]: SESSION
[2]: TOKEN
Authentication mechanism: 2

JIRA login: <redacted>@gmail.com
[1]: enter my password
[2]: ask my password each time

Select option: 1

Password: 
Attempting to login with credentials...
Checking project ...

NOTE: There are a few optional configuration values that you can additionally
set in your git configuration to influence the behavior of the bridge. Please
see the notes at:
https://github.com/MichaelMure/git-bug/blob/master/doc/jira_bridge.md
Successfully configured bridge: default
josh@cobbler:~/Codes/git-bug-test$ git bug bridge pull
Error: no credential for this bridge
```

Here's the content of the git config afterward:
```
[core]
	repositoryformatversion = 0
	filemode = true
	bare = false
	logallrefupdates = true
[git-bug]
	identity = 834fb28a352d594e11811535b35e0fe0ae39df45
[git-bug "bridge.default"]
	target = jira
	base-url = https://git-bug.atlassian.net
	project = GBTC
	credentials-type = TOKEN
```

BTW I'm pretty sure I pushed some bugs to the git-bug JIRA which can probably be used to do some quick testing.

MichaelMure:
> Looks like the credential config didn't get saved during the bridge config

Credential are now stored in the (user) global git config so they can be reused from another repo. Try `git bug bridge auth` and under, or `git config --global -l`.

MichaelMure:
> I don't think there's a good way to detect whether or not a password is a token.

Not even an API call somewhere or a HTTP header in the response ? Or maybe try to login with both methods and pick whichever works ?

cheshirekow:
> Credential are now stored in the (user) global git config

They aren't there either (`~/.gitconfig`)
```
[user]
	name = <redacted>
	email = <redacted>
[mergetool]
	prompt = false
	keepBackup = false
[difftool]
	prompt = false
[merge]
	tool = meld
[credential]
	helper = gnome-keyring
```

Trying `git bug bridge auth` shows nothing:
```
josh@cobbler:~/Codes/git-bug-test$ git bug bridge auth
josh@cobbler:~/Codes/git-bug-test$ echo $?
0
```

Also this, which should be the same as the file above:
```
josh@cobbler:~/Codes/cheshiresoft-bugs$ git config --global -l
user.name=<redacted>
user.email=<redacted>
mergetool.prompt=false
mergetool.keepbackup=false
difftool.prompt=false
merge.tool=meld
credential.helper=gnome-keyring
```

In any case, note that `git bug bridge pull` did not work after configure:
```
josh@cobbler:~/Codes/git-bug-test$ git bug bridge pull
Error: no credential for this bridge
```

> Or maybe try to login with both methods and pick whichever works ?

That seems reasonable.

cheshirekow:
I see this in the github bridge but not the jira bridge, is this needed:
```
	// don't forget to store the now known valid token
	if !auth.IdExist(repo, cred.ID()) {
		err = auth.Store(repo, cred)
		if err != nil {
			return nil, err
		}
	}
```

cheshirekow:
Also, regarding: 
> Credential are now stored in the (user) global git config so they can be reused from another repo

How does a bridge know which one to use? Using this branch, I don't see anything in the user .gitconfig which associates the credential with anything other than `target=jira`. 

The way the github bridge looks up credentials:

```
func (gi *githubImporter) Init(_ context.Context, repo *cache.RepoCache, conf core.Configuration) error {
	gi.conf = conf

	creds, err := auth.List(repo, auth.WithTarget(target), auth.WithKind(auth.KindToken))
	if err != nil {
		return err
	}

	if len(creds) == 0 {
		return ErrMissingIdentityToken
	}

	gi.client = buildClient(creds[0].(*auth.Token))

	return nil
}
```
Looks to me like it will pick the first `github` credential out of the user configuration. But what if you need to use two different account for two different bridges?

MichaelMure:
> I see this in the github bridge but not the jira bridge, is this needed:

Yes, it's very much needed, there even is a `// don't forget` comment on it :man_facepalming:

MichaelMure:
> But what if you need to use two different account for two different bridges?

The github bridge doesn't support self-hosted instance at the moment.

cheshirekow:
Ok, with those two changes I think the configure step is fixed. I'm getting this now:

```
josh@cobbler:~/Codes/git-bug-test$ git bug bridge pull
So far so good. Have 6 issues to import
import error: issue creation: can't commit an identity with invalid data: avatarUrl is not a valid URL
imported 0 issues and 0 identities with default bridge
```
Does this `avatarUrl` validity thing look like something that may have changed in master? I don't recall writing that error message. 

> Note: the bridge gained the ability to reuse the same credentials when configuring a new bridge, without having to enter them again.

This did not seem to work. When I deleted the repo and started from scratch the config step asked me for everything again and It created a duplicate credential in my user config. I see the code in `jira/config.go` I think... but I'm not sure what went wrong. 

> The github bridge doesn't support self-hosted instance at the moment.

Just FYI: At my previous company I had a github account with my company email in addition to my personal account. We did not host our own github.

cheshirekow:
Hm... I can't seem to push to this branch so I pushed to the same branch name on my fork:
https://github.com/cheshirekow/git-bug/tree/jira-auth
Has one commit beyond this branch with the two fixes to save the credential and save the URL metadata with it.

MichaelMure:
As I said, I didn't test the workflow as I'm not using jira, so it's entirely possible there is more issue lurking.

> Does this avatarUrl validity thing look like something that may have changed in master?

nah, that didn't change in a while

cheshirekow:
> nah, that didn't change in a while 

That's weird, these bugs imported successfully on my branch. I'll look into this and get back to you. Do you mind if I go ahead and merge this pull request?

MichaelMure:
Please do.

cheshirekow:
>> Does this avatarUrl validity thing look like something that may have changed in master?

> nah, that didn't change in a while

Ah, you change the signature of `NewIdentityFull` and removed `login` so now that parameter is `avatarUrl` (commit 74e91144)

```
-func (c *RepoCache) NewIdentityFull(name string, email string, login string, avatarUrl string) (*IdentityCache, error) {
-       return c.NewIdentityRaw(name, email, login, avatarUrl, nil)
+func (c *RepoCache) NewIdentityFull(name string, email string, avatarUrl string) (*IdentityCache, error) {
+       return c.NewIdentityRaw(name, email, avatarUrl, nil)
 }
 
-func (c *RepoCache) NewIdentityRaw(name string, email string, login string, avatarUrl string, metadata map[string]string) (*IdentityCache, error) {
-       i := identity.NewIdentityFull(name, email, login, avatarUrl)
+func (c *RepoCache) NewIdentityRaw(name string, email string, avatarUrl string, metadata map[string]string) (*IdentityCache, error) {
+       i := identity.NewIdentityFull(name, email, avatarUrl)
+       return c.finishIdentity(i, metadata)
+}
 ```

PR #331, opened 2020-02-16
Webui: add comments

MichaelMure:
For when it's ready.


MichaelMure:
Merging as it does improve significantly the webUI :tada:

PR #344, opened 2020-02-25
github: fix matching cred/identity with badly cased login

MichaelMure:


MichaelMure:
Changed strategy: the github bridge now check and fix logins before it's used and stored. Should be good now.

PR #346, opened 2020-02-26
Fix compile errors on Windows

quincunx:
This fixes a compile error on Windows:

```
go generate
# github.com/MichaelMure/git-bug/input
input\prompt.go:94:37: cannot use syscall.Stdin (type syscall.Handle) as type int in argument to terminal.GetState
input\prompt.go:100:26: cannot use syscall.Stdin (type syscall.Handle) as type int in argument to terminal.Restore
input\prompt.go:108:45: cannot use syscall.Stdin (type syscall.Handle) as type int in argument to terminal.ReadPassword
git-bug.go:1: running "go": exit status 2
make: *** [Makefile:18: build] Error 1
```

### DIFF
--- a/input/prompt.go
+++ b/input/prompt.go
@@ -91,13 +91,13 @@ loop:
 
 // PromptPassword is a specialized text input that doesn't display the characters entered.
 func PromptPassword(prompt, name string, validators ...PromptValidator) (string, error) {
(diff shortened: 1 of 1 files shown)


MichaelMure:
Thanks :)

Issue #351, opened 2020-03-05
[jira] No way to adopt an identity


cheshirekow:
In summary of our gitter discussion:
With the new credentials system, there is no way to adopt an JIRA imported identity. The credentials that are created are associated with an identity that git-bug itself creates and there is no way to re-associate them after adopting an identity. 

@MichaelMure said that there should be an equivalent of `git bug bridge auth add-token` for login/password credentials.

MichaelMure:
That would be https://github.com/MichaelMure/git-bug/issues/336

That command would accept the credential as well as a user ID and the bridge login. It would then tag the user with the correct login metadata (the bridge expose the correct key to user).

This would solve this problem but it's not perfect either. I think we might also need a way to either:
- associate an existing credential to an existing user
- tag an existing user with a login for a specific bridge

It's almost the same thing but framed slightly differently. This might be part of those `add-xxx` commands or a different command, or even all of those merged into one.

Issue #336, opened 2020-02-21
feat: add "bridge auth add-login" and  "bridge auth add-loginpassword"


MichaelMure:
In the same vein as `bridge auth add-token`. There is most likely a fair amount of code that could be factorized between those.

edef1c:
Would this allow adopting an identity that has otherwise not interacted with the particular repository? This comment is partially a way of allowing me to adopt my GitHub identity for this repo, because I'm not sure if I can merge the two if I use the exporter — I certainly couldn't figure out how to add new information to an existing identity.

PR #352, opened 2020-03-05
Jira bridge users can adopt an identity in credential metadata

cheshirekow:
It sounds like @MichaelMure has a plan for the "correct" thing to do. But this was easy and it got me unblocked. It allows an `identity` to be specified in the metadata of a credential (e.g. in the git config). 

Fixes: #351

### DIFF
--- a/bridge/jira/export.go
+++ b/bridge/jira/export.go
@@ -101,16 +101,22 @@ func (je *jiraExporter) cacheAllClient(ctx context.Context, repo *cache.RepoCach
 			continue
 		}
 
-		user, err := repo.ResolveIdentityImmutableMetadata(metaKeyJiraLogin, login)
-		if err == identity.ErrIdentityNotExist {
-			continue
+		explicitIdentity, ok := cred.GetMetadata(auth.MetaKeyIdentity)
+		var user *cache.IdentityCache
+		if ok {
+			user, err = repo.ResolveIdentityPrefix(explicitIdentity)
+		} else {
+			user, err = repo.ResolveIdentityImmutableMetadata(metaKeyJiraLogin, login)
+			if err == identity.ErrIdentityNotExist {
+				continue
+			}
 		}
 		if err != nil {
 			return nil
 		}
 
 		if _, ok := je.identityClient[user.Id()]; !ok {
-			client, err := buildClient(ctx, je.conf[confKeyBaseUrl], je.conf[confKeyCredentialType], creds[0])
+			client, err := buildClient(ctx, je.conf[confKeyBaseUrl], je.conf[confKeyCredentialType], cred)
 			if err != nil {
 				return err
 			}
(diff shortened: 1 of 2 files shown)


bridge/jira/export.go
```diff
@@ -101,16 +101,22 @@ func (je *jiraExporter) cacheAllClient(ctx context.Context, repo *cache.RepoCach
 			continue
 		}
 
-		user, err := repo.ResolveIdentityImmutableMetadata(metaKeyJiraLogin, login)
-		if err == identity.ErrIdentityNotExist {
-			continue
+		explicitIdentity, ok := cred.GetMetadata(auth.MetaKeyIdentity)
```
MichaelMure: So, you associate the credential to an identity if the credential has been tagged with an identity metadata. This metadata was set in the version 0.6.0 when I wandered (incorrectly) towards this kind of tagging. They are not set anymore so it would only work as a "legacy" support, not as a solution onward.

Honestly I'm torn on this. Yes it's nice to stay compatible even for temporary mistake but at the same time git-bug is at the stage where it's ok to have light breaking changes instead of accumulating tech debt. Bridges are part of those as you just have to delete the config and set it up again. I understand your problem is a bit different but it seems to me that the correct solution would the the ones mentioned in #336 and #335 

Does that sounds correct to you ?

cheshirekow: >  I understand your problem is a bit different but it seems to me that the correct solution would the the ones mentioned in #336 and #335

Well, as you mentioned in #351:

> This would solve this problem but it's not perfect either. I think we might also need a way to either:
>
>associate an existing credential to an existing user
>tag an existing user with a login for a specific bridge

This is one way to associate an existing credential to an existing user. If the credential metadata is scheduled for removal then we can push the identity field into the main credential config instead of the meta config heading right? 

The alternative solution of tagging an existing user with a login for a specific bridge is fine too, but it sounds more complicated to implement and simple is nice. 

I'm happy with either solution. I also wouldn't mind if you take this solution for now and replace it with a better one later.

MichaelMure: > If the credential metadata is scheduled for removal then we can push the identity field into the main credential config instead of the meta config heading right?

It's not scheduled for removal. The wrong path I went in is to tag the credential with the identity ID. It lead to a bunch of problems because there is multiple scenario when configuring a bridge: you already have your identity, your identity will be imported by the bridge, ... It's a mess and doesn't work out. The correct way it seems is to tag both credential and identity with the bridge login and match them later.

My impression at the moment is that we just need 1) the missing command for the login/password credential and 2) some extra capability to manipulate those tags for the complex cases. I could be convinced otherwise though.

sudoforge:

```
$ git fetch https://github.com/git-bug/git-bug.git refs/pull/261/head
$ git checkout FETCH_HEAD
```

PR #355, opened 2020-03-14
replace the all-in-one query parser by a complete one with AST/lexer/parser

MichaelMure:
This PR introduce a new parser/compiler for the query DSL, with an intermediary representation in the form of an AST. By breaking down the previously single step into multiple stages (lexer/parser/compiler), it's now easier to test and improve, but more importantly, it's now possible to compile the AST into different target (bug matcher for the cache, query for a remote bugtracker in a bridge ...).

@cheshirekow it's a good first step towards https://github.com/MichaelMure/git-bug/pull/241. Could you have a look an see if you like it ?


MichaelMure:
> I have a slight concern that there will likely be a need to implement more complex logic.

Yeah I understand that. The current query language is inspired from the github one and is a compromise between expressiveness and shortness. As you can see in https://github.com/MichaelMure/git-bug/blob/query-parser-ast/cache/filter.go#L154-L184, the logic operator are implied, contextual and irregular: the one that should make the most sense is picked. Status, Author, Participant and Actor are OR combined, Label and Title are AND combined. I think it make sense to query a bug database and I hope it's enough to have an import filter (at least I don't see why it wouldn't). Maybe in the future explicit logical operator could be added but that would be a whole another level of complexity and I suspect not worth it at the moment.

Maybe a solution would be to be able to specify multiple query to filter the imports ? It's a bit inelegant but could work. WDYT ?

> Also, if the "field split" function is too coarse of a tool for the job, there's got to be a regex based lexer library already in go.

The lexer is just the previous version refactored out. For some reason, writing a proper lexer has managed to escape my formation and my experience until now so I'll have to learn a bit before fixing that, unless someone want to play with that. At least it's now properly extracted and tested, ready for improvements.

> This coronavirus thing is really messing with my schedule.

Tell me about it :-|

MichaelMure:
Merging as this is an improvement. We can improve further later.

MichaelMure:
More robust tokenizer in ecde909b0a87c329db3cd62562c7bf3902139320

PR #358, opened 2020-03-15
gitlab: fix issues import url

A-Hilaly:


MichaelMure:
@A-Hilaly I change the matching of bugs on import for Gitlab to use the IID and the base URL instead of the web URL that seems to be unstable

@cheshirekow I added the base URL tagging on bugs and changes the matching the same way

MichaelMure:
Let me know both of you if that was a mistake ;)

PR #371, opened 2020-04-11
Supports git config includes

mindriot101:
I like to have "private" settings in a git include file, and store the
main .gitconfig under version control. I do not want any authentication
keys or tokens (even if encrypted) in version control, so I have by main
.gitconfig include another file which is local and not tracked.

The current implementation calls `git config --global --get-regexp
<keyPrefix>` and for some reason, this command does not follow git
include files.

The changes suggested in this PR add the `--includes` flag to the
command, which then reads any included files.


MichaelMure:
> --[no-]includes
>     Respect include.* directives in config files when looking up values. Defaults to off when a specific file is given (e.g., using --file, --global, etc) and on when searching all config files.

We are using `--global` here and that switch of the support.

MichaelMure:
Thanks!

mindriot101:
Glad it is useful!

PR #372, opened 2020-04-12
termui: make the help visually easier to parse

MichaelMure:
image

cc @nh2

### DIFF
--- a/termui/bug_table.go
+++ b/termui/bug_table.go
@@ -23,6 +23,16 @@ const bugTableInstructionView = "bugTableInstructionView"
 const defaultRemote = "origin"
 const defaultQuery = "status:open"
 
+var bugTableHelp = helpBar{
+	{"q", "Quit"},
+	{"s", "Search"},
+	{"←↓↑→,hjkl", "Navigation"},
+	{"↵", "Open bug"},
+	{"n", "New bug"},
+	{"i", "Pull"},
+	{"o", "Push"},
+}
+
 type bugTable struct {
 	repo         *cache.RepoCache
 	queryStr     string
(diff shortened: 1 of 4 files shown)


termui/bug_table.go
```diff
@@ -117,9 +127,8 @@ func (bt *bugTable) layout(g *gocui.Gui) error {
 		}
 
 		v.Frame = false
-		v.BgColor = gocui.ColorBlue
 
-		_, _ = fmt.Fprintf(v, "[q] Quit [s] Search [←↓↑→,hjkl] Navigation [↵] Open bug [n] New bug [i] Pull [o] Push")
+		_, _ = fmt.Fprintf(v, bugTableHelp.Render())
```
golangcibot, line 131: SA1006: printf-style function with dynamic format string and no further arguments should use print-style function instead (from `staticcheck`)

termui/show_bug.go
```diff
@@ -92,11 +101,10 @@ func (sb *showBug) layout(g *gocui.Gui) error {
 
 		sb.childViews = append(sb.childViews, showBugInstructionView)
 		v.Frame = false
-		v.BgColor = gocui.ColorBlue
 	}
 
 	v.Clear()
-	_, _ = fmt.Fprintf(v, "[q] Save and return [←↓↑→,hjkl] Navigation [o] Toggle open/close [e] Edit [c] Comment [t] Change title")
+	_, _ = fmt.Fprintf(v, showBugHelp.Render())
```
golangcibot, line 107: SA1006: printf-style function with dynamic format string and no further arguments should use print-style function instead (from `staticcheck`)

MichaelMure:
Toying with that again, I think simply having the text in bold has the best balance between readability and aggressivity. What do you think ?

image

nh2:
@MichaelMure I don't think that helps, because now you're back in ambiguous mode when looking at this:

image

Does the `[i]` bind to the left or to the right?

Issue #130, opened 2019-05-02
feat: Global user/"gpg signed users"


EclectickMedia:
With user's becoming a thing in the codebase, I thought it would be cool to utilize git's gpg signed commit features to allow a user to "prove" their identity via public key.

Having user functionality is wonderful, but what is to stop someone from "adopting" the key of a repository maintainer? Using gpg signatures gives a git native method of verifying who said what on the timeline.

This is defeated however if there is already a system in place for handling identity verification.

MichaelMure:
Yes, that's the general idea. I haven't written much about it, but here goes the gist:
- identities in git-bug are mutable even after creation, they use a chain of commits (each holding a full timestamped version of the identity) to be editable. The current active version is the last one in the chain, but git-bug can go back in time to check if the rules in a bug are valid with the identity state **at that time**. 
- each version of the identity store either zero, one or more crypto public key
- if an identity has zero key, it is *unprotected*. That means that anybody can use it. This is useful for importers that don't necessarily have a public key at hand.
- if an identity has at least one crypto key, it is *protected*. This means a few things:
  - to create a new version for this identity, the new commit *must* be signed with a key present in the precedent version, otherwise it's rejected
  - all bug editions with this identity *must* be signed with a crypto key valid *at that time*. This means that only the author can use that identity.
  - it's not possible to revert back to unprotected
  - key rotation/revocation can be implemented: simply create a new version without the old key and a new one
- once this key system exist, we can add a permission scheme and roles (example: an admin can edit any comments but normal users can only edit their own)
- if the keys are the same for git and git-bug, we can start to check commit signature on the normal (code) commits (!)
- the identity subsystem could possibly be made into a standalone tool (ie, without git-bug) to embed and share public keys inside the git repo to check automatically commit signatures

There is a few design decision to be made:
- does that should use GPG ? How ? CLI parameters don't seems to be standardized and that can be a pain for multi-platform.
- should we embed our own GPG implementation ?
- does those public keys should be published in a public GPG keystore ?
- should we import keys from GPG ? 
- if/when the webUI start to accept users from external systems, how do we create/store keys for them ?

TODO is:
1. [ ] implement identity edition (mostly a matter of UI/UX, core is ready)
1. [ ] implement the storage of public keys in the identities + UI/UX to interact with
1. [ ] design & implement the storage of the corresponding private keys somewhere in the system
1. [ ] implement commit's signature verification for git-bug data
[...]

Issue #55, opened 2018-10-02
Identity management


MichaelMure:
Currently, each `Operation` store the author independently. While it's easier and works well for now, there is some shortcoming:

- no way to update an identity (update an email, update an avatar ...)
- no easy way to recognize that two author are the same
- data is stored multiple time (no that much of an issue with the git compression)
- crypto signature & authorization scheme is hard to implement

So I've been thinking it could be changed for the following (rough idea):

- identity (including a public key would be neat) is stored as a JSON in a git blob
- this blob hash give the `id` for this identity
- each time an identity is updated, a new blob is created with the new values and these blob are linked together in a chain of commit, the head being the last state.
- in each operations, actor(s) are simply identified with the id.

Now, I'm not sure yet how these identities should be stored and distributed. a git ref for each identity ? Each required identity linked in the `OperationPack` git tree ? A master ref aggregating all known identities ?

Hopefully this would be the last of the backward-incompatible change in the data model

MichaelMure:
Here is the CLI UX I thought about, comments ?

```markdown
# Display current identity (any identity if <id> is specified)
git bug id [<id>] [<time>]

# Display the current name (any identity if <id> is specified)
git bug id name [<id>] [<time>]

# Display the current email (any identity if <id> is specified)
git bug id email [<id>] [<time>]

# Display the current login (any identity if <id> is specified)
git bug id login [<id>] [<time>]

# Display the current avatarURL (any identity if <id> is specified)
git bug id avatarURL [<id>] [<time>]

# Set the current name
git bug id set name <name>

# Set the current email
git bug id set email <email>

# Set the current login
git bug id set login <login>

# Set the current avatarURL
git bug id set avatarURL <avatarURL>

# Display the current keys (any identity if <id> is specified)
git bug keys [<id>] [<time>]

# Manage keys (to be determined)
git bug keys import ...
git bug keys export ...
git bug keys add ...
git bug keys generate ...
git bug keys revoke ...
```

MichaelMure:
#89 has been merged with the bulk of the changes, but there is a few things left to iron properly before a release can happen.

MichaelMure:
As of now, this is pretty much done. All the core features are there, tested and working.

What is left is the few command line tool to get/set particular identity fields:

```
# Display the current name (any identity if <id> is specified)
git bug user name [--time <time>] [<id>]

# Display the current email (any identity if <id> is specified)
git bug user email [--time <time>] [<id>]

[...]

PR #376, opened 2020-04-15
git: Allow signing of commits

aleb:
We could add more ways of specifying a key, such as picking one from the
adopted identity, but for now with this change we allow users who know what
they are doing to sign their bug changes.

By simply passing "-S" to `git commit-tree` when "commit.gpgsign" is "true", we
use git's mechanism for key choosing which is either by the configured user
name and email OR by whatever is specified in "user.signingkey".


repository/git_testing.go
```diff
@@ -42,6 +53,93 @@ func CreateTestRepo(bare bool) *GitRepo {
...
+	armoredPub := pubBuilder.String()
+
+	// Create a custom gpg keyring to be used when creating commits with `git`.
+	keyring, err := ioutil.TempFile("", "keyring")
+	check(err)
+
+	// Import the armored private key to the custom keyring.
+	priv, err := ioutil.TempFile("", "privkey")
+	check(err)
+	_, err = fmt.Fprintf(priv, armoredPriv)
```
golangcibot, line 98: ineffectual assignment to `err` (from `ineffassign`)

repository/git_testing.go
```diff
@@ -42,6 +53,93 @@ func CreateTestRepo(bare bool) *GitRepo {
...
+	err = priv.Close()
+	check(err)
+	err = exec.Command("gpg", "--no-default-keyring", "--keyring", keyring.Name(), "--import", priv.Name()).Run()
+	check(err)
+
+	// Import the armored public key to the custom keyring.
+	pub, err := ioutil.TempFile("", "pubkey")
+	check(err)
+	_, err = fmt.Fprintf(pub, armoredPub)
```
golangcibot, line 107: ineffectual assignment to `err` (from `ineffassign`)

repository/git_test.go
```diff
@@ -65,3 +66,33 @@ func TestConfig(t *testing.T) {
...
+// checkStoreCommit creates a commit and checks if it has been signed.
+// See https://git-scm.com/docs/git-log#Documentation/git-log.txt-emGem
+// for possible signature status values.
+func (repo *GitRepo) checkStoreCommit(t *testing.T, expectedSignedStatus string) {
+	content := fmt.Sprintf("file content %d", repo.CreateTime())
+	blobHash, err := repo.StoreData([]byte(content))
+	assert.NoError(t, err)
+
+	var entries = []TreeEntry{{Blob, blobHash, "filename"}}
+	treeHash, err := repo.StoreTree(entries)
```
golangcibot, line 79: ineffectual assignment to `err` (from `ineffassign`)

aleb:
Do you want to review this by itself, or in the context of the entire keys add/rm/validation functionality?

repository/git_test.go
```diff
@@ -65,3 +66,34 @@ func TestConfig(t *testing.T) {
 	err = repo.LocalConfig().RemoveAll("section.key")
 	assert.Error(t, err)
 }
+
+// checkStoreCommit creates a commit and checks if it has been signed.
+// See https://git-scm.com/docs/git-log#Documentation/git-log.txt-emGem
+// for possible signature status values.
+func (repo *GitRepo) checkStoreCommit(t *testing.T, expectedSignedStatus string) {
```
MichaelMure, line 73: This helper should not be part of the normal `GitRepo` as this is only for testing. What about `func checkStoreCommit(t *testing.T, repo *GitRepo, expectedSignedStatus string) {` ?

repository/git_testing.go
```diff
@@ -42,6 +53,95 @@ func CreateTestRepo(bare bool) *GitRepo {
...
+	check(err)
+
+	if err := config.StoreString("user.signingkey", entity.PrivateKey.KeyIdString()); err != nil {
+		log.Fatal("failed to set user.signingkey for test repository: ", err)
+	}
+
+	// Armor the private part.
+	privBuilder := &strings.Builder{}
+	w, err := armor.Encode(privBuilder, openpgp.PrivateKeyType, nil)
+	check(err)
```
MichaelMure, line 74: Please use `assert.NoError(t, err)` or `require.NoError(t, err)` if you want to stop the test immediately.

repository/git_testing.go
```diff
@@ -42,6 +53,95 @@ func CreateTestRepo(bare bool) *GitRepo {
 	return repo
 }
 
+// setupSigningKey creates a GPG key and sets up the local config so it's used.
+// The key id is set as "user.signingkey". For the key to be found, a `gpg`
+// wrapper which uses only a custom keyring is created and set as "gpg.program".
+// Finally "commit.gpgsign" is set to true so the signing takes place.
+func (repo *GitRepo) setupSigningKey() {
```
MichaelMure, line 60: Same thing, this should not be part of the normal `GitRepo`, if only for the semantic.

repository/git_testing.go
```diff
@@ -42,6 +53,95 @@ func CreateTestRepo(bare bool) *GitRepo {
...
+		log.Fatal("failed to set gpg.program for test repository: ", err)
+	}
+
+	if err := config.StoreString("commit.gpgsign", "true"); err != nil {
+		log.Fatal("failed to set commit.gpgsign for test repository: ", err)
+	}
+}
+
+// createGPGWrapper creates a shell script running gpg with a specific keyring.
+func createGPGWrapper(keyringPath string) string {
```
MichaelMure, line 127: Can you explain what's the point of this ?

aleb: When git signs the commit it runs by default `gpg`, or whatever is in the "gpg.program" config entry. We use this `gpg` wrapper so `gpg` can find the temporary keys we create so we can keep them separate from the user's keyring.

MichaelMure: Is there no way to pass the keys when doing the commit or something ? Deploying a payload like this is not exactly awesome, and doesn't look portable either :-|

MichaelMure: I think migrating over gogit would help a lot here, but that's some work ...

aleb: This is not about git, it's about the tool used by git to sign the commit. We need to create a key pair, and we can specify that key just fine by setting `user.signingkey` which is passed as an extra argument to `gpg --status-fd=2 -bsau <user.signingkey>` from what I've noticed.
But the default `gpg` tool will not be able to find it.

Indeed the shell script is not portable, but should be trivial to add the Windows alternative when it's needed.

When migrating to gogit this could be simplified if gogit allows to do it easier?

MichaelMure: I didn't dig so much in go-git, but it seems to me that it is a relatively complete implementation of git in go, which include support for signatures. It's a bit tricky to find but there is for instance https://github.com/go-git/go-git/blob/master/plumbing/object/commit.go#L348 and https://pkg.go.dev/github.com/go-git/go-git/v5?tab=doc#CommitOptions

It would also have other great benefits: no more calling a external binary, no need to support every version of git anymore, most likely faster, no more constrained by the sometimes awkward git CLI

aleb: We could start using go-git only for specific operations, such as the commit-tree one, but I don't want to look into this.

commands/key_add.go
```diff
@@ -0,0 +1,89 @@
...
+	if err != nil {
+		return err
+	}
+
+	if len(args) > 0 {
+		return fmt.Errorf("unexpected arguments: %s", args)
+	}
+
+	if keyAddArmoredFile != "" && keyAddArmored == "" {
+		keyAddArmored, err = input.BugCommentFileInput(keyAddArmoredFile)
```
MichaelMure, line 38: I suppose that works, but please make a properly named version of this function, if only for the semantic. Maybe a really generic one like `input.FileInput(fileName string)` ? I suppose you don't need the post-process step as well.

input/input.go
```diff
@@ -112,6 +113,25 @@ func BugCommentEditorInput(repo repository.RepoCommon, preMessage string) (strin
...
+// the key.
+func IdentityVersionKeyEditorInput(repo repository.RepoCommon, preMessage string) (string, error) {
+	template := fmt.Sprintf(identityVersionKeyTemplate, preMessage)
+	raw, err := launchEditorWithTemplate(repo, keyFilename, template)
+
+	if err != nil {
+		return "", err
+	}
+
+	return processComment(raw)
```
MichaelMure, line 132: you should probably have a dedicated post-process function, tailored for your inputs.

identity/identity.go
```diff
@@ -147,6 +149,11 @@ func read(repo repository.Repo, ref string) (*Identity, error) {
 
 		// tag the version with the commit hash
 		version.commitHash = hash
+
+		if err := version.Validate(); err != nil {
```
MichaelMure, line 153: hmm, this is done already in `identity_actions.go:MergeAll()` and `Identity.Commit`. It should be enough.

identity/identity.go
```diff
@@ -544,6 +551,46 @@ func (i *Identity) ValidKeysAtTime(time lamport.Time) []*Key {
 	return result
 }
 
+func (i *Identity) AddKey(key *Key) {
```
MichaelMure, line 554: There is already the `Identity.Mutate()` mechanism to modify an identity. It's a little weird but it allow to update multiple element of an identity in one go (ie, one commit) and update the cache easily. Those key changes should use this mechanism.

identity/identity.go
```diff
@@ -544,6 +551,46 @@ func (i *Identity) ValidKeysAtTime(time lamport.Time) []*Key {
...
+	if last.commitHash != "" {
+		last = last.Clone()
+		i.versions = append(i.versions, last)
+	}
+
+	last.keys = append(last.keys, key)
+}
+
+func (i *Identity) RmKey(fingerprintHex string) error {
+	fingerprintBytes, err := hex.DecodeString(fingerprintHex)
```
MichaelMure, line 565: put those decoding stuff into reusable helpers in `key.go` ?

identity/identity.go
```diff
@@ -544,6 +551,46 @@ func (i *Identity) ValidKeysAtTime(time lamport.Time) []*Key {
...
+	var fingerprint [20]byte
+	if len(fingerprintBytes) != 20 {
+		return fmt.Errorf("expected")
+	}
+	copy(fingerprint[:], fingerprintBytes)
+
+	last := i.lastVersion()
+	for j, key := range last.keys {
+		if key.PublicKey.Fingerprint == fingerprint {
+			if last.commitHash != "" {
```
MichaelMure, line 579: if you use the `Mutate` mechanism, you don't need all of this.

identity/key.go
```diff
@@ -1,18 +1,73 @@
...
+)
+
 type Key struct {
-	// The GPG fingerprint of the key
-	Fingerprint string `json:"fingerprint"`
-	PubKey      string `json:"pub_key"`
+	// PubKey is the armored PGP public key.
+	ArmoredPublicKey string `json:"pub_key"`
+	// ExpireUnixTime is when the key becomes invalid.
+	ExpireUnixTime int64 `json:"expire_unix_time"`
```
MichaelMure, line 15: It might makes sense to have an expire time but `Identity.ValidKeyAtTime` should then be updated to filter the expired one. Using a timestamp also make sense but it open a whole can of problems : what happen if clocks are incorrect ? There is a lot of failure mode here. Do you **really** need that concept ? What for ?

commands/key.go
```diff
@@ -0,0 +1,50 @@
...
+
+var keyCmd = &cobra.Command{
+	Use:     "key [<user-id>]",
+	Short:   "Display, add or remove keys to/from a user.",
+	PreRunE: loadRepoEnsureUser,
+	RunE:    runKey,
+}
+
+func init() {
+	RootCmd.AddCommand(keyCmd)
```
MichaelMure, line 49: Let's place those commands under `user` instead of directly at the root.

commands/key_add.go
```diff
@@ -0,0 +1,89 @@
...
+		if err != nil {
+			return err
+		}
+	}
+
+	if keyAddArmoredFile == "" && keyAddArmored == "" {
+		keyAddArmored, err = input.IdentityVersionKeyEditorInput(backend, "")
```
MichaelMure, line 45: what about generating a new key instead ? Ideally using git-bug, even with signature, should not require for the user to learn PGP or even knowing what that is. Power user should still be allowed to input their own keys but that's their choice to go the hard way.

commands/key_add.go
```diff
@@ -0,0 +1,89 @@
...
+
+func runKeyAdd(cmd *cobra.Command, args []string) error {
+	backend, err := cache.NewRepoCache(repo)
+	if err != nil {
+		return err
+	}
+	defer backend.Close()
+	interrupt.RegisterCleaner(backend.Close)
+
+	id, args, err := _select.ResolveUser(backend, args)
```
MichaelMure, line 28: you should only be able to add a key on a user you are allowed to control. There is two scenario where this is the case:
- the identity has no key yet ("unprotected")
- you have the private key of one of the public key present in the last version

This rules should be implemented in the `Validate()` function of Identity so they are checked 1) before a commit and 2) when merging remote data.

MichaelMure:
Not sure if you came across https://github.com/MichaelMure/git-bug/issues/130 and https://github.com/MichaelMure/git-bug/issues/55, have a look :)

commands/key_add.go
aleb: Generating a new key means the private key has to be stored in a file, and to be used with `gpg` it has to be imported. Can be done, but we can live without it.

commands/key_add.go
aleb: If the user loses the private key, the identity can't be used anylonger. 

(In our case, we trust everybody to add/remove keys to anybody. The server ("when merging remote data"?) would run a custom check to make sure the keyring is valid and that everything originates from one key.)

aleb: Is there a notion of "admin identity"?

identity/key.go
aleb: Good question. I'm not sure what's the usefulness of the expiring time.

The `ValidKeysAtTime(time lamport.Time)` method gets only a `lamport.Time` value, so it can only be used for validating the identities. 

For validating a commit we check differently. Let's say a key is removed. We have a commit signed with the said removed key. How do we check if the key was valid at the time the commit was made (committer time)? We have the commit unix time, and the commit unix times of the two operations introducing and removing the key.

The point is that the expiring time would be used in the commit validation mechanism, not in the identity validation one, so we can add it there in case somebody needs it. 

Makes sense?

MichaelMure: Are you talking about commits *made by git-bug*, ie operations, or actual code commits ?

commands/key_add.go
MichaelMure: Storing the private key in the git config, even is it's not perfect seems like an OK solution to me. It would be nice to be entirely independent of a `gpg` binary, but I don't know if that's possible without migrating to go-git.

commands/key_add.go
MichaelMure: > If the user loses the private key, the identity can't be used any longer.

Yes, that's the kind of hard rules you end up having with distributed applications. Some sort of backup or trust system could be imagined but for now that looks like the good way to go to me.

> The server ("when merging remote data"?) would run a custom check to make sure the keyring is valid and that everything originates from one key.)

Why do you need code server side ? I understand you have different requirements, but is there a way for you to implement those extra requirements while staying close to git-bug's design as a distributed app ? It would help both of us if we can manage that.

> Is there a notion of "admin identity"?

Not yet but that would be nice. To reach that point the plan is:
- have a shared project config data structure (valid labels , .... )
- store in that new structure who is admin, with the mechanism to update it
- implement this admin role as rules to check when accepting new data (bugs + project config at least)

identity/key.go
MichaelMure: If you are talking of validating operations, you don't need any real time. Each operation has a Lamport time that you can use to query which keys are valid at that logical time. As the identity is a chain of version, we keep around all the keys even when removed.

Now, for validating code commits ... it's another story. Without adding an extra expiry time to a key, we can use the existing unixtime of the version the keys are attached to to derive a validity period. This way we stay close to the normal git-bug way and not open a window for a lot of new failure scenario. It would "just" require a new `Identity.ValidKeysAtUnixTime()` function.

As we can't trust time, we are subject to false positive / false negative. It might be fine but rules need to be decided for what happen in those cases. One possible way to make this process a bit more robust could be to check that the unix time of both identity's version and code commit are monotonic (ie, stated time and data location match in ordering). This would at least bring us closer to the full logical ordering of the Lamport time and it's nice properties. Problem is, it's not a complete guarantee and if we detect such a problem, the best we can do is report that to the user as "something's weird I don't know".

identity/identity.go
aleb: Done. It's confusing because the "Mutator" name of the class suggests it should be responsible for doing the job, while that is done by a separate function.

identity/key.go
aleb: Ok, let's forget about expiration times, and use only the unix time of the commit which removed the key.

commands/key_add.go
aleb: > > If the user loses the private key, the identity can't be used any longer.
> 
> Yes, that's the kind of hard rules you end up having with distributed applications. Some sort of backup or trust system could be imagined but for now that looks like the good way to go to me.

What would be the problem with trusting everybody to make changes to identities?

> > The server ("when merging remote data"?) would run a custom check to make sure the keyring is valid and that everything originates from one key.)
> 
> Why do you need code server side ? I understand you have different requirements, but is there a way for you to implement those extra requirements while staying close to git-bug's design as a distributed app ? It would help both of us if we can manage that.

The problem is the users can push anything working around the local validation done when they use git-ticket. The pre-receive server-side git hook prevents merging bad changes.

commands/key_add.go
aleb: Can be done, but can also be added later?

identity/identity.go
MichaelMure: Yeah that's a good point. A little bit more doc would not hurt here.

commands/key_add.go
MichaelMure: Well, I'm not going to refuse this useful contribution because it doesn't meet 100% of what I think should happen but I'm considering that a requirement for an actual release.

I'm also saying that because it *looks like* that you are going on a more complicated path but I may be wrong.

commands/key_add.go
MichaelMure: > What would be the problem with trusting everybody to make changes to identities?

In the general case people that have a write access to a git repo are trustworthy but that's not always the case. Git support signature to be able to prove that someone made something and improve the trust you can have in the system, but if someone is able to change the valid keys of anyone else, this signature system is worth nothing.

In a similar fashion, this identity system will be the base of the role management. Some users will be admin of a project (or more complex roles) and some not. If you can tamper the keys of anyone, this system is equally useless.

If everyone is trustworthy it's all good and you don't even have to bother with keys and you can leave your identity unprotected. When it's not the case, you should be protected from abuse.

> The problem is the users can push anything working around the local validation done when they use git-ticket. The pre-receive server-side git hook prevents merging bad changes.

Why is that the case ? If your user use `git ticket pull` (which is much simpler than manually merging remote hidden branches), data will be checked. Everything that doesn't follow the rules will be discarded.

My point is, even if a **malicious** user manage to push bad changes, those changes will not propagate. A non-malicious user should not be able to create bad data as it is checked before writing to git.

Now I agree that its not perfect and a server side hook to control the data is useful. Would you share the one you have ?

MichaelMure: Huhu sorry, reading myself again, I realize I rewrote what you were saying.

commands/key_add.go
aleb: We do need `gpg` and the private key to be in the gpg keyring to be able to create signed commits with `git`.

MichaelMure: I see. I'll be happy when I can ditch the git CLI exec repo, it cause so many problems :(

commands/key_add.go
aleb: The signatures are useful to prove the changes have not been made by an authorized user pretending to be some other authorized user.

The valid keys can be changed by any authorized user, but it will be obvious who changed them, since the server shall not accept forced pushes. 

The malicious authorized users can work around the local validation, but they shall not be allowed to push, because we can. :) Yes, I'll share the validator!

commands/key_add.go
aleb: I mean the private key will be used also to create the normal commits in the repository, so `gpg` will be used since creating those commits will be done with `git`, not with anything git-bug or go-git.

commands/key_add.go
MichaelMure: > The valid keys can be changed by any authorized user, but it will be obvious who changed them, since the server shall not accept forced pushes.

Yes, you can do forensics to find out who dropped the ball, but it would be better to actually prevent it, no ?

MichaelMure: Thinking out loud, there should probably be a `Keystore` interface somewhere with multiple implementations:
- git config
- gpg
- some centralized system that allow recovery for "corporate" environment

Maybe even an implementation that allow to have multiple sub-keystore for example git config + gpg.

westurner:
cc'ing from #383 "DOC,UBY: What do I specify for avatarURL? Gravatar? W3C DID?"

 https://www.w3.org/TR/did-core/ ::

>  Decentralized identifiers (DIDs) are a new type of identifier that enables verifiable, decentralized digital identity. A DID identifies any subject (e.g., a person, organization, thing, data model, abstract entity, etc.) that the controller of the DID decides that it identifies. These new identifiers are designed to enable the controller of a DID to prove control over it and to be implemented independently of any centralized registry, identity provider, or certificate authority. DIDs are URLs that associate a DID subject with a DID document allowing trustable interactions associated with that subject. Each DID document can express cryptographic material, verification methods, or service endpoints, which provide a set of mechanisms enabling a DID controller to prove control of the DID. Service endpoints enable trusted interactions associated with the DID subject. A DID document might contain semantics about the subject that it identifies. A DID document might contain the DID subject itself (e.g. a data model).
>
> This document specifies a common data model, a URL format, and a set of operations for DIDs, DID documents, and DID methods. 

https://github.com/topics/decentralized-identifiers?l=go

https://github.com/decentralized-identity/github-did#what-can-i-do-with-my-did

> Use your DIDs to test Linked Data Signatures, such OpenPgpSignature2019 which is currently being developed. When DID Documents are signed, they include a proof attribute, which is used to provide proof that someone controlled the private key associated with the public key listed in the did document at the created datetime.

westurner:
... GPG keyservers could also solve for PKI in a centralized way.

GitHub and GitLab already expose SSH and GPG public keys if they're added to a username:

- API: https://api.github.com/users/defunkt
  - `avatar_url`, `gravatar_id`
- SSH: https://github.com/defunkt.keys
- GPG: https://github.com/defunkt.gpg
- GPG: https://api.github.com/users/defunkt/gpg_keys

GitLab 
- API: https://docs.gitlab.com/ee/api/users.html
- SSH: https://gitlab.com/username.keys
- GPG keys url: https://gitlab.com/gitlab-org/gitlab/-/issues/21584

Keybase

- API: https://keybase.io/docs/api/1.0/call/user/pgp_keys.asc
- GPG: https://keybase.io/chris/pgp_keys.asc 
- SSH: https://github.com/keybase/keybase-issues/issues/710

aleb:
@MichaelMure 
I added a 3rd commit "Verify signed identities".

I'm not sure what verification logic would be common to git-bug and git-ticket (hah, good thing the name is different). I pushed the verification logic we'll use so you can have a look. It's not yet merged in git-ticket. If there is something common and reusable, I could separate the logic from this 3rd commit.

I introduced go-git to be able to read the commits to be verified. The go-git repo is added as `repo2` along `repo repository.ClockedRepo` in `RepoCache`, could not come up with anything better.

The very first commit is very important. It must introduce a key and be signed with that key.  I changed `git-bug user create` to be able to specify a key and have it included in this first commit. When you validate the identities, the fingerprint of the first key is reported like this:

```
$ b validate
first commit signed with key: C77E1D7542889EC0E45BA88899DA3BE167DA2410
```

When checking a git reference in the main repo, the key introduced by the first commit guarantees that everything is coming from us.

I made a small change to `Identity.Commit` to increase the lamport time so each identity version has a different lamport time so they can be sorted. The logic requires the lamport time for each version introducing/removing keys to be distinct. This might bite us, but we'll see. 

The validate logic then sorts the identity versions by lamport time and iterates them. A keyring is built gradually and used to verify each commit before applying the key changes to the keyring.

We use an `openpgp.EntityList` which is just a list of `openpgp.Entity` which has the keys and also the expire time. Technically, we could use the same Entity and add keys to it and revoke them .. but I found it simpler to create a separate `Entity` each responsible for a single key.

The time of the commit adding the key is the Identity.SelfSignature. CreationTime. The time of the commit removing a key is used to calculate the lifetime and stored in Identity.SelfSignature.KeyLifetimeSecs, this being the expiry time. These are used in `verifyCommitSignature` which is used when checking both identities and normal commits (from the main repo). This means for identities both the lamport time and the unix time is checked.

I had to expose private fields such as `Identity.Versions`. Maybe it can be done more cleanly.

TODO: I should move at least `identity/key_test.go` to the previous commit, maybe some more small stuff.

We could have used key fingerprints instead of key ids to identify the keys, but it seemed simpler to use the key id. The key fingerprint is composed of the key id and some other bits.

TL;DR: The first two commits would make sense to merge into git-bug, maybe some common part of validate functionality.

Please have a look.

Any requests?

aleb:
@MichaelMure In light of our validation logic which will never make it into git-bug, could you please comment on the first two commits if they are roughly ok and could be merged in git-bug? Thanks.

MichaelMure:
Hey @aleb this is a pretty big change to review, it might take time for me to go through it.

As a preamble, I find that some of this code doesn't follow the architecture design ans "leak" through the layers. Two things comes to mind:
- `Key`: all the tricky handling of keys, encoding, parsing should be abstracted behind the public interface of `Key` instead of spreading up to the commands. I made a tentative refactoring with https://github.com/MichaelMure/git-bug/commit/1e760c3b559c7d00eec81c0060b8a94b0156dbe1 . It's not done, not everything compile, but it should at least give you the idea.
- I'm yet to understand what this validation is, but the go-git thing should be in the repo package, Ideally behind the same `Repo` interface and in the same `GitRepo` implementation, even if that make a small Frankenstein here. The cache layer is really not meant for that. That would also be a first step towards migrating.

Speaking of this Validator thing, I'm a bit confused tbh, what is it for ? When is it supposed to run ? What should happen if the validation fail ? Why not stay with the `Validate()` functions everywhere and have a piece of code reload everything and validate through those ? Maybe Identity/Bug should also check the commit signature when loading the data ?

> I'm not sure what verification logic would be common to git-bug and git-ticket

Unless I'm mistaken, both share a very similar need here, or at least we can find a common ground from where you can build your extra validations without breaking everything at each rebase. It seems to me that you want a different set of data/operations (Bug vs ticket) but the "framework" part could be largely shared.

> The validation logic requires the newly created identity to have a key and be signed with that key.

git-bug will keep allowing identities without keys, but a flag somewhere to have a mode where they **need** to have at least one is entirely possible. Useful for both.

> The very first commit in the identities history is very important. It is of course a commit which creates a new identity. The validation logic requires the newly created identity to have a key and be signed with that key.

Hmmm, when I originally thought about this the idea was, to add a key to an identity, add the private key in the version as well as a signature for the previous **data**. Here you are adding the signature in the commit itself, which is a bit different and tie git-bug more to git itself. Not sure what is the best way really. I need to think more about this.

> I made a small change to Identity.Commit to increase the lamport time so each identity version has a different lamport time so they can be sorted. 

Adding ordering to identities is fine, but that needs to be a distinct Lamport clock to not mess with the bugs. It looks like it's time to update the `ClockedRepo` to something simpler that support an arbitrary number of clocks. What about this:
```go
// ClockedRepo is a Repo that also has Lamport clocks
type ClockedRepo interface {
	Repo

	GetClock(name string) Clock
}

type Clock interface {
	Time() lamport.Time
	Increment() (lamport.Time, error)
	Witness(time lamport.Time) error
}
```

MichaelMure:
> Hmmm, when I originally thought about this the idea was, to add a key to an identity, add the private key in the version as well as a signature for the previous data. Here you are adding the signature in the commit itself, which is a bit different and tie git-bug more to git itself. Not sure what is the best way really. I need to think more about this.

So I think this boils down to the question of whether or not rebase are needed. If the commit is signed by the author, rebasing and rearranging the commits is not possible anymore without invalidating the signature. If the data is signed, we are still free to do that.

In other words, if the commit is signed, the whole history is signed. If the data is signed, only this specific piece of data is signed, which only prove authorship (which is really nice already).

The problem is that at the moment, git-bug **does** make rebase when changes are pulled. Most of the time people use a unique central git repo so it could work because the commit that are rebased are your own, so you could sign them again. But as soon as you use more than one remote or push/pull directly this model fall apart.

What I was thinking about in https://github.com/MichaelMure/git-bug/issues/212 was to replace the rebase with merges to do the conflict resolution. This would make the data model entirely immutable and would combine very well with signing the commits.

So, short of doing this #212 (which would be awesome but a lot of work), the choice end up being either:
1. explicitly saying that fully decentralized workflow is not supported
1. working on the commit signing without fully enabling it yet and wait for the data model rework to complete
1. do data signing and change to commit signing when the data model change is completed

I think option 2 make the most sense here. What do you think ?

aleb:
Sorry for the late reply.

> As a preamble, I find that some of this code doesn't follow the architecture design ans "leak" through the layers. Two things comes to mind:
> 
> * `Key`: all the tricky handling of keys, encoding, parsing should be abstracted behind the public interface of `Key` instead of spreading up to the commands. I made a tentative refactoring with [1e760c3](https://github.com/MichaelMure/git-bug/commit/1e760c3b559c7d00eec81c0060b8a94b0156dbe1) . It's not done, not everything compile, but it should at least give you the idea.

I cherry-picked your commit and fix the issues.

> * I'm yet to understand what this validation is, but the go-git thing should be in the repo package, Ideally behind the same `Repo` interface and in the same `GitRepo` implementation, even if that make a small Frankenstein here. The cache layer is really not meant for that. That would also be a first step towards migrating.

Sounds good. Done.


> Speaking of this Validator thing, I'm a bit confused tbh, what is it for ? When is it supposed to run ? What should happen if the validation fail ? Why not stay with the `Validate()` functions everywhere and have a piece of code reload everything and validate through those ? Maybe Identity/Bug should also check the commit signature when loading the data ?

An identity can't fully validate itself if we want to allow others to make changes to an identity. A separate validator is needed to keep track of what keys have been added and removed and when. (It would have been simpler if only the keys added by the identity can sign the identity's changes, like you wanted(?), but we wanted something more flexible.)

The `validate` command verifies that the keys are in good order and that the identity changes are properly signed with the keys available at that lamport time. It will be used for example on the server-side git hook to prevent pushing bad data to the repo. 

> > I'm not sure what verification logic would be common to git-bug and git-ticket
> 
> Unless I'm mistaken, both share a very similar need here, or at least we can find a common ground from where you can build your extra validations without breaking everything at each rebase. It seems to me that you want a different set of data/operations (Bug vs ticket) but the "framework" part could be largely shared.

The more that can be shared, the better. 


> git-bug will keep allowing identities without keys, but a flag somewhere to have a mode where they **need** to have at least one is entirely possible. Useful for both.

What would be a good place to store this flag in the repo?


> > The very first commit in the identities history is very important. It is of course a commit which creates a new identity. The validation logic requires the newly created identity to have a key and be signed with that key.
> 
> Hmmm, when I originally thought about this the idea was, to add a key to an identity, add the private key in the version as well as a signature for the previous **data**. Here you are adding the signature in the commit itself, which is a bit different and tie git-bug more to git itself. Not sure what is the best way really. I need to think more about this.

I don't completely follow. Why add the private key in the version? 

I'm thinking the commits need to be signed anyway, so then we use those signatures. The whole point for us is to have every commit signed (with the "valid keys").


> > I made a small change to Identity.Commit to increase the lamport time so each identity version has a different lamport time so they can be sorted.
> 
> Adding ordering to identities is fine, but that needs to be a distinct Lamport clock to not mess with the bugs. It looks like it's time to update the `ClockedRepo` to something simpler that support an arbitrary number of clocks. What about this:
> 
> ```go
> // ClockedRepo is a Repo that also has Lamport clocks
> type ClockedRepo interface {
> 	Repo
> 
> 	GetClock(name string) Clock
> }
> 
> type Clock interface {
> 	Time() lamport.Time
> 	Increment() (lamport.Time, error)
> 	Witness(time lamport.Time) error
> }
> ```
> 
> And yes, it seems that LoadClocks and WriteClocks are never used outside of GitRepo itself so they should not be part of the interface. Cleanup time !

Ok, let me see..

aleb:
> > Hmmm, when I originally thought about this the idea was, to add a key to an identity, add the private key in the version as well as a signature for the previous data. Here you are adding the signature in the commit itself, which is a bit different and tie git-bug more to git itself. Not sure what is the best way really. I need to think more about this.
> 
> So I think this boils down to the question of whether or not rebase are needed. If the commit is signed by the author, rebasing and rearranging the commits is not possible anymore without invalidating the signature. If the data is signed, we are still free to do that.
> 
> In other words, if the commit is signed, the whole history is signed. If the data is signed, only this specific piece of data is signed, which only prove authorship (which is really nice already).

If the data/operation is signed instead of the commit introducing the data, we'd have to sign a combination of data plus the hash of the parent commit. Sounds like a git system in a git system. It would be good to use git's system if possible, to keep things simple.


> The problem is that at the moment, git-bug **does** make rebase when changes are pulled. Most of the time people use a unique central git repo so it could work because the commit that are rebased are your own, so you could sign them again. But as soon as you use more than one remote or push/pull directly this model fall apart.

We'll use a single remote, so it would work for us. Good to know about the local rebasing when changes are pulled.


> What I was thinking about in #212 was to replace the rebase with merges to do the conflict resolution. This would make the data model entirely immutable and would combine very well with signing the commits.
> 
> So, short of doing this #212 (which would be awesome but a lot of work), the choice end up being either:
> 
> 1. explicitly saying that fully decentralized workflow is not supported
> 2. working on the commit signing without fully enabling it yet and wait for the data model rework to complete
> 3. do data signing and change to commit signing when the data model change is completed
> 
> I think option 2 make the most sense here. What do you think ?

We are not interested in the fully decentralized workflow. I'm not sure about the data signing benefits, but it sounds complicated. Option 2 sounds great since we'll be able to use it already and when the data model is reworked we'll be compatible.

MichaelMure:
> An identity can't fully validate itself if we want to allow others to make changes to an identity. A separate validator is needed to keep track of what keys have been added and removed and when. (It would have been simpler if only the keys added by the identity can sign the identity's changes, like you wanted(?), but we wanted something more flexible.)
> 
> The `validate` command verifies that the keys are in good order and that the identity changes are properly signed with the keys available at that lamport time. It will be used for example on the server-side git hook to prevent pushing bad data to the repo.

I still don't really get what you want to achieve, I going to guess you want to have some sort of centralized authority for identities. If so, why not have this centralized create the identities with 2 keys, one private to this authority, one given to the user. If you want to prevent the user to make changes to the identity, we could have **two** list of keys in an identity:
- keys that allow full control
- keys that only allow signing commits, bugs ... but not change the identity itself.

Or maybe we should go all in already and encode that with *capabilities* on each keys. That would be more extensible for the future.

That's a little change that could enable nice new usecases.

On a side note, maybe those keys should have an optional name ? That might help for complex schemes.

> The more that can be shared, the better.

100%. It seems that we are aiming for a sort of framework to build decentralized app on top of git, where git-bug itself would just be an example. That's awesome.

> > git-bug will keep allowing identities without keys, but a flag somewhere to have a mode where they **need** to have at least one is entirely possible. Useful for both.
> 
> What would be a good place to store this flag in the repo?

For now, a hard-coded flag in the code would be good so we don't conflict each other. Later this could also be included in the project settings for optional enabling in git-bug.

> I don't completely follow. Why add the private key in the version?

Not the private keys. I was thinking about adding the **signature** of the previous version with one of the valid keys but that's a flawed model. Your idea is better.

So to recap, the current idea is:
- have a chain of commit, each of them hold a version
- a version hold zero, one or many public keys
- if **the first** version has at least one public key, the associated commit **must** be signed with one of those keys
- after that, if the previous version has at least one public key, the commit of this new version **must** be signed with a key of the previous version. This new version doesn't have to keep the public key associated to the signature. This allow for key rotation an possibly removing entirely the keys (not sure that one would be a good thing, we can add a rule to disallow that).

Sounds good to you?

MichaelMure:
> We are not interested in the fully decentralized workflow. I'm not sure about the data signing benefits, but it sounds complicated. Option 2 sounds great since we'll be able to use it already and when the data model is reworked we'll be compatible.

Option 2 it is!

MichaelMure:
FYI, I'm working on the architecture/framework changes that will make this and adding new entities (PR, config ...) easier. It also happen to make the creation of forked tool like git-ticket easier.  Have a look at https://github.com/MichaelMure/git-bug/pull/411

Following closely will be a WIP PR for a go-git repository implementation, which would make signing way easier.

MichaelMure:
go-git repo PR: https://github.com/MichaelMure/git-bug/pull/412

MichaelMure:
@aleb FYI, something similar to your last changes already now exist in master: https://github.com/MichaelMure/git-bug/blob/master/repository/repo.go#L108

aleb:
Yes, I noticed too late because I started a while ago..

I rebased and I think I don't have other changes planned at the moment. 

Anything we should be aware of if merging this in git-ticket, so we remain compatible with upstream?

aleb:
.. other than the failing test. :D

MichaelMure:
> Anything we should be aware of if merging this in git-ticket, so we remain compatible with upstream?

Yes.

I've been increasingly convinced that the good way to go to ultimately implement all those features (identities, signatures, ACL ...) is to:
- ditch the git CLI based Repo in favor of the go-git implementation (almost done in https://github.com/MichaelMure/git-bug/pull/412): not only the CLI version is frustrating, but that will give an easier access to commit signatures and friends
- augment the repo with https://github.com/99designs/keyring to have access to the system credential storage, with a fallback on an encrypted file with password
- stop storing bridges credential in the global git config
- store identities private keys there instead of using GPG: if I'm not mistaken that means that GPG can be avoided entirely

In addition, more refactoring is need in the Identity, to store one time reference *per entity* (see https://github.com/MichaelMure/git-bug/blob/master/identity/version.go#L24-L27). This is so that it can be possible to say what version of an identity is valid for what element of an entity (Bug's operation, Ticket's operation, Config, PullRequest ...).

With all that we should have something that allow git-bug to grow with more entities, let git-ticket implement its own entities with minimal conflict and have an easier time implementing signatures for everyone.

Btw, thanks for spearheading this effort, it helped a lot to define what was necessary and what needed to change.

MichaelMure:
What do you think of this plan btw?

aleb:
> I've been increasingly convinced that the good way to go to ultimately implement all those features (identities, signatures, ACL ...) is to:
> 
> * ditch the git CLI based Repo in favor of the go-git implementation (almost done in #412): not only the CLI version is frustrating, but that will give an easier access to commit signatures and friends

Great!


> * augment the repo with https://github.com/99designs/keyring to have access to the system credential storage, with a fallback on an encrypted file with password
> * store identities private keys there instead of using GPG: if I'm not mistaken that means that GPG can be avoided entirely

IIUC the system credential storage would be used to get the private keys for signing commits created by git-bug. 

When creating normal commits in the repo, the `git` CLI will use GPG. Ideally the private keys from GPG can be used with 99designs/keyring. 


> * stop storing bridges credential in the global git config

Sounds good.


> In addition, more refactoring is need in the Identity, to store one time reference _per entity_ (see https://github.com/MichaelMure/git-bug/blob/master/identity/version.go#L24-L27). This is so that it can be possible to say what version of an identity is valid for what element of an entity (Bug's operation, Ticket's operation, Config, PullRequest ...).

In this case sorting all the Identity Versions with `sort.Sort(ByLamportTime(v.versions))` won't be possible anymore, but we'd still be able to establish an order since we know exactly which Identity Version created the operation. If an Identity is disabled/canceled, then the central authority would have to prevent accepting new commits signed by it.

I think we could work with this.

MichaelMure:
> IIUC the system credential storage would be used to get the private keys for signing commits created by git-bug.
> 
> When creating normal commits in the repo, the `git` CLI will use GPG. Ideally the private keys from GPG can be used with 99designs/keyring.

Yes. `keyring` can store arbitrary secrets, as long as they can be serialized into a `[]byte`. `go-git` still use GPG but it's `golang.org/x/crypto/openpgp` which will be much easier to work with than CLI tool that may or may not be there.

> In this case sorting all the Identity Versions with `sort.Sort(ByLamportTime(v.versions))` won't be possible anymore, but we'd still be able to establish an order since we know exactly which Identity Version created the operation. If an Identity is disabled/canceled, then the central authority would have to prevent accepting new commits signed by it.

Not sure why you need that. Versions within an identity already have an implied logical order defined by the order of the commits in git. It would be possible to order identities by creation or edition by adding two extra Lamport clocks there but I don't see much the use of that.

aleb:
> > In this case sorting all the Identity Versions with `sort.Sort(ByLamportTime(v.versions))` won't be possible anymore, but we'd still be able to establish an order since we know exactly which Identity Version created the operation. If an Identity is disabled/canceled, then the central authority would have to prevent accepting new commits signed by it.
> 
> Not sure why you need that. Versions within an identity already have an implied logical order defined by the order of the commits in git. It would be possible to order identities by creation or edition by adding two extra Lamport clocks there but I don't see much the use of that.

We need to be able to sort the Versions across all Identities. This way it's possible to check each version has been created by somebody with a valid key at the time.

I think we'll try with a lamport clock used only for `Version`s.

aleb:
Anything from here you can use upstream at this point?

cache/repo_cache.go
```diff
@@ -66,6 +69,10 @@ type RepoCache struct {
 
 	// the user identity's id, if known
 	userIdentityId entity.Id
+
+	// the cache of commits
+	muCommit sync.RWMutex
+	commits  map[repository.Hash]*object.Commit
```
MichaelMure, line 75: Is that really useful? Maybe I'm missing something but it looks to me that once a commit is read for validation, it's not useful anymore as it won't be shared with any other identities.

identity/identity.go
```diff
@@ -54,7 +56,12 @@ func NewIdentity(name string, email string) *Identity {
 	}
 }
 
-func NewIdentityFull(name string, email string, login string, avatarUrl string) *Identity {
```
MichaelMure, line 57: I think a key change for the identities would be to have a `KeyringAtTime(time lamport.Time) openpgp.EntityList` in `identity.Interface`. This could be used later with gogit's `Commit.Verify` function to let them figure out the correct key from the possible candidate valid at that time.

MichaelMure:
Sorry it took me a while to digest all that and figure out what would be the right thing to do.

> I think we'll try with a lamport clock used only for `Version`s.

Make sense to me. That would be similar to the edit Lamport time for bugs. Might as well throw another for the creation to mirror Bug.

> Anything from here you can use upstream at this point?

Yes, but not as it is currently I'm afraid (sorry about that). There is a lot of good things though.

As you can see in https://github.com/MichaelMure/git-bug/pull/412, the go-git repo is pretty much ready to be merged (one missing thing is a migration for the bridge credentials in https://github.com/MichaelMure/git-bug-migration/, so nothing that concern you). Notably this repo will allow for a much more native and direct access to everything signature related. Hopefully this also allow to do less in git-bug and delegate to go-git and openpgp.

Additionally, something I'd like changed in this PR is how things are located. As of now the signature validation code is built *around* the core instead of integrating with it. That is, have signature validation available within `bug` and `identities`. I believe changing that would have multiple benefits:
- split the massive `validate` package into smaller and more understandable pieces: validating the whole repo would then just be iterating over all the entities and call the validation function
- allow to validate individual entities if the need arise
- critically, allow to validate the data while reading it and *before* merging remote updates

Does that make sense to you?

cache/repo_cache.go
aleb: Yes, it does not seem useful.

identity/identity.go
aleb: I did introduce a lamport clock for identity versions, but for the regular commits we don't have any lamport time. This could be used only when checking the Identity/Version commits.

MichaelMure:
@aleb FYI, https://github.com/MichaelMure/git-bug/pull/532 bring a major update to the data model that should interest you:
- allow to easily define new entities without caring about all the complex conflict resolution, reading/writing and a lot more difficult to write code (see how Bug is defined now: https://github.com/MichaelMure/git-bug/blob/master/bug/bug.go)
- support for commit signature

This should make your adventure with git-ticket much easier if you migrate as well.

As such, I'll close this PR.

Even if your code here has not been merged, it has been a very useful contribution and what is in master now is better because of it. Thank you.

PR #381, opened 2020-04-30
DOC: commands/root.go: syntax

westurner:
This probably needs to also be built in order to update the cli --help and the man page? IDK how that's done?


MichaelMure:
Just run `make`.

I merged manually with the generated doc. Thanks :)

westurner:
Thanks!

Issue #392, opened 2020-05-23
feature: ls dump issues as plaintext (markdown/org-mode)


karlicoss:
Basically a command like `git bug dump` that would dump the issues rendered as a single markdown/org-mode file (or a directory of files). This would allow to instantly search in issues within the IDE without having to switch to the terminal/web UI.

Previously I was using [org-sync](https://github.com/arbox/org-sync) for this, but just discovered `git bug`, and it seems way more mature and reasonable :)

Format is not super important, IMO just merely having it as text gives you 90% of the benefits regardless the format, but:
- markdown would be easier to support natively since github uses it
- org-mode is kinda nice because you can use map github's tags and states (open/closed). 

Possible future improvements:
- dump plaintext automatically on `git bug bridge pull`
- allow two way sync, i.e. modifying comments, adding new issues by adding new entries to text files... but that seems hard and maybe not worth it

Possibly related to:
- https://github.com/MichaelMure/git-bug/issues/12
- https://github.com/MichaelMure/git-bug/issues/342

Let me know if you think it makes sense, and if you have any thoughts on possible implementation, I would be happy to contribute. And thanks for this awesome tool :)

MichaelMure:
Hoooo, I like this idea of org-mode. I'm not so familiar with it though, could you provide an example with all kind of relevant features ?

Yes, git-bug would benefit from having this kind of export. It has been touched on in #45 and #370.

In term of UX I'm a bit unsure what is the good way forward. It could be:
- a flag of the `ls` command to change the output style
- ~a new `export` command~

As this new feature would share basically everything `ls` does to do the filtering, I think it make more sense to just add flags there to change the output (it could also be done with other commands). So let's add:

- `--json`
- `--orgmode`
- `--markdown`
- `--plain` (just the list of bugs, one title per line)



> * dump plaintext automatically on `git bug bridge pull`

I would leave that to scripting, that's what those CLI commands are for.

> * allow two way sync, i.e. modifying comments, adding new issues by adding new entries to text files... but that seems hard and maybe not worth it

Yeah, probably not worth it. Far too complex for what it would bring.

MichaelMure:
> Let me know if you think it makes sense, and if you have any thoughts on possible implementation, I would be happy to contribute.

Yes please ;-)

In term of implementation, it'd be nice if the `ls` command could do the listing/loading of bugs (as currently) and delegate the output to a set of formatters, one for each supported format.

karlicoss:
[...]

PR #403, opened 2020-06-16
[ls] add support for different output formats

Invincibot:
This PR adds support for different output formats for the ls command using the `-f` or `-format` flag. Current accepted formats are:
- `default`: outputs the normal ls result. If an invalid specifier for the flag is received or the flag is not specified, `default` will be used.
- `plain` / `plaintext`: outputs only the status of the bug and the title.
- `json`: outputs each bug as a JSON object.

Works towards #392 and gets the groundwork in for other flags/output formats in the future. Markdown and org-mode to come!


commands/ls.go
```diff
@@ -48,20 +51,125 @@ func runLsBug(cmd *cobra.Command, args []string) error {
...
+			jsonBug.Participants = append(jsonBug.Participants, JSONIdentity{
+				participant.Id.String(),
+				participant.Id.Human(),
+				participant.DisplayName(),
+				participant.Login,
+			})
+		}
+
+		jsonObject, _ := json.MarshalIndent(jsonBug, "", "    ")
+		fmt.Printf("%s\n", jsonObject)
```
MichaelMure, line 159: printing bugs one by one means the output is not a well formed JSON document. It should be an array.

Invincibot: This should be resolved now, oops.

Issue #132, opened 2019-05-07
cache: memory management


MichaelMure:
Currently, with an import from scratch by the github importer on a 700 issues real world repo, all stored in memory leads to a 67MB memory usage, so It's really premature at this point, but it will be necessary eventually.

Here is what we can do:
- [ ] implement a [LRU policy](https://en.wikipedia.org/wiki/Cache_replacement_policies#Least_recently_used_(LRU)) or equivalent, with a configurable maximum number of entities kept in memory
- optionally, allow for an importer to explicitly not store the newly created issues in the cache. This could be done by either an explicit API, or by setting the LRU limit very low.

bhavukkalra:
hello @MichaelMure could i take up this issue?
I would love to implement a least recently used cache for git-bug for proper memory utlisation

bhavukkalra:
also could i get some tips on how to proceed with this...
A heads up would be really great..

PR #407, opened 2020-06-18
Add support for read-only mode for web UI.

### LINKED ISSUE #402: [WebUI] Read-only mode for serving publicly
Hey, and thanks for git-bug, it is an amazing idea!

We've started using git bug for our community monorepo at [code.tvl.fyi](https://code.tvl.fyi) and would like to be able to host a read-only web UI so that we can link to bugs etc., without people needing to pull them locally.

It seems that the web UI currently expects to be launched with an identity, and that it always provides features such as a comment field. Have you considered adding a read-only mode which could be used for serving it publicly? Alternatively, would patches for supporting this be welcome?

Thanks!

lukegb:
Fixes #402.

1) Adds the ability to attach an identity to the context for a particular repository.
1) Modifies the GraphQL handlers to use the identity from the context for making mutations, as well as for looking up the current user's identity.
1) Modifies the file upload handler to require an identity when receiving file uploads (I'm not refactoring this to use the GraphQL file upload functionality yet).
1) Adds an error handler to the JS which detects a particular GraphQL error extension which indicates that the server would like the client to reset its store and rerun any active queries.
1) Adds a new component to the JS, "IfLoggedIn", which renders its children iff there is an active user.
1) Adds a flag to the WebUI command (--read-only), which disables attaching the default identity to the context.


tazjin:
Neat. This will still require adopting an identity, right?

Some brief experimentation in our other repo shows that it also needs a worktree, I wonder if we can get around that.

tazjin:
`nix-build` yields:

> golang.org/x/text@v0.3.3: is explicitly required in go.mod, but vendor/modules.txt indicates golang.org/x/text@v0.3.2

lukegb:
Repacked the web UI (oops) - should work now.

tazjin:
Cool, that version works fine. I've deployed this on camden as a test, and I'm wondering whether the identity and bug caches might be an issue for keeping this as a long-running process?

A workaround might be using systemd to watch the git folder and restarting appropriately, which would lead to intermittent downtime.

webui/src/pages/bug/CommentForm.graphql
```diff
@@ -1,5 +1,7 @@
 mutation AddComment($input: AddCommentInput!) {
   addComment(input: $input) {
-    operation { id }
+    operation {
+      id
+    }
```
tazjin, line 5: Is this reformat intentional?

webui/src/pages/bug/Bug.tsx
```diff
@@ -88,9 +89,11 @@ function Bug({ bug }: Props) {
       <div className={classes.container}>
         <div className={classes.timeline}>
           <TimelineQuery id={bug.id} />
-          <div className={classes.commentForm}>
-            <CommentForm bugId={bug.id} />
-          </div>
+          <ReadonlyHidden>
```
tazjin, line 92: Does this hide the elements at runtime (i.e. they're still present in the dom?), or do they not get rendered at all?

I suppose - either way - this is safe because the mutating routes don't get mounted at all.

misc/powershell_completion/git-bug
```diff
@@ -235,9 +235,10 @@ Register-ArgumentCompleter -Native -CommandName 'git-bug' -ScriptBlock {
...
             [CompletionResult]::new('-p', 'p', [CompletionResultType]::ParameterName, 'Port to listen to (default is random)')
             [CompletionResult]::new('--port', 'port', [CompletionResultType]::ParameterName, 'Port to listen to (default is random)')
+            [CompletionResult]::new('--read-only', 'read-only', [CompletionResultType]::ParameterName, 'Whether to run the web UI in read-only mode')
             break
         }
     })
     $completions.Where{ $_.CompletionText -like "$wordToComplete*" } |
         Sort-Object -Property ListItemText
-}
+}
```
tazjin, line 244: Nit: whitespace error

MichaelMure:
> `nix-build` yields:
> 
> > golang.org/x/text@v0.3.3: is explicitly required in go.mod, but vendor/modules.txt indicates golang.org/x/text@v0.3.2
> 
> Bumped to this fork [in this CL](https://cl.tvl.fyi/c/depot/+/462) and I still see the "Comment" UI, but the user display in the corner is gone.

This might be my fault, I just bumped this package with https://github.com/MichaelMure/git-bug/pull/405

MichaelMure:
> Cool, that version works fine. I've deployed this on camden as a test, and I'm wondering whether the identity and bug caches might be an issue for keeping this as a long-running process?
> 
> A workaround might be using systemd to watch the git folder and restarting appropriately, which would lead to intermittent downtime.

There is no memory management in the cache at the moment (see https://github.com/MichaelMure/git-bug/issues/132). If something is loaded it will stay there forever. That said I would be surprised if it cause actual problem as this is fairly lightweight data. Contribution welcome :)

lukegb:
> > `nix-build` yields:
> > > golang.org/x/text@v0.3.3: is explicitly required in go.mod, but vendor/modules.txt indicates golang.org/x/text@v0.3.2
> > 
> > 
> > Bumped to this fork [in this CL](https://cl.tvl.fyi/c/depot/+/462) and I still see the "Comment" UI, but the user display in the corner is gone.
> 
> This might be my fault, I just bumped this package with #405

I think we determined this was just down to some nix stuff we'd written, rather than anything actually in this repo.

> My understanding is that you only want 1 and 2, but it should be done in a way that allow 3 in the future. In particular, this means that the GraphQL request would carry user information or not. To do that, you can't replace the whole mutation resolver with a read-only one as this is done only once when the server start. Instead it should be a unique resolver that return an error if there is no user. Without changing anything to the resolver you will get that error so maybe leave it like that until we figure out how to do 3. properly ?

Makes sense - I think it makes sense to attach the identity to the context in a middleware. That way we can attach "nil" in the read-only case, and attach an actual user if we're not read only, and it's obvious where the extension point is for decoding an identity once proxy support is added (which is also something @tazjin and I are interested in)

MichaelMure:
FYI, there is an example of such a middleware on the gqlgen website: https://gqlgen.com/recipes/authentication/

webui/src/pages/bug/Bug.tsx
lukegb: This causes them to not get rendered at all - they're not present in the DOM.

But yes, as you point out, the mutating routes are inaccessible in any case.

webui/src/pages/bug/CommentForm.graphql
lukegb: I just ran `make lint-fix` so this is whatever prettify decided to do.

lukegb:
Made some changes - I've updated the PR description with what this actually does now.

identity/context.go
```diff
@@ -0,0 +1,26 @@
...
+)
+
+// identityCtxKey is a unique context key, accessible only in this struct.
+type identityCtxKey struct {
+	repo string
+}
+
+// AttachToContext attaches an Identity to a context.
+func AttachToContext(ctx context.Context, r repository.RepoCommon, u *Identity) context.Context {
+	return context.WithValue(ctx, identityCtxKey{r.GetPath()}, u)
```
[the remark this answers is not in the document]
MichaelMure: After more thought spent, I think you should store the identity's ID instead of the full identity:
- the repo to search in is already known from the request: the GraphQL query has it already. The git file upload and retrieval handler don't, but that's because they only work with a single repo for now. This will need to change.
- a search in the cache with and identity ID is instant.
- it will makes things easier in the mutation resolver: instead of bypassing the cache as you are doing now, you just need to load the identity and everything is done for you

graphql/resolvers/mutation.go
```diff
@@ -125,13 +166,19 @@ func (r mutationResolver) OpenBug(_ context.Context, input models.OpenBugInput)
...
 	if err != nil {
 		return nil, err
 	}
 
-	op, err := b.Close()
+	id := identity.ForContext(ctx, repo)
+	if id == nil {
+		return nil, ErrNotAuthenticated
+	}
+	author := cache.NewIdentityCache(repo, id)
```
MichaelMure, line 179: This bypass the repo level caching and will break the uniqueness of an identity in memory. See my other comment on having the identity ID in the context.

graphql/resolvers/repo.go
```diff
@@ -149,13 +150,12 @@ func (repoResolver) Identity(_ context.Context, obj *models.Repository, prefix s
...
 }
 
-func (repoResolver) UserIdentity(_ context.Context, obj *models.Repository) (models.IdentityWrapper, error) {
-	excerpt, err := obj.Repo.GetUserIdentityExcerpt()
-	if err != nil {
-		return nil, err
+func (repoResolver) UserIdentity(ctx context.Context, obj *models.Repository) (models.IdentityWrapper, error) {
+	id := identity.ForContext(ctx, obj.Repo)
+	if id == nil {
+		return nil, nil
```
MichaelMure, line 156: Shouldn't that be `ErrNotAuthenticated` as well ?

graphql/resolvers/mutation.go
```diff
@@ -2,13 +2,26 @@ package resolvers
...
+	"time"
 
 	"github.com/MichaelMure/git-bug/bug"
 	"github.com/MichaelMure/git-bug/cache"
 	"github.com/MichaelMure/git-bug/graphql/graph"
 	"github.com/MichaelMure/git-bug/graphql/models"
+	"github.com/MichaelMure/git-bug/identity"
 )
 
+type errNotAuthenticated struct{}
```
MichaelMure, line 14: Maybe put that in a `errors.go` file ?

graphql/resolvers/repo.go
lukegb: IMO no: this isn't an error case - we're explicitly telling you that you don't have an identity at the moment, rather than informing you that you've tried to access an endpoint for which authentication is required.

identity/context.go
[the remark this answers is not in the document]
lukegb: The reason I added the identity ID was to (in a multi-repo world) avoid people from e.g. adding bugs to a repo in which their identity doesn't exist, but I don't know enough about what your plan for multi-repo support is.

identity/context.go
```diff
@@ -0,0 +1,26 @@
+package identity
+
+import (
+	"context"
+
+	"github.com/MichaelMure/git-bug/repository"
+)
+
```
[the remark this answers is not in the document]
lukegb: `graphql` causes a cyclic dependency as well, because graphql/resolvers is depended on by graphql, so it can't depend on graphql.

graphql/resolvers/repo.go
lukegb: I've changed my mind; it makes most of the call-sites more convenient.

tazjin:
> There is no memory management in the cache at the moment (see #132). If something is loaded it will stay there forever. That said I would be surprised if it cause actual problem as this is fairly lightweight data.

I think the problem we're more worried about is whether it actually refreshes the running web UI if the folder it looks at is changing simultaneously. Our setup would most likely point the running process directly at a Gerrit repository on disk, which is the actual `.git` that pushes end up happening to. This is fairly easy to figure out though.

MichaelMure:
> I think the problem we're more worried about is whether it actually refreshes the running web UI if the folder it looks at is changing simultaneously. Our setup would most likely point the running process directly at a Gerrit repository on disk, which is the actual `.git` that pushes end up happening to. This is fairly easy to figure out though.

So you want to accept edit from both the webUI and `git bug push` at the same time. That make sense.

When the webUI or any other command run, the cache lock the repo, which prevent any other command to process. Any ... but a push as this is git itself, not git-bug. It works well for a personal usage: you can't alter the repo concurrently and you can't pull new changes while something else happen. This mechanism doesn't protect from pushes though, which is a problem in hosting situation.

The other problem is indeed that the cache is at the moment not aware that a push happen and won't invalidate or update the entities in memory or update the index.

Maybe a server-side git hook could solve that ? This hook would monitor git pushes, look if a git-bug entity has been updated, and inform the running process (the one started with `webui`) of that through an additional API endpoint. The cache would then perform the required invalidation.

lukegb:
> > There is no memory management in the cache at the moment (see #132). If something is loaded it will stay there forever. That said I would be surprised if it cause actual problem as this is fairly lightweight data.
> 
> I think the problem we're more worried about is whether it actually refreshes the running web UI if the folder it looks at is changing simultaneously. Our setup would most likely point the running process directly at a Gerrit repository on disk, which is the actual `.git` that pushes end up happening to. This is fairly easy to figure out though.

I think we should move this discussion out of band rather than here; it's conflating this PR with a different feature request specific to our deployment :^)

MichaelMure:
Moved to #409

MichaelMure:
So, this PR made me realize that the API/webUI code was quite tangled. I went ahead and massaged that until it looked better.     

Included in the changes:
- create a new /api root package to hold all API code, migrate /graphql in there, with /http for the git handlers
- git API handlers now use the cache instead of the repo directly
- git API handlers are now tested
- git API handlers now require a "repo" mux parameter
- lots of untangling of API/handlers/middleware
- less code in commands/webui.go
- things should now be in a better position to accept external auth

Hopefully I didn't break anything in the process.

Sorry for making this PR basically unreadable. Please give it a look and let me know if that looks good to you.

webui/src/apollo.ts
```diff
@@ -8,6 +8,25 @@ import introspectionQueryResultData from './fragmentTypes';
...
+  onError: ({ graphQLErrors, networkError }) => {
+    if (graphQLErrors) {
+      console.log(graphQLErrors);
+      for (const { message, locations, path, extensions } of graphQLErrors) {
+        if (extensions && extensions.resetStore) {
+          console.log(
+            'remote end returned error requesting that we reset the store'
+          );
+          client.resetStore();
+        }
```
sandhose, line 20: This does not seem necessary and may lead to request loops. If the server asks to reset the store for a query, the client would reset the store, and therefore re-run the same query, which would make a loop.

webui/src/layout/IfLoggedIn.tsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+
+import { useCurrentIdentityQuery } from './CurrentIdentity.generated';
+
+type Props = { children: React.ReactNode };
```
sandhose, line 5: If you make that a function, it would avoid rendering the child nodes if the user is not logged in

```suggestion
type Props = { children: () => React.ReactNode };
```

webui/src/layout/IfLoggedIn.tsx
```diff
@@ -0,0 +1,14 @@
...
+const IfLoggedIn = ({ children }: Props) => {
+  const { loading, error, data } = useCurrentIdentityQuery();
+
+  if (error || loading || !data?.repository?.userIdentity) return null;
+
+  return <>{children}</>;
```
sandhose, line 11:
```suggestion
  return <>{children()}</>;
```

webui/src/pages/bug/Bug.tsx
```diff
@@ -88,9 +89,11 @@ function Bug({ bug }: Props) {
...
         <div className={classes.timeline}>
           <TimelineQuery id={bug.id} />
-          <div className={classes.commentForm}>
-            <CommentForm bugId={bug.id} />
-          </div>
+          <IfLoggedIn>
+            <div className={classes.commentForm}>
+              <CommentForm bugId={bug.id} />
+            </div>
+          </IfLoggedIn>
```
sandhose, line 96:
```suggestion
          <IfLoggedIn>
            {() => (
              <div className={classes.commentForm}>
                <CommentForm bugId={bug.id} />
              </div>
            )}
          </IfLoggedIn>
```

webui/src/apollo.ts
lukegb: Hmm. This is specifically to make sure the UI updates in response to being told "no, actually, you're not logged in"

MichaelMure:
I rebased on master and added @sandhose 's suggestions.

We now just need to resolve [this debate](https://github.com/MichaelMure/git-bug/pull/407#discussion_r444428584).

webui/src/apollo.ts
MichaelMure: @lukegb is that a problem you saw with the current UI code or is it something you expect to be a problem in the future? As @sandhose said, this can trigger a request loop.

lukegb: It's mostly something I expect to be a problem once "proper" auth exists - I mostly ran into it while restarting the git-bug server between readonly and non-readonly mode, although this is a bit of an edge case.

MichaelMure: Ok, well, let's cross that bridge when we get there then.

MichaelMure:
All good now :+1: 

Thanks for the help, sorry it became so messy.

Issue #337, opened 2020-02-21
feat: graphql transaction


MichaelMure:
Preliminary work at https://github.com/MichaelMure/git-bug/tree/graphql-transaction

The idea is to be able to
1. open a transaction
2. trigger a bunch of mutations (ex: comment+close)
3. commit the transaction, rollback on error

Benefits would be:
- allow to deal with errors on commit in a single unit of changes and leave bugs in an untouched state
- allow to deal with concurrent editing: the actual bug would not be changed until the transaction commit
- allow to stack together multiple operations in the same git commit

Code wise, the idea is, within the graphql server, attach the raw operation generated by a mutation in a list *per transaction* and apply it only on commit.

MichaelMure:
cc @sandhose

MichaelMure:
@sandhose I don't remember, does it make sense to have those explicit `startTransaction`, `commit` and `rollback` mutation or would it be better to consider all the mutation within a request to be one transaction ? Do we need to have transaction's element spread across multiple call ? Do we need to have the webUI control commit/rollback ?

MichaelMure:
After discussing with @sandhose it appears that explicit transaction are not worth the added complexity server side. 

Best to do an implicit transaction: whatever mutation exist in a single request must be executed server side **in order** within a transaction. If any one of them fail the whole thing will be discarded and no changes will be made to the data model.

MichaelMure:
Reflecting back on this, it looks like transaction are not really needed. The main use case was commenting and closing on a bug as a single step but that's pretty much it. This can be implemented as a single GraphQL mutation.

PR #408, opened 2020-06-20
GraphQL transaction

MichaelMure:
fix https://github.com/MichaelMure/git-bug/issues/337

### DIFF
--- a/bug/op_add_comment.go
+++ b/bug/op_add_comment.go
@@ -122,11 +122,23 @@ func AddComment(b Interface, author identity.Interface, unixTime int64, message 
 	return AddCommentWithFiles(b, author, unixTime, message, nil)
 }
 
+func OpAddComment(author identity.Interface, unixTime int64, message string) (*AddCommentOperation, error) {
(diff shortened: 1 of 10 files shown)


MichaelMure:
To be simplified as implicit transaction.

MichaelMure:
Random thought: transaction should either fail if the underlying bug changed between the beginning and the commit, or the transaction should lock the bug against other changes.

MichaelMure:
Also, while still being exposed in the graphql api, this concept of transaction should most likely move to the core as this will eventually be needed for concurrent access at the cache layer.

The idea is to:
- gate BugCache/Snapshot with mutexes to protect against concurrent write and read during write (RWMutex)
- reading access would 1) take a read lock 2) access the bug 3) release the lock
- editing the bug would 1) take a write lock 2) make the changes 3) *not release the write lock*
- commit the bug would release the write lock

This has some consequences:
- edition operation need to be short lived: transaction that accumulate changes elsewhere over a longer period would help
- transactions need to fail if another has modified the bug before

MichaelMure:
It can be revisited, but it seems that this is not useful at the moment.

PR #417, opened 2020-07-01
identity: remove support for legacy identity

MichaelMure:


MichaelMure:
Rebased on master, test should be fine now.

PR #410, opened 2020-06-23
Add formatting options to the 'show' and 'user ls' commands

Invincibot:
This PR adds flags to the mentioned commands which allow the user to specify what output format they would like the commands to output in.

In addition, this adds support for a new output flag: `org-mode`, a format used by emacs users. Hopefully, this improves the functionality of the commands.

Example org-mode outputs:
- ls:
```
* 04174ef Open [2020-05-23 18:01:27 +0800 CST] karlicoss (karlicoss): [https://github.com/MichaelMure/git-bug/issues/392][feature: ls dump issues as plaintext (markdown/org-mode)] :[CLI:bridge:enhancement]:
** Last Edited: 2020-06-18 19:59:59 +0800 CST
** Actors:
: 528f341 karlicoss (karlicoss)
: 571aa44 Michael Muré (MichaelMure)
: ff7d2b9 vince
** Participants:
: 528f341 karlicoss (karlicoss)
: 571aa44 Michael Muré (MichaelMure)
: ff7d2b9 vince
```

- show:
```
04174ef [open] feature: ls dump issues as plaintext (markdown/org-mode)
* Author: karlicoss (karlicoss)
* Creation Time: 2020-05-23 18:01:27 +0800 CST
* Last Edit: 2020-06-18 19:59:59 +0800 CST
* Labels:
** CLI
** bridge
** enhancement
* Actors: 
** 528f341 karlicoss (karlicoss)
** 571aa44 Michael Muré (MichaelMure)
** ff7d2b9 vince
* Participants:
** 528f341 karlicoss (karlicoss)
** 571aa44 Michael Muré (MichaelMure)
** ff7d2b9 vince
* Comments:
** #0 karlicoss (karlicoss)
: Basically a command like `git bug dump` that would dump the issues rendered as a single markdown/org-mode file (or a directory of files). This would allow to instantly search in issu
es within the IDE without having to switch to the terminal/web UI.
** #1 Michael Muré (MichaelMure)
: Hoooo, I like this idea of org-mode. I'm not so familiar with it though, could you provide an example with all kind of relevant features ?
:
: Yes, git-bug would benefit from having this kind of export. It has been touched on in #45 and #370.
...
```

- user ls:
```
* 8eec8d4 Glendon Solsberry (gms8994)
* 6310f3a Martin Delille (MartinDelille)
* 92e0130 Alexandre Franke (afranke)
* 23f2a2e Anachron (Anachron)
* 9ddf4dc Tanguy ⧓ Herrmann (dolanor)
[...]

PR #425, opened 2020-07-12
ls --format org-mode enhancements

karlicoss:
- fix and align OPEN/CLOSED states
- fix org-mode links format
- santize tags (org-mode only allows _ and @ as special characters)
- format datetimes as org-mode

Follow up of https://github.com/MichaelMure/git-bug/issues/392 and https://github.com/MichaelMure/git-bug/pull/410

### DIFF
--- a/commands/ls.go
+++ b/commands/ls.go
@@ -3,6 +3,7 @@ package commands
 import (
 	"encoding/json"
 	"fmt"
+	"regexp"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


MichaelMure:
Thanks, most helpful!

I made some light code improvement on top of it. You will also found that the command is now very much faster now that I removed the `time.Sleep(5*time.Second)` I forgot there.

PR #427, opened 2020-07-13
Trim titles in list views

wavexx:
I've been working on a GH repository which includes a lot of leading whitespace in issues due to a template including it.

I think the bug data shouldn't be altered, but it still makes sense to strip titles in list/table views so that more of the bug title is readily visible. The WebUI probably does that implicitly, while I'm unsure about other modes.


MichaelMure:
I wonder if `text.LeftPadMaxLine()` should do that really. in the mean time, that's reasonable. Thanks!

PR #428, opened 2020-07-13
Set the Fg color of the status bar to White

wavexx:
Always set the Fg color when Bg is set. This fixes poor contrast on
terminals with non-standard foreground colors.


MichaelMure:
I guess that make sense. Thanks!

Issue #330, opened 2020-02-15
bridges: generate `bridge configure` doc from the declared valid parameters


MichaelMure:
Each bridge support some but not all parameters defined in the core (see `BridgeParams`). Which are supported is announced by the result of `ValidParams()`.

To make sure to have an up to date doc and for simplicity, generating the `Example` doc of the `bridge configure` command from that would be nice.

PR #432, opened 2020-07-16
Generate bridge configure docs on runtime

vincetiu8:
This generates the documentation for each bridge implementation at runtime.


bridge/core/bridge.go
```diff
@@ -52,8 +53,21 @@ func Register(impl BridgeImpl) {
 	if bridgeLoginMetaKey == nil {
 		bridgeLoginMetaKey = make(map[string]string)
 	}
+	if bridgeValidParams == nil {
+		bridgeValidParams = make(map[string][]string)
+	}
 	bridgeImpl[impl.Target()] = reflect.TypeOf(impl).Elem()
 	bridgeLoginMetaKey[impl.Target()] = impl.LoginMetaKey()
+
+	paramMap := reflect.New(bridgeImpl[impl.Target()]).Interface().(BridgeImpl).ValidParams()
```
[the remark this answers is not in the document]
Invincibot: Oops, yep.

commands/bridge_configure.go
```diff
@@ -62,28 +59,65 @@ Private:
...
-# For Gitlab
-git bug bridge configure \
-    --name=default \
-    --target=github \
-    --url=https://github.com/michaelmure/git-bug \
-    --token=$(TOKEN)`,
+Successfully configured bridge: default`)
+	targets := bridge.Targets()
+	for i, b := range targets {
+		_, err := fmt.Fprintf(&exampleText, "# For %s\ngit bug bridge configure \\\n", strings.Title(strings.Split(b, ".")[0]))
```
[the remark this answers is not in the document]
MichaelMure: Why `strings.Split(b, ".")[0]` ?

commands/bridge_configure.go
```diff
@@ -23,15 +23,12 @@ type bridgeConfigureOptions struct {
...
-	env := newEnv()
-	options := bridgeConfigureOptions{}
+var bridgeExampleFilePath = "commands/bridge_configure_example.go"
 
-	cmd := &cobra.Command{
-		Use:   "configure",
-		Short: "Configure a new bridge.",
-		Long: `	Configure a new bridge by passing flags or/and using interactive terminal prompts. You can avoid all the terminal prompts by passing all the necessary flags to configure your bridge.`,
-		Example: `# Interactive example
+func GenBridgeConfig() error {
```
MichaelMure, line 28: What about moving all that into `gen_docs.go` ?

commands/bridge_configure.go
```diff
@@ -23,15 +23,12 @@ type bridgeConfigureOptions struct {
 	tokenStdin bool
 }
 
-func newBridgeConfigureCommand() *cobra.Command {
-	env := newEnv()
-	options := bridgeConfigureOptions{}
+var bridgeExampleFilePath = "commands/bridge_configure_example.go"
```
MichaelMure, line 26:
```suggestion
var bridgeExampleFilePath = "commands/bridge_configure_doc.go"
```

commands/bridge_configure.go
```diff
@@ -62,28 +59,65 @@ Private:
...
+
+	_ = os.Remove(bridgeExampleFilePath)
+
+	f, err := os.Create(bridgeExampleFilePath)
+	if err != nil {
+		return err
+	}
+	defer f.Close()
+
+	_, err = f.WriteString(`// Code generated by go generate; DO NOT EDIT.
```
MichaelMure, line 96:
```suggestion
	_, err = f.WriteString(`// Code generated by doc/gen_docs.go; DO NOT EDIT.
```

doc/man/git-bug-bridge-configure.1
```diff
@@ -105,28 +105,38 @@ Private:
 	\- 'repo'       : to be able to read private repositories
 
 Enter token: 87cf5c03b64029f18ea5f9ca5679daa08ccbd700
-Successfully configured bridge: default
+Successfully configured bridge: default# For Github
+git bug bridge configure \\
+    \-\-%!s(int=0)= \\
```
MichaelMure, line 110: Not sure what's happening but that doesn't seem correct.

commands/bridge_configure.go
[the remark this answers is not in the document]
Invincibot: > Why strings.Split(b, ".")[0] ?
As we are getting the name of the bridge from the type name, the types come in their own packages. For example, Jira would be expressed as `Jira.Jira`. Hence we split to cut everything after the dot.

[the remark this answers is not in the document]
Invincibot: > suggestion: split that into multiple call to `exampleText.WriteString()`. This way you don't have to use `Fprintf` which means not having to deal with an error.

exampleText.WriteString still can return error though. Should I just ignore the error?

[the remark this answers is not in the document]
MichaelMure: > As we are getting the name of the bridge from the type name, the types come in their own packages. For example, Jira would be expressed as `Jira.Jira`. Hence we split to cut everything after the dot.

As you are not exposing the reflection stuff anymore, the value you get is already clean.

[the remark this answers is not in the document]
MichaelMure: > exampleText.WriteString still can return error though. Should I just ignore the error?

Yes. It returns an error but only to match the `StringWriter` interface:
```
// WriteString appends the contents of s to b's buffer.
// It returns the length of s and a nil error.
```

MichaelMure:
Alright, so the last thing to do is to generate proper values to go with the flags. That's when it get a little tricky:
- `--target` is missing
- ideally `--url` would have a specific value for each bridge
- the name of the param is not necessarily the name of the actual flag (sorry about that)
- some params are mutually exclusive (TokenRaw and CredPrefix: let's show TokenRaw)

Otherwise it's fine to have "realistic" placeholder like `$(TOKEN)`.

doc/gen_docs.go
```diff
@@ -39,6 +45,177 @@ func main() {
...
+		}
+		exampleText.WriteString("# For ")
+		exampleText.WriteString(strings.Title(b))
+		exampleText.WriteString("\ngit bug bridge configure \\\n")
+
+		exampleText.WriteString("    --target=")
+		exampleText.WriteString(strings.ToLower(b))
+		exampleText.WriteString(" \\\n")
+
+		exampleText.WriteString("    --url=")
```
MichaelMure, line 137: --url is not meant to be a flag that exist in all bridges so it should probably be deal with below. Also ideally it would fallback to the default $(URL) placeholder but that starts to be really really nitpicking

MichaelMure: ha well, --url is not supported by the Jira bridge

Invincibot: ahhh I see. Hmm i couldn't find URL in validParams though, so i assumed all bridges supported it.

MichaelMure: it's this one: https://github.com/MichaelMure/git-bug/blob/master/bridge/core/params.go#L8

MichaelMure:
You need to rebase so that the CI pass (the gitlab issue is "fixed" on master now)

MichaelMure:
fix #330

sudoforge:

```
$ git fetch https://github.com/git-bug/git-bug.git refs/pull/432/head
$ git checkout FETCH_HEAD
```

Issue #66, opened 2018-10-07
feat: delete a bug locally


MichaelMure:
Todo:

`git bug rm <id> [<remote>]`: remove a bug locally or on a remote

Notes: this will require a way to pull only partially the bugs (for instance, only bug active in the last year).

sladyn98:
Deleting a bug locally could be a fairly easy task by removing it from the ```tree``` and the `cache`, one possible barrier could be the timestamp, does the ```func (b BugsByCreationTime)``` return a bug instance? because that would make things much more easier in terms of implementation.

pawamoy:
So there is currently no way to delete a bug locally? Like the one I just created, called "Test", with nothing in it?

sladyn98:
No actually there is no way apart from going to `git` file and deleting it. We are working on developing this feature.

pawamoy:
Okay, thanks, here's how I've done it:

```bash
rm .git/git-bug/bug-cache
rm .git/refs/bugs/BUG_ID
```
...where `BUG_ID` is of course the identifier of the bug (type the first few letters then autocomplete).

sladyn98:
Yeah that is wonderful. If you need any help fell free to ask.

sladyn98:
@MichaelMure I was trying to work on deleting a bug locally. According to me a bug needs to be delted from three different components of our software:
1) The Bug cache...
2) The local refs folder
3) The Bug Excerpt Cache(I guess this populates itself from point 1)
So The way I thought of appraching this would be create a function to remove an entry from the caches and one function to remove the same from the corresponding local refs folder.
Is this apporach correct or do you have different views?

MichaelMure:
Done! Thanks @Invincibot

PR #433, opened 2020-07-20
Add the 'rm' command

Invincibot:
This PR adds a rm command, with usage `rm <id> [<remote>]`.  
This will remove a bug locally by its bug prefix and, if specified, the name of the remote it was imported from, addressing #66.
Note this does not support removing remote bugs, as pushing deletes is not recommended. In the future, if a feature is added to selectively import remote bugs, perhaps that issue can be addressed there.

### DIFF
--- a/bug/bug.go
+++ b/bug/bug.go
@@ -244,17 +244,18 @@ func readBug(repo repository.ClockedRepo, ref string) (*Bug, error) {
 
 // RemoveBug will remove a local bug from its entity.Id
 func RemoveBug(repo repository.ClockedRepo, id entity.Id) error {
-	matching := entity.Id("")
-	fullMatches := []string{}
+	var fullMatches []string
 
 	refs, err := repo.ListRefs(bugsRefPattern + id.String())
 	if err != nil {
 		return err
-	} else if num := len(refs); num > 1 {
+	}
+	if len(refs) > 1 {
 		return NewErrMultipleMatchBug(refsToIds(refs))
-	} else if num == 1 {
-		matching = refToId(refs[0])
-		fullMatches = []string{refs[0]}
+	}
+	if len(refs) == 1 {
+		// we have the bug locally
+		fullMatches = append(fullMatches, refs[0])
 	}
 
 	remotes, err := repo.GetRemotes()
@@ -267,20 +268,17 @@ func RemoveBug(repo repository.ClockedRepo, id entity.Id) error {
 		remoteRefs, err := repo.ListRefs(remotePrefix)
 		if err != nil {
 			return err
-		} else if num := len(remoteRefs); num > 1 {
+		}
+		if len(remoteRefs) > 1 {
 			return NewErrMultipleMatchBug(refsToIds(refs))
-		} else if num == 1 {
-			id := refToId(remoteRefs[0])
+		}
+		if len(remoteRefs) == 1 {
+			// found the bug in a remote
 			fullMatches = append(fullMatches, remoteRefs[0])
-			if matching.String() == "" {
-				matching = id
-			} else if id != matching {
-				return NewErrMultipleMatchBug([]entity.Id{matching, id})
-			}
 		}
 	}
 
-	if matching == "" {
+	if len(fullMatches) == 0 {
 		return ErrBugNotExist
 	}
 
--- a/bug/bug.go
+++ b/bug/bug.go
@@ -243,15 +243,18 @@ func readBug(repo repository.ClockedRepo, ref string) (*Bug, error) {
 }
 
 // RemoveLocalBug will remove a local bug from its hash
-func RemoveLocalBug(repo repository.ClockedRepo, id entity.Id) error {
-	ref := bugsRefPattern + id.String()
-	return repo.RemoveRef(ref)
-}
-
-// RemoveRemoteBug will remove a remote bug locally from its hash
-func RemoveRemoteBug(repo repository.ClockedRepo, remote string, id entity.Id) error {
-	ref := fmt.Sprintf(bugsRemoteRefPattern, remote) + id.String()
-	return repo.RemoveRef(ref)
+func RemoveBug(repo repository.ClockedRepo, id entity.Id) error {
+	refs, err := repo.ListRefs(id.String())
+	if err != nil {
+		return err
+	}
+	for _, ref := range refs {
+		err = repo.RemoveRef(ref)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
 }
 
 type StreamedBug struct {
--- a/bug/bug.go
+++ b/bug/bug.go
@@ -242,8 +242,15 @@ func readBug(repo repository.ClockedRepo, ref string) (*Bug, error) {
 	return &bug, nil
 }
 
+// RemoveLocalBug will remove a local bug from its hash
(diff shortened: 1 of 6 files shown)


MichaelMure:
@pawamoy as you seem interested by this, could you explain a bit why it's important for you?

bug/bug.go
```diff
@@ -242,9 +242,19 @@ func readBug(repo repository.ClockedRepo, ref string) (*Bug, error) {
 	return &bug, nil
 }
 
-func RemoveLocalBug(repo repository.ClockedRepo, id entity.Id) error {
-	ref := bugsRefPattern + id.String()
-	return repo.RemoveRef(ref)
+// RemoveLocalBug will remove a local bug from its hash
+func RemoveBug(repo repository.ClockedRepo, id entity.Id) error {
+	refs, err := repo.ListRefs(id.String())
```
MichaelMure, line 247: That looks quite dangerous. Could you add a test that make sure that this function remove only this particular bug and nothing else ? Maybe start with git a repo, fill it with `FillRepo`, add/pick one bug to remove and check the count after.

commands/rm.go
```diff
@@ -0,0 +1,43 @@
...
+	"github.com/spf13/cobra"
+)
+
+func newRmCommand() *cobra.Command {
+	env := newEnv()
+
+	cmd := &cobra.Command{
+		Use:      "rm <id>",
+		Short:    "Remove an existing bug.",
+		Long:     "Remove an existing bug in the local repository. If the bug was imported from a bridge, specify the remote name to remove it from. Note removing bugs that were imported from bridges will not remove the bug remote, and will only remove the local copy of the bug.",
```
MichaelMure, line 15: outdated doc

bug/bug.go
MichaelMure: I'm surprised that even works to be honest.

Invincibot: I'll just return an error if we find more than 1 reference.

Invincibot: For the testing, i can't use `FillRepo` as it creates an import cycle with random_bugs. Should I move the tests somewhere else (i.e. in the repository package) to circumvent this problem?

MichaelMure: this cyclic thing is annoying. You can do something like that instead:
```go
func TestRemove(t *testing.T) {
	repo := repository.CreateTestRepo(false)
	defer repository.CleanupTestRepos(repo)

	repoCache, err := NewRepoCache(repo)
	require.NoError(t, err)
	
	// generate a bunch of bugs
	rene, err := repoCache.NewIdentity("René Descartes", "rene@descartes.fr")
	require.NoError(t, err)

	for i := 0; i < 100; i++ {
		_, _, err := repoCache.NewBugRaw(rene, time.Now().Unix(), "title", fmt.Sprintf("message%v", i), nil, nil)
		require.NoError(t, err)
	}

	// and one more for testing
	b1, _, err := repoCache.NewBugRaw(rene, time.Now().Unix(), "title", "message", nil, nil)
	require.NoError(t, err)
	
	// ...
}
```

MichaelMure: > I'll just return an error if we find more than 1 reference.

if the bug has been pushed to multiple git remote, it's possible to have it multiple times and it should be removed multiple times.

MichaelMure: Actually you will need to push to some remote and fetch to have the correct refs in the repo:
```go
func TestRemove(t *testing.T) {
	repo := repository.CreateTestRepo(false)
	remoteA := repository.CreateTestRepo(true)
	remoteB := repository.CreateTestRepo(true)
	defer repository.CleanupTestRepos(repo, remoteA, remoteB)

	err := repo.AddRemote("remoteA", "file://"+remoteA.GetPath())
	require.NoError(t, err)

	err = repo.AddRemote("remoteB", "file://"+remoteB.GetPath())
	require.NoError(t, err)

	repoCache, err := NewRepoCache(repo)
	require.NoError(t, err)

	// generate a bunch of bugs
	rene, err := repoCache.NewIdentity("René Descartes", "rene@descartes.fr")
	require.NoError(t, err)

	for i := 0; i < 100; i++ {
		_, _, err := repoCache.NewBugRaw(rene, time.Now().Unix(), "title", fmt.Sprintf("message%v", i), nil, nil)
		require.NoError(t, err)
	}

	// and one more for testing
	// b1, _, err := repoCache.NewBugRaw(rene, time.Now().Unix(), "title", "message", nil, nil)
	// require.NoError(t, err)

	_, err = repoCache.Push("remoteA")
	require.NoError(t, err)

	_, err = repoCache.Push("remoteB")
	require.NoError(t, err)

	_, err = repoCache.Fetch("remoteA")
	require.NoError(t, err)

	_, err = repoCache.Fetch("remoteB")
	require.NoError(t, err)
	
	// ...
}
```

Invincibot: > this cyclic thing is annoying. You can do something like that instead:  
The suggested method makes use of the repoCache, which also isn't available in the bug package. What I'll do is create two separate tests. The first one (in the bug package) will test whether a bug can be removed at all (without errors). The second (in the cache package) will test whether bug removal can cause clashes or errors and is removed from all remotes.

MichaelMure: you can use the code I gave you as a test in the `cache` package and use a variation without cache as a test in the `bug` package:
```go
func TestRemove(t *testing.T) {
	repo := repository.CreateTestRepo(false)
	remoteA := repository.CreateTestRepo(true)
	remoteB := repository.CreateTestRepo(true)
	defer repository.CleanupTestRepos(repo, remoteA, remoteB)

	err := repo.AddRemote("remoteA", "file://"+remoteA.GetPath())
	require.NoError(t, err)

	err = repo.AddRemote("remoteB", "file://"+remoteB.GetPath())
	require.NoError(t, err)

	// generate a bunch of bugs
	rene := identity.NewIdentity("René Descartes", "rene@descartes.fr")
	err = rene.Commit(repo)
	require.NoError(t, err)

	for i := 0; i < 100; i++ {
		b := NewBug()
		createOp := NewCreateOp(rene, time.Now().Unix(), "title", fmt.Sprintf("message%v", i), nil)
		b.Append(createOp)
		err = b.Commit(repo)
		require.NoError(t, err)
	}

	// and one more for testing
	b := NewBug()
	createOp := NewCreateOp(rene, time.Now().Unix(), "title", "message", nil)
	b.Append(createOp)
	err = b.Commit(repo)
	require.NoError(t, err)

	_, err = Push(repo, "remoteA")
	require.NoError(t, err)

	_, err = Push(repo, "remoteB")
	require.NoError(t, err)

	_, err = Fetch(repo, "remoteA")
	require.NoError(t, err)

	_, err = Fetch(repo, "remoteB")
	require.NoError(t, err)

	// ...
}
``

pawamoy:
@MichaelMure sorry I just saw your comment.

It feels important to me because the first time I tried `git-bug`, I created a dummy issue in a repo and then was left wondering how to delete it. To me, being able to delete a local bug allows users to experiment with `git-bug`, and therefore can increase adoption 🙂

MichaelMure:
@pawamoy that's fair, thanks

ankostis:
That's really the 2nd command you need, after experimenting with `git bug add`.
Why is it not pushed to stable?

PR #442, opened 2020-08-11
Cache lock

Invincibot:
This introduces locking into the bugcache for future updates.

Changes:
- Bugcache has Lock and Unlock methods for locking and unlocking
- Snapshot method on bugcache returns a copy of the snapshot
- Bugcache is locked internally when writing

### DIFF
--- a/cache/bug_cache.go
+++ b/cache/bug_cache.go
@@ -19,7 +19,7 @@ var ErrNoMatchingOp = fmt.Errorf("no matching operation found")
 // 3. Deal with concurrency.
 type BugCache struct {
 	repoCache *RepoCache
-	mu        *sync.RWMutex
+	mu        sync.RWMutex
 	bug       *bug.WithSnapshot
 }
 
--- a/cache/bug_cache.go
+++ b/cache/bug_cache.go
@@ -88,15 +88,17 @@ func (c *BugCache) AddCommentWithFiles(message string, files []repository.Hash) 
 
 func (c *BugCache) AddCommentRaw(author *IdentityCache, unixTime int64, message string, files []repository.Hash, metadata map[string]string) (*bug.AddCommentOperation, error) {
 	c.mu.Lock()
-	defer c.mu.Unlock()
 	op, err := bug.AddCommentWithFiles(c.bug, author.Identity, unixTime, message, files)
 	if err != nil {
+		c.mu.Unlock()
 		return nil, err
 	}
 
 	for key, value := range metadata {
 		op.SetMetadata(key, value)
 	}
+
+	c.mu.Unlock()
 
 	return op, c.notifyUpdated()
 }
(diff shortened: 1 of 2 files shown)


cache/bug_cache.go
```diff
@@ -78,6 +96,9 @@ func (c *BugCache) AddCommentWithFiles(message string, files []repository.Hash)
 }
 
 func (c *BugCache) AddCommentRaw(author *IdentityCache, unixTime int64, message string, files []repository.Hash, metadata map[string]string) (*bug.AddCommentOperation, error) {
+	c.Lock()
```
MichaelMure, line 99: If the locking is done externally with `b.Lock()` and `b.Unlock()`, you should not write lock here anymore.

You need to keep the read locks though.

cache/bug_cache.go
```diff
@@ -272,7 +317,7 @@ func (c *BugCache) SetMetadataRaw(author *IdentityCache, unixTime int64, target
 	return op, c.notifyUpdated()
 }
 
-func (c *BugCache) Commit() error {
+func (c *BugCache) Commit() error { // TODO: Do we unlock here? If so, only 1 edit possible at a time?
```
MichaelMure, line 320: Locking would be done externally, so no need to lock here anymore.

cache/repo_cache.go
```diff
@@ -209,7 +209,7 @@ func repoIsAvailable(repo repository.Repo) error {
 	// practice because using a repository will be done from user interaction
 	// or in a context where a single instance of git-bug is already guaranteed
 	// (say, a server with the web UI running). But still, that might be nice to
-	// have a mutex or something to guard that.
+	// have a mu or something to guard that.
```
MichaelMure, line 212: too eager refactoring

bug/snapshot.go
```diff
@@ -131,3 +133,14 @@ func (snap *Snapshot) HasAnyActor(ids ...entity.Id) bool {
 
 // Sign post method for gqlgen
 func (snap *Snapshot) IsAuthored() {}
+
+// Clone returns a copy of the snapshot
+func (snap *Snapshot) Clone() (*Snapshot, error) {
+	clone := &Snapshot{}
+	err := deepcopier.Copy(snap).To(clone)
```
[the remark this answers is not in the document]
MichaelMure: Also, if you do that manually, you can't have an error anymore, which will simplify a lot of things.

cache/bug_cache.go
```diff
@@ -28,19 +31,27 @@ func NewBugCache(repoCache *RepoCache, b *bug.Bug) *BugCache {
...
 }
 
 func (c *BugCache) Id() entity.Id {
+	c.mu.RLock()
+	defer c.mu.RUnlock()
 	return c.bug.Id()
 }
 
 func (c *BugCache) notifyUpdated() error {
+	c.mu.Lock()
```
MichaelMure, line 46: you are double locking here

PR #449, opened 2020-08-25
LRU Cache

### LINKED ISSUE #132: cache: memory management

Invincibot:
This PR implements a LRU into the bugCache. It currently has a fixed size of 100 and will automatically evict bugs from the cache if the number of bugs exceeds 100. Note evicting bugs doesn't result in their removal from the repository, just in memory.

Fixes #132

### DIFF
--- a/bug/bug.go
+++ b/bug/bug.go
@@ -749,22 +749,3 @@ func (bug *Bug) Compile() Snapshot {
 
 	return snap
 }
-
-// EquivalentBug returns true if two bugs are equal
-func EquivalentBug(expected, actual *Bug) bool {
-	if len(expected.packs) != len(actual.packs) {
-		return false
-	}
-
-	for i := range expected.packs {
-		for j := range expected.packs[i].Operations {
-			actual.packs[i].Operations[j].base().id = expected.packs[i].Operations[j].base().id
-		}
-	}
-
-	if expected != actual {
-		return false
-	}
-
-	return true
-}
--- a/bug/bug.go
+++ b/bug/bug.go
@@ -749,3 +749,22 @@ func (bug *Bug) Compile() Snapshot {
 
 	return snap
 }
+
+// EquivalentBug returns true if two bugs are equal
+func EquivalentBug(expected, actual *Bug) bool {
+	if len(expected.packs) != len(actual.packs) {
+		return false
+	}
+
+	for i := range expected.packs {
+		for j := range expected.packs[i].Operations {
+			actual.packs[i].Operations[j].base().id = expected.packs[i].Operations[j].base().id
+		}
+	}
+
+	if expected != actual {
+		return false
+	}
+
+	return true
+}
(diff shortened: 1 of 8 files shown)


bug/bug.go
```diff
@@ -749,3 +749,22 @@ func (bug *Bug) Compile() Snapshot {
 
 	return snap
 }
+
+// EquivalentBug returns true if two bugs are equal
+func EquivalentBug(expected, actual *Bug) bool {
```
MichaelMure, line 754: This should definitely not be a public API of this package, especially as it **change** `actual`.

cache/repo_cache_bug.go
```diff
@@ -108,36 +111,72 @@ func (c *RepoCache) writeBugCache() error {
...
+	defer c.muBug.Unlock()
+	if c.presentBugs.Len() < c.presentBugs.maxSize {
+		return nil
+	}
+
+	for _, id := range c.presentBugs.GetAll() {
+		if b := c.bugs[id]; !b.NeedCommit() {
+			b.mu.Lock()
+			c.presentBugs.Remove(id)
+			delete(c.bugExcerpts, id)
```
MichaelMure, line 173: evicting from the cache should not change `c.bugExcerpts` as those as the full collection of bug excerpt to make search on it.

MichaelMure (changes requested):
A few things left to fix but that looks pretty much done.

cache/repo_cache.go
```diff
@@ -144,7 +160,7 @@ func (c *RepoCache) Close() error {
 
 	c.identities = make(map[entity.Id]*IdentityCache)
 	c.identitiesExcerpts = nil
-	c.bugs = make(map[entity.Id]*BugCache)
+	c.bugs = nil
```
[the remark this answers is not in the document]
Invincibot: hmm, thats weird. ill change it back.

cache/lru_id_cache.go
```diff
@@ -0,0 +1,56 @@
...
+
+	"github.com/MichaelMure/git-bug/entity"
+)
+
+type LRUIdCache struct {
+	parentCache *lru.Cache
+}
+
+func NewLRUIdCache() *LRUIdCache {
+	// Ignore error here
```
MichaelMure, line 16:
```suggestion
	// we can ignore the error here as it would only fail if the size is negative.
```

MichaelMure (approved):

Issue #430, opened 2020-07-13
TermUI: dense UI


wavexx:
In all TermUI views there always seems to be a space for an empty frame around the main view, which results in two columns and most importantly two lines of wasted space.

Could this space be removed?

I could contribute such a change.

MichaelMure:
Some of the spacing are helpful for readability imho but yes, contribution much welcome. It's not something that has had a lot of attention.

wavexx:
Relevant gocui issue https://github.com/jroimartin/gocui/issues/70

![2020-07-15T122547](https://user-images.githubusercontent.com/1017726/87534506-6891c880-c696-11ea-9635-f27c1f7e3a56.png)

Still seems very readable to me thanks to the coloring. In general I do not find frames useful, unless it's a modal window where some separation is necessary.

But of course, it's to you to decide :)

 I find the bug table view quite convenient when scanning for issues compared to regular ls (with the cycle: search/narrow/examine/repeat).

"COMMENTS" still takes a lot of space IMHO, which I'd shorten to CMT, along with "LAST EDIT" (which seems always "a long while ago" for some reason), so that the author/title can get wider without having to enter the bug.

MichaelMure:
Damn, though choice. I agree that it's still readable but I also find the extra separating space visually relaxing. Maybe a good compromise would be to remove only the top spacing ? What do I know, I'm no designer ;)

> I find the bug table view quite convenient when scanning for issues compared to regular ls (with the cycle: search/narrow/examine/repeat).

Yeah that's the idea. The low level CLI commands are really meant for scripting and integration into other tools.

> "COMMENTS" still takes a lot of space IMHO, which I'd shorten to CMT

I could see that happening. Is CMT a common way to abbreviate comment though? I'm not native.

There is also a bug in this column: when no extra comment has been made on a bug after the creation, `0 💬` should be displayed there, or even nothing.

> with "LAST EDIT" (which seems always "a long while ago" for some reason)

this is due to https://github.com/MichaelMure/git-bug/issues/426

wavexx:
> Damn, though choice. I agree that it's still readable but I also find the extra separating space visually relaxing. Maybe a good compromise would be to remove only the top spacing ? What do I know, I'm no designer ;)

For terminal UIs space generally comes as a premium. Then again, it's probably one of those things that needs some sort of configuration as it's never going to satisfy everybody.

After fiddling some more, this is what I get:
[...]

PR #450, opened 2020-08-26
Packed bugview

wavexx:
This partially addresses #430 for the bug table, although this offers no configuration.


MichaelMure:
Thank you :)

MichaelMure:
I made some adjustments:
- keep the bottom spacing line: I know vertical space is expensive but I think it makes a big difference in readability
- make sure there is at least 1 space between the title and the labels
- show the number of *additional* comments instead of counting the original message
- make sure to re-render the help bar on window resize

PR #455, opened 2020-09-16
cache,bug,identity: structural change

MichaelMure:
- bug doesn't commit identities anymore, only make sure they are commit
- cache use an IdentityResolver to attach cached identities to bugs (deps injection): this make sure that we don't load multiple version of the same identity and improve perfs
- IdentityCache now are identity.Interface
- some naming simplification


MichaelMure:
cc @aleb this one might interest you

PR #459, opened 2020-09-30
webui: dependencies upgrades

sandhose:
Most upgrades are minor version without breaking changes.

One big upgrade is Apollo 2 -> 3, which required a few changes in the Apollo setup (see src/apollo.ts).

Other changes are due to changes in Prettier, requiring to add parentheses around arrow function parameters.

Because of an incompatibility in older version of typescript-eslint with newer versions of typescript, I had to switch to the "next" unstable version of react-scripts. It should not break stuff though.


MichaelMure:
image

Issue #385, opened 2020-05-09
[github] git-bug bridge pull incomplete comments wrong status


frodeaa:
There seems to be an bug in how the github bridge pull issues. Some issues gets wrong status and incomplete comment history

Tested the bridge on this repo and got wrong number of issues open, 75 vs 73. 

```
git bug bridge pull
git bug ls -s open | wc -l
75
```

frodeaa:
Notices also that there are issues with wrong title

```
 f09dbd9   open    feat: Link to code from comme… ◼ ◼  Tanguy ⧓ Herrm…     5 💬    4 minutes ago
 fd71207   open    feat: Link to code from comme… ◼ ◼  Michael Muré (…    10 💬    3 weeks ago
```

The `fd71207` has wrong title, it should be `Reflections on the change needed in the data model` https://github.com/MichaelMure/git-bug/issues/212

frodeaa:
https://github.com/MichaelMure/git-bug/issues/167 is an issue that the wrong status of open

rng-dynamics:
Fixed by #534

frodeaa:
Nice!

```
git bug ls status:open | wc -l
88
git bug ls status:closed | wc -l
128
```

PR #472, opened 2020-10-24
bridge: fix Github pull bug status/comments

frodeaa:
The iterator can in some cases return the same issue that
was just handled. It seems to be the cause why we could see
bug getting the wrong status and comments that from another issue

Fixes: https://github.com/MichaelMure/git-bug/issues/385

### DIFF
--- a/bridge/github/import.go
+++ b/bridge/github/import.go
@@ -57,12 +57,21 @@ func (gi *githubImporter) ImportAll(ctx context.Context, repo *cache.RepoCache, 
 	out := make(chan core.ImportResult)
 	gi.out = out
 
+	previousIssueId := ""
 	go func() {
 		defer close(gi.out)
 
 		// Loop over all matching issues
 		for gi.iterator.NextIssue() {
 			issue := gi.iterator.IssueValue()
+			currentIssueId := parseId(issue.Id)
+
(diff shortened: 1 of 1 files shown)


MichaelMure:
Ha, thanks a lot for looking into this. That said, it looks to me that the issue should be fixed in the iterator itself instead of working around this edge case from the outside. Any idea what's happening or when ?

frodeaa:
> Ha, thanks a lot for looking into this. That said, it looks to me that the issue should be fixed in the iterator itself instead of working around this edge case from the outside. Any idea what's happening or when ?

Suspect the query created in NextItem to fetch the next "page" of items is the problem. I'm not sure if this is hapening for every items query

MichaelMure:
Maybe you could print something or set a breakpoint when this `previousIssueId == currentIssueId` condition happen?

frodeaa:
It seems to be expected behavour that NextIssue returns the same item multiple times, as it needs to traverse the complete timeline for the issue. The problem is that the NextTimelineItem/NextTimelineItem will for some issues return timelineitem that belongs to a different issue (the next issue)

So this PR does not actually fix anything, just avoid some of the issue by just fetching the first page of timelineitems that are correct

PR #490, opened 2020-10-30
read name from both global and local config

### LINKED ISSUE #480: Cannot create bridge "user.name not configured"
I tried building from source because I wanted to see the changes from #372 (building 86faedc966), but when I then try to create a remote bridge through `git bug bridge configure` I get this error regardless of the fact that the user.name is set in my `.gitconfig` (it also worked in the latest release just now).

```
Error: user name is not configured in git yet. Please use `git config --global user.name "John Doe"`
```

Also get it if I do the manual bridge config command.

Edit: It works if I configure `user.name` and `user.email` locally in the repo.

xloem:
fixes #480


MichaelMure:
So I did forget something during this refactor. Good catch :)

PR #491, opened 2020-10-30
fix config loading during 'git-bug bridge pull'

karlicoss:
Hey, not sure how/when it broke, but hopefully the fix makes sense. I guess it's kind of unfortunate that reading config is a bit scattered across the codebase, but perhaps the test will make it easier to refactor and cleanup in the future :)

### DIFF
--- a/bridge/bridges_test.go
+++ b/bridge/bridges_test.go
@@ -1,0 +1,27 @@
+package bridge
+
+import (
+	"testing"
+	"github.com/stretchr/testify/require"
+
+	"github.com/MichaelMure/git-bug/cache"
+	"github.com/MichaelMure/git-bug/repository"
+)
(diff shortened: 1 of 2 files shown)


MichaelMure:
Interesting, thanks! It likely broke during the transition to go-git, as the config reading is a whole new codebase.

While you likely correctly identified a problem, I'll have to dig a bit deeper as I'm not sure this is the right fix. It might be better to fix the config reading code instead of fixing the problem afterwards.

MichaelMure:
Hey sorry, this got out of my radar for a while.

I tried to reproduce on the current codebase but all seems fine now. I believe this got fixed by 8a158d1f94ed74c0c3b88888fa60795373eaa159

Issue #489, opened 2020-10-30
feature: basic pipe bridge


xloem:
It would be nice to have an all-purpose cli bridge that accepts maybe http://ndjson.org/ or somesuch, for export and import of bugs, and so that users can write small special-purposes bridges in their familiar languages.

xloem:
a bridge to https://github.com/driusan/bug (which follows https://github.com/driusan/PoormanIssueTracker ) or https://github.com/dspinellis/git-issue or somesuch might work here instead of pipes, too

MichaelMure:
Could you detail a bit this idea? Not sure I follow.

If you are talking about a tool to transfer issues from say, github to gitlab, to me it would be outside of the git-bug scope. That said, it's possible as an external project can import git-bug as a library and just do that.

xloem:
I'm looking for a way to import (possibly also export) the contents of a git-bug repository, from arbitrary new archive formats.  My use case at the moment is recovering github issues from the gharchive event logs, but it seemed the niche of making a general user interface for import or export was missing.  git has `git-fast-import` and `git-fast-export` for this.  They read or write to pipes in a way that anybody can write a script to interoperate with, to import or export repositories.

I've found it is less work for me to simply augment the shell commands to accept more arguments and output more data if requested,which basically accomplishes the same goal: the using script ends up running commands instead of writing to or reading from a pipe.

Does this make sense?

In the much longer term, if there were some simple general purpose pipe format, then making many projects interoperate can get much easier (consider git-issue, BugsEverywhere, git-ssb) because it comes down to getting them to support one import and export format, rather than writing bridges between every combination of pairs.

What do you think about small obscure use cases like gharchive import?

xloem:
#501 appears to address the exporting half of this wonderfully.  Github ended up reopening the repository I was trying to recover from archive, so I no longer personally need this.  Closing since nobody else has requested it.

PR #492, opened 2020-10-31
Timestamp Arguments

xloem:
This PR adds a command line argument to bug creation, commenting, and status changing, that lets one change the timestamp from the present.

The purpose is to facilitate using git-bug inside single-use scripts to import issues from archives, addressing the reason I opened issue #489 .

The changes are minor, but I have never coded in go before, and mostly copied structures from elsewhere in the repository to accomplish this.


6543:
I would not just add option to pass unix timestamp. Instead use a time parsing lib to be able to use all kinds of time arguments.

Also we dont need other func we are no lib, just refactor the code

PR #496, opened 2020-11-09
Replace Github authorization endpoint by device authorization grant

rng-dynamics:
Fix issue #484

Suggestions for changes/improvements are very welcome.

### DIFF
--- a/bridge/github/config.go
+++ b/bridge/github/config.go
@@ -216,7 +216,7 @@ func requestUserVerificationCode(scope string) (map[string]string, error) {
 	}
 	defer resp.Body.Close()
 	if resp.StatusCode != http.StatusOK {
-		return nil, fmt.Errorf("unexpected response status code from Github API:", resp.StatusCode)
+		return nil, fmt.Errorf("unexpected response status code %d from Github API", resp.StatusCode)
 	}
 	data, err := ioutil.ReadAll(resp.Body)
 	if err != nil {
@@ -257,7 +257,7 @@ func pollGithubForAuthorization(deviceCode string, intervalSec int64) (string, e
 		}
 		defer resp.Body.Close()
 		if resp.StatusCode != http.StatusOK {
-			return "", fmt.Errorf("unexpected response status code from Github API:", resp.StatusCode)
+			return "", fmt.Errorf("unexpected response status code %d from Github API", resp.StatusCode)
 		}
 		data, err := ioutil.ReadAll(resp.Body)
 		if err != nil {
--- a/bridge/github/config.go
+++ b/bridge/github/config.go
@@ -170,21 +170,24 @@ func (*Github) ValidateConfig(conf core.Configuration) error {
 	return nil
 }
 
+type githRespT struct {
+	uri        string
+	userCode   string
+	deviceCode string
+	interval   int64
+}
+
 func requestToken() (string, error) {
 	scope, err := promptUserForProjectVisibility()
 	if err != nil {
 		return "", errors.WithStack(err)
 	}
-	ghResp, err := requestUserVerificationCode(scope)
+	resp, err := requestUserVerificationCode(scope)
 	if err != nil {
 		return "", err
 	}
-	promptUserToGoToBrowser(ghResp["verification_uri"], ghResp["user_code"])
-	interval, err := strconv.ParseInt(ghResp["interval"], 10, 64) // base 10, bitSize 64
-	if err != nil {
-		return "", errors.Wrap(err, "Error parsing integer received from Github API")
-	}
-	return pollGithubForAuthorization(ghResp["device_code"], interval)
+	promptUserToGoToBrowser(resp.uri, resp.userCode)
+	return pollGithubForAuthorization(resp.deviceCode, resp.interval)
 }
 
 func promptUserForProjectVisibility() (string, error) {
@@ -203,7 +206,7 @@ func promptUserForProjectVisibility() (string, error) {
 	return []string{"public_repo", "repo"}[index], nil
 }
 
-func requestUserVerificationCode(scope string) (map[string]string, error) {
+func requestUserVerificationCode(scope string) (*githRespT, error) {
 	params := url.Values{}
 	params.Set("client_id", githubClientID)
 	params.Set("scope", scope)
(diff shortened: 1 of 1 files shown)


bridge/github/config.go
```diff
@@ -169,63 +169,112 @@ func (*Github) ValidateConfig(conf core.Configuration) error {
...
+	params.Set("client_id", GithubClientID)
+	params.Set("scope", scope)
 	client := &http.Client{
 		Timeout: defaultTimeout,
 	}
-
-	return client.Do(req)
+	resp, err := client.PostForm("https://github.com/login/device/code", params)
+	if err != nil {
+		return nil, err
```
MichaelMure, line 206: adding some context with `errors.Wrap()` could be helpful

bridge/github/config.go
```diff
@@ -169,63 +169,112 @@ func (*Github) ValidateConfig(conf core.Configuration) error {
...
+	}
+	if resp.StatusCode != http.StatusOK {
+		defer resp.Body.Close()
```
MichaelMure, line 209: Is there no a body when `resp.StatusCode == http.StatusOK` ? Might make sense to always `Close()`

bridge/github/config.go
```diff
@@ -169,63 +169,112 @@ func (*Github) ValidateConfig(conf core.Configuration) error {
...
+	fmt.Println()
+	fmt.Println("  URL: https://github.com/login/device")
+	fmt.Println("  user authentiation code: ", code)
+	fmt.Println()
+}
 
-	err := json.Unmarshal(data, &aux)
-	if err != nil {
-		return "", err
+func pollGithubUntilUserAuthorizedGitbug(values1 *url.Values) (string, error) {
```
MichaelMure, line 224: Instead of passing a weird-ish `*url.Values` here, it quickly get cleaner if you decode the response before calling and pass `deviceCode string` instead.

MichaelMure: Also make testing easier.

bridge/github/config.go
```diff
@@ -169,63 +169,112 @@ func (*Github) ValidateConfig(conf core.Configuration) error {
...
+		resp, err := client.PostForm("https://github.com/login/oauth/access_token", params)
+		if err != nil {
+			return "", err
+		}
+		defer resp.Body.Close()
+		if resp.StatusCode != http.StatusOK {
+			bb, _ := ioutil.ReadAll(resp.Body)
+			return "", fmt.Errorf("error creating token %v, %v", resp.StatusCode, string(bb))
+		}
+		data2, err := ioutil.ReadAll(resp.Body)
```
MichaelMure, line 246: why the 2 ?

bridge/github/config.go
```diff
@@ -169,63 +169,112 @@ func (*Github) ValidateConfig(conf core.Configuration) error {
...
+			if apiError == "authorization_pending" || apiError == "slow_down" {
+				// no-op
+			} else {
+				// apiError equals on of: "expired_token", "unsupported_grant_type",
+				// "incorrect_client_credentials", "incorrect_device_code", or "access_denied"
+				return "", fmt.Errorf("error creating token %v, %v", apiError, values2.Get("error_description"))
+			}
+			time.Sleep(interval * time.Second)
+			continue
+		}
```
MichaelMure, line 270: Looks like this could be cleaner with a `switch` structure.

bridge/github/config.go
```diff
@@ -169,63 +169,112 @@ func (*Github) ValidateConfig(conf core.Configuration) error {
...
+		token = values2.Get("access_token")
+		if token == "" {
+			panic("invalid Github API response")
```
MichaelMure, line 273: panic should really be only for extreme cases. Here, somehow Github return an unexpected result. An error will do fine.

MichaelMure (commented):
I nitpicked a lot but it's really superficial. Thanks :)

bridge/github/config.go
```diff
@@ -169,63 +169,112 @@ func (*Github) ValidateConfig(conf core.Configuration) error {
...
-		Scopes      []string `json:"scopes"`
-		Note        string   `json:"note"`
-		Fingerprint string   `json:"fingerprint"`
-	}{
-		Scopes:      []string{scope},
-		Note:        note,
-		Fingerprint: randomFingerprint(),
+func requestToken() (string, error) {
+	// prompt project visibility to know the token scope needed for the repository
+	index, err := input.PromptChoice("repository visibility", []string{"public", "private"})
```
MichaelMure, line 174: Ha, maybe print the explanations for the user like before?

rng-dynamics:
I am not really sure why the tests in Travis CI are failing. Is it my error or is it a configuration problem?

rng-dynamics:
In any case, further suggestions for improvements are welcome.

rng-dynamics:
I also would like to ask you about your opinion of the new implementation of
`func requestUserVerificationCode(scope string) (map[string]string, error)`.
It returns a `map[string]string` which holds 4 values originating from the response of the Github API. I would like to use all the values but do not know what would be a clean way to return all of them.

MichaelMure:
The tests fail because of:
> bridge/github/config.go:219:25: Errorf call has arguments but no formatting directives
> bridge/github/config.go:260:25: Errorf call has arguments but no formatting directives

Easy to fix

bridge/github/config.go
```diff
@@ -169,63 +170,124 @@ func (*Github) ValidateConfig(conf core.Configuration) error {
...
 	}
-
-	return client.Do(req)
+	resp, err := client.PostForm("https://github.com/login/device/code", params)
+	if err != nil {
+		return nil, errors.Wrap(err, "error requesting user verification code")
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("unexpected response status code from Github API:", resp.StatusCode)
```
MichaelMure, line 219:
```suggestion
		return nil, fmt.Errorf("unexpected response status code %d from Github API", resp.StatusCode)
```

bridge/github/config.go
```diff
@@ -169,63 +170,124 @@ func (*Github) ValidateConfig(conf core.Configuration) error {
...
 	}
+	interval := time.Duration(intervalSec * 1100) // milliseconds, add 10% margin
+	for {
+		resp, err := client.PostForm("https://github.com/login/oauth/access_token", params)
+		if err != nil {
+			return "", errors.Wrap(err, "error polling the Github API")
+		}
+		defer resp.Body.Close()
+		if resp.StatusCode != http.StatusOK {
+			return "", fmt.Errorf("unexpected response status code from Github API:", resp.StatusCode)
```
MichaelMure, line 260:
```suggestion
			return "", fmt.Errorf("unexpected response status code %d from Github API", resp.StatusCode)
```

MichaelMure:
> I also would like to ask you about your opinion of the new implementation of
> `func requestUserVerificationCode(scope string) (map[string]string, error)`.
> It returns a `map[string]string` which holds 4 values originating from the response of the Github API. I would like to use all the values but do not know what would be a clean way to return all of them.

Returning a map is not great indeed. You can either make up your own struct with 4 fields to return those values or return 4 values + an error as go allow you to.

rng-dynamics:
I have a question since I am not very familiar with Github: Can I rebase my branch now onto your current master and force-push into my fork? I am wondering if it would create problems in this pull request.

MichaelMure:
Rebase and `push --force` is the correct way. It sometimes get github a bit confused if it can't reattach the comments at the correct place but that's not a bit deal. Just don't do it to often. Rebasing to have the updates from master is a very valid reason to do that.

rng-dynamics:
Rebase completed @MichaelMure.

MichaelMure:
Well, that's awesome, thank you!
FYI, I cleaned it up just a little bit with https://github.com/MichaelMure/git-bug/commit/9daa8ad04def334ff2ee23f85a5f583044428290, notably the `defer` in the loop (defer trigger when the function return, not after each loop, which means the bodies were accumulating in memory).

PR #500, opened 2020-11-22
github actions

MichaelMure:


MichaelMure:
Still failing on windows, not exactly sure why. I tried to fix something regarding the paths using `\` on windows instead of `/`, but it's not there yet.

MichaelMure:
Manually merged into https://github.com/MichaelMure/git-bug/pull/510

PR #501, opened 2020-11-22
commands: add a command to export bugs as raw operations (NDJSON)

MichaelMure:
@xloem is that what you expected?


xloem:
Yes, the `import` parallel to this, with inclusion for metadata, was exactly what I was yearning for when trying to interoperate with git-bug without knowing golang.  For youtube-dl, the normal github backend should be sufficient now because github restored their repository.  Still, the functionality can be very useful for future things.

MichaelMure:
Now what I wonder is:
- does that output format need to include identities ? (I would expect to have import/export for identities as well in the future, so this is a question about self-contained or not)
- does that output format need to include lamport clocks ? (not exactly easy to do as those are attached to OperationPack, not operation themselves)
- does that output format need to include operation format version ? (again, attached to OperationPack, and possibly different within the same bug)

mcepl:
Could you please rebase it against the current master? I could try to give it some testing.

mcepl:
https://youtu.be/Y6ljFaKRTrI

mcepl:
I have tried to rebase this PR myself in https://github.com/mcepl/git-bug/tree/bug-export, but I have apparently screwed up something, because on `go mod tidy` (or `go build`) I get this error:

```sh
mitmanek:git-bug (bug-export) $ go mod tidy
go: finding module for package github.com/git-bug/git-bug/bug
go: github.com/git-bug/git-bug/commands imports
	github.com/git-bug/git-bug/bug: no matching versions for query "latest"
mitmanek:git-bug (bug-export) $ go build
commands/export.go:8:2: no required module provides package github.com/git-bug/git-bug/bug; to add it:
	go get github.com/git-bug/git-bug/bug
mitmanek:git-bug (bug-export) $ 
```

MichaelMure:
@mcepl the "bug" package moved into /entities, you have to adjust the import path. You might have to adjust more like that.

mcepl:
https://github.com/mcepl/git-bug/commit/f8700d1b5771d2f96776991ade04e472e4b49bd2 is apparently not enough (see build error in the commit message).

mcepl:
And we are down to one build problem, where I am a little lost:

```
# github.com/git-bug/git-bug/commands
commands/export.go:103:16: cannot use b.Snapshot().Operations (variable of type []dag.Operation) as []dag.OperationWithApply[*bug.Snapshot] value in struct literal
```

Moreover, I am very uncertain about

https://github.com/mcepl/git-bug/blob/1793138a1b943e71a180dc6aa360b4599f97ebdd/commands/export.go#L76
```go
q, err = query.Parse(strings.Join(args, " "))
```

I probably should do something more clever, right?

sudoforge:
supplanted by #1291.

PR #504, opened 2020-11-26
Add instructions for using godoc to view developer docs to README

happybeing:
I'm not sure if you will want to add this, but thought it might be helpful as it took me a while to figure out.


sudoforge:
Thanks for the contribution! Closing this as there is a badge to the hosted godoc reference in the README.

PR #506, opened 2020-12-01
Add RepoStorage interface (for draft PR)

happybeing:
The draft PR is for sharing and discussion of code changes which change git-bug cache from using the `os` filesystem to a go-billy `billy.Filesystem`. Initial proposal and discussion is on issue https://github.com/happybeing/git-bug/issues/2

@MichaelMure In this push I've only added the RepoStorage interface to Repo as you suggested, but I'm wondering if it would be better added to GoGitRepo instead? If you try `make test` on this you'll see it generates errors due to GitRepo not implementing the storage interface, and since that code is I think deprecated maybe it would be best to add the RepoStorage i/f to GoGitRepo and move it back once the old code is removed. 

I'll try that out while awaiting comment. At last I'm getting to code but its slow as there's still much I have to learn :smile: (Also this is my first draft PR and I've only submitted a couple of PRs before this!)


MichaelMure:
I would suggest to make two new component (struct+functions) that implement this new interface: one for actually writing on disk, one for memory only. Then you can expose those implementation in the GoGitRepo, in the GitRepo and the mock repo.

This would be very similar to the Keyring: there is an implementation using go-keyring used in GitRepo and GoGitRepo and another one using `keyring.ArrayKeyring` in the mock.

Does that make sense?

happybeing:
I thought the idea was to switch `GoGitRepo` to use a `billy.Filesystem` so you can supply whatever fs you want, including one which sits on p2p storage such as Safe Network. This may be more work than what you're suggesting, but maybe it will be more capable and extendible.

This is what I've done today and I have `InitGoGitRepo()`, `InitGoGitRepoBare()` and `NewGoGitRepo()` working using go-billy/osfs and passing the first few tests using these and . I suggest I finish this of and then push it for you to look at.

What do you thing about remaining `NewGoGitRepo()` as `OpenGoGitRepo()`? I think 'new' suggests creating a repo rather than opening a repo which exists.

MichaelMure:
I went ahead and expended towards what I have in mind to hopefully give you a good footing on this idea. Let me know if that makes more sense to you.

> I thought the idea was to switch GoGitRepo to use a billy.Filesystem so you can supply whatever fs you want, including one which sits on p2p storage such as Safe Network. This may be more work than what you're suggesting, but maybe it will be more capable and extendible.

Yes but that's a different work (which will actually be simple, see below). The first problem to resolve is that some code in git-bug is going around the Repo interface and directly access the filesystem, which is problematic in your case. Reworking this code to hide the filesystem behind a `billy.Filesystem` (which means killing the `GetPath()` function) allow later to swap to a memory-only `billy` thing.

Now to run *everything* (keyring, cache ... AND git itself) in memory, we just have to compose a new Repo with all the pieces we have:
- mockRepoConfig
- mockRepoKeyring
- mockRepoCommon
- mockRepoStorage
- mockRepoClock
- the part of GoGitRepo that implement `RepoData` but instead of opening a real on-disk repo with `gogit.PlainOpen(path)` it will open/create a repo in memory (`gogit.Init(memory.NewStorage(), nil)` or something like that)

> What do you thing about remaining NewGoGitRepo() as OpenGoGitRepo()? I think 'new' suggests creating a repo rather than opening a repo which exists.

That make sense yes.

happybeing:
> the part of GoGitRepo that implement RepoData but instead of opening a real on-disk repo with gogit.PlainOpen(path) it will open/create a repo in memory (gogit.Init(memory.NewStorage(), nil) or something like that)

I already had this working and have modified TestNewGoGitRepo (gogit_test.go) to use a `billy.Filesystem` (passed in as an optional extra parameter, so it will use go-billy/osfs by default) and the first few tests are passing (`go test -v ./repository -run TestNewGoGitRepo`).

I'll have a look at your commit and see if I can understand what you're telling me. Thanks again for helping me with this.

happybeing:
I see how this works for mockRepo, and having got things working-ish for `GoGitRepo` have an idea what else we need. One thing to note is that for mock and git, you now only need the repo.path and `LocalStorage()` with the later mapping to `$GIT_DIR/git-bug/` (if I understand correctly).

For gogit we have several paths which relate differently for plain and bare repos:
- repo.path  // e.g. `/path/to/some/directory` which contains `.git/` (gitDir) for a plain repo, or is same as gitDir for a bare repo
- gitDir        // e.g. `repo.path + .git/` for a plain repo, or `repo.path` for a bare repo
- worktree   // e.g. repo.path for a plain repo and undefined for a bare repo (usually, see below)

The git-bug path will always be `gitDir/git-bug/` whereas the other paths can have different relations to each other as noted. I've never used multiple work-trees with git, but notice that you can have more than one for a given repo, which is why I've separated it from `repo.path` in the above list, but I haven't thought about support for this.

With a `GoGitRepo` determining these is a bit more complicated than for a `GitRepo` which only needs to know where `gitDir` is (although I notice you still keep `repo.path` which I want too!)

Notice that we have multiple meanings of 'storage' for a GoGitRepo:
- your `LocalStorage()` is a `billy.Filesystem` at `gitDir/git-bug/`
- when making a `GoGitRepo` we start with a `billy.Filesystem` which contains `repo.path`, and use this to create a gogit `Storage` (git files API) for the `gitDir`, and for a plain repo, also a `billy.Filesystem` with its root at the `repo.path`.

To start, I passed a `billy.Filesystem` directly into `NewGoGitRepo()` and used this to create the gogit `Storage`, the git-bug Filesystem and the worktree Filesystem. Once this is done, I think everything you need should work using the repo interfaces you have defined, so the only question is how to provide a `billy.Filesystem` when creating a `GoGitRepo`.

To support different kinds of filesystem we could add methods to set and get the overall filesystem instead of passing it to `NewGoGitRepo()`, and have it default to `billy/osfs`. I'm not sure about that, it seems restrictive than allowing the creator to specify the filesystem it wants when creating a repo.

Regardless, I think we should think about the naming as 'filesystem' and 'storage' as I think there's scope for confusion, particularly when dealing with `GoGitRepo`. So maybe instead of `RepoStorage`:
```golang
type RepoFilesystem interface {
    GitBugStore()  billy.Filesystem // A repo's `.git/git-bug/`, to be used instead of `GetPath()`
}
```
I suggest  `GitBugStore()` instead of `LocalStorage()` which to me suggests something local to the device or browser, and can be confused with gogit `Storage` which is not a filesystem.
___
Let me know what you think. Sorry this a bit wordy, but I want to make sure I understand what you want and explain what I'm thinking. The differences are fairly minor I think, so I can still get on and we can tweak things later if needed.

happybeing:
@MichaelMure I've pushed my first attempt to finish off what you started. 

It compiles, and it passes some of the `go test -v ./repository -run TestNewGoGitRepo` tests which I'm now fixing (I had a stash which was passing all of those so expect to fix that shortly). But it is failing `make test` in lots of places due to pointer errors, so I'll start looking at those. My stash code was also failing like this on `make test` even for the `TestNewGoGitRepo` test, so it may be due to concurrency.

In the mean time please can you review my code changes and check they're not way off. One thing I did is retain a `GitDirPath()` function and place it on `RepoCommon` so that it was available in `input.go` for spawning an editor. Mostly though I hope this fits what you had in mind.

happybeing:
@MichaelMure I've not yet solved the null pointer issue(s) but have squashed several other bugs and pushed my changes. 

I'd missed a few places still using os filesystem, but think everything is now using billy.Filesystem. To do this I moved RepoStorage into RepoCommon as a couple of things using RepoCommon need access to storage (I think input and select, but would need to check).

## Remaining crashes
I've looked a bit at the remaining crashes, which seem very similar. It might help if you can look at these as you might spot the problem easier than I. They appear to be in serialisation of the data from disk (e.g. reading `.git/refs/bugs/*` during `ListCommits()`).

I wonder if the data on disk is wrong because if I list the content of the files in `.git/refs/bugs/` each file contains its own name/hash, rather than the hash of a different object (as I would expect (details included after the following stack dump).

If you run `make test` you'll see most tests pass, but a few fail with similar errors such as here for TestQueries:
```
$ go test -v ./api/graphql -run TestQueries
=== RUN   TestQueries
InitGoGitRepo()... at: /tmp/132180537
Building identity cache... ListCommits() for ref:  refs/identities/011870472bf8fe404c2836618236c1388
5dfe499
repo.r.CommitObject(r.Hash()) hash: 011870472bf8fe404c2836618236c13885dfe499
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x81a7aa]

goroutine 8 [running]:
github.com/go-git/go-git/v5/storage/filesystem.(*ObjectStorage).getFromUnpacked(0xc0001868d8, 0x40fe
f82b47701801, 0x38c136826136284c, 0x99e4df85, 0x0, 0x0, 0x0, 0x0)
	/home/mrh/.gvm/pkgsets/go1.15/global/pkg/mod/github.com/go-git/go-git/v5@v5.2.0/storage/file
system/object.go:388 +0x1ca
github.com/go-git/go-git/v5/storage/filesystem.(*ObjectStorage).EncodedObject(0xc0001868d8, 0xfef82b
4770180101, 0xc136826136284c40, 0x99e4df8538, 0x0, 0x0, 0x0, 0x0)
	/home/mrh/.gvm/pkgsets/go1.15/global/pkg/mod/github.com/go-git/go-git/v5@v5.2.0/storage/file
system/object.go:322 +0x3ec
github.com/go-git/go-git/v5/plumbing/object.GetCommit(0xba5aa0, 0xc0001868c0, 0x40fef82b47701801, 0x
38c136826136284c, 0xc099e4df85, 0x0, 0x0, 0x4fab7b)
	/home/mrh/.gvm/pkgsets/go1.15/global/pkg/mod/github.com/go-git/go-git/v5@v5.2.0/plumbing/obj
ect/commit.go:55 +0x4e
github.com/go-git/go-git/v5.(*Repository).CommitObject(...)
	/home/mrh/.gvm/pkgsets/go1.15/global/pkg/mod/github.com/go-git/go-git/v5@v5.2.0/repository.g
o:1334
github.com/MichaelMure/git-bug/repository.(*GoGitRepo).ListCommits(0xc00007d7c0, 0xc000618c80, 0x38,
 0x0, 0x0, 0x2, 0xc00007f9b0, 0x3)
	/home/mrh/src/go/src/github.com/happybeing/git-bug/repository/gogit.go:632 +0x225
github.com/MichaelMure/git-bug/identity.read(0x7faafe633c58, 0xc00007d7c0, 0xc000618c80, 0x38, 0xc00
007d7c0, 0x8, 0x0)
	/home/mrh/src/go/src/github.com/happybeing/git-bug/identity/identity.go:129 +0xfd
github.com/MichaelMure/git-bug/identity.readAll.func1(0xc00002a1e0, 0x7faafe629178, 0xc00007d7c0, 0x
ab895e, 0x10)
	/home/mrh/src/go/src/github.com/happybeing/git-bug/identity/identity.go:208 +0x151
created by github.com/MichaelMure/git-bug/identity.readAll
	/home/mrh/src/go/src/github.com/happybeing/git-bug/identity/identity.go:198 +0x85
exit status 2
```
If I list the identities for the above repo, we have:
```
$ ls -al /tmp/132180537/.git/refs/identities/
total 28
drwxr-xr-x 2 mrh mrh 4096 Dec  4 17:19 .
drwxr-xr-x 6 mrh mrh 4096 Dec  4 17:19 ..
-rw-rw-r-- 1 mrh mrh   41 Dec  4 17:19 011870472bf8fe404c2836618236c13885dfe499
-rw-rw-r-- 1 mrh mrh   41 Dec  4 17:19 0f541887c87a04f6e3a896e3cad1013b011b09d8
-rw-rw-r-- 1 mrh mrh   41 Dec  4 17:19 2b4d215c16cc020b28917ffdccbb5f2e47d20076
-rw-rw-r-- 1 mrh mrh   41 Dec  4 17:19 715cf04a2ebefec568f5a659c15d7be195143324
-rw-rw-r-- 1 mrh mrh   41 Dec  4 17:19 cdc29c10ef4bfd9fc7afcfb0168418c14daa839c
```
And if I dump the contents of those files, each contains its own hash, for example:
```
cat /tmp/132180537/.git/refs/identities/011870472bf8fe404c2836618236c13885dfe499
011870472bf8fe404c2836618236c13885dfe499
```
I'm guessing that's a problem! Looks like it's nearly there though.

MichaelMure:
As discussed on gitter I went ahead and finished the implementation the way I had in mind. To avoid overwriting your commits it's in another branch at https://github.com/MichaelMure/git-bug/compare/gobilly.

Could you give it a try ? `InitMemoryGoGitRepo` especially.

What's left though is rebasing on master and deal with Bleve.

happybeing:
I've just looked through the diff of billy with master. The main functional difference from my code that I noticed is that I hadn't changed the lamport clock implementation (perhaps the cause of the remaining error in my code). Elsewhere the changes you've made seem to correspond with slight differences and streamlining (where I over complicated the repo path for example!)

The one area we might reconsider is where this branch has different functions to init GoGitRepos with each kind of billy.Filesystem. This makes it harder to use any additional billy.Filesystem implementations and leads to further duplication of code for each one. The billy branch currently has two implementations: InitGoGitRepo() and InitMemoryGoGitRepo(), but will also need two for OpenGoGitRepo() in order to open a repo in memfs. Then for my use case I'll need to add one each for Safe, someone might want to use AWS, IPFS etc.

An alternative which avoids this would be to allow the billy.Filesystem to be passed to the OpenGoGitRepo and InitGoGitRepo functions. For simplicity you can then have a default (nil billy.Filesystem) which allows you create the osfs as a default and to set an "isOsfs" flag if you needed that for clean-up (see next). 

I noted that  repo.isMemory is not being used so wondered if that might be the purpose of isMemory: i.e. to prevent the call to os.RemoveAll() for a non osfs filesystem? I think it would be better to have a way to clean up that works on all. I did find an OS.RemoveAll() in go-billy but I couldn't figure out how to access it.

Anyway, if this is passing tests I can work from it but will still need to make changes for my use case so I can use a different billy.Filesystem. Let me know if/how you would like this branch to be extended for other kinds of billy.Filesystem. I'm not sure what the best way to support a Safe p2p billy.Filesystem would be based on the scheme here, and it would be a shame to need to keep a fork just for this.

I haven't tried building yet so will get onto that.

happybeing:
Just realised that your InitMemoryGoGitRepo() API will I think work for your use case (testing), but not mine which will be simplified if it, git-bug and go-git all operate on the same billy.Filesystem. This is what I was trying to achieve in my branch. 

By having go-git git and git-bug use separate instances so there's no unified view available of the repo filesystem (.git, .git/git-bug and any worktree will all have to be separate). It might be possible to work around that, but I think it makes things harder to understand and use as an API outside of git-bug so I think I would continue with my original approach.

From here I can either help make git-bug an API/package for other use cases, or maintain a separate fork which does that and build my application using that. I'd much rather the former, but let me know what you think or if I'm missing something.

Was the difference in how my PR was from your expectation due to these different perspectives?

happybeing:
Starting from your gobilly branch I've added functions to Init and Open a GoGitRepo using a supplied billy.Filesystem as previously and these do pass all tests (including with `make test` which was failing before). So I think your lamport clocks fix may well be the difference.

I don't think this is ideal, but it will be something I can now use to progress the git portal poc while we work out what to do here. As you note there's also the bleve issue, but we should finalise the API for GoGit using an arbitrary billy fs first. My working branch is based off your gobilly, here: https://github.com/happybeing/git-bug/tree/gobilly-for-gitbug-with-api Do you want it as a PR?

MichaelMure:
> An alternative which avoids this would be to allow the billy.Filesystem to be passed to the OpenGoGitRepo and InitGoGitRepo functions. For simplicity you can then have a default (nil billy.Filesystem) which allows you create the osfs as a default and to set an "isOsfs" flag if you needed that for clean-up (see next).

IMHO, this idea breaks the layer separation and should be avoided if possible. It's acceptable for your purpose (in the form of an alternative constructor or an entirely different Repo implementation) but let's not change the normal API for that. 

> I noted that repo.isMemory is not being used so wondered if that might be the purpose of isMemory: i.e. to prevent the call to os.RemoveAll() for a non osfs filesystem? I think it would be better to have a way to clean up that works on all. I did find an OS.RemoveAll() in go-billy but I couldn't figure out how to access it.

Yes that was the idea. Did I miss something?

> EDIT: At first I thought it was passing all tests but I found that TestGoGitMemoryRepo() is commented out and fails!

Yes, I found that the config implementation in go-git doesn't behave the same for a memory-only storage :-(. Otherwise, everything works.

> By having go-git git and git-bug use separate instances so there's no unified view available of the repo filesystem (.git, .git/git-bug and any worktree will all have to be separate). It might be possible to work around that, but I think it makes things harder to understand and use as an API outside of git-bug so I think I would continue with my original approach.

Yeah, I realized that while all the other changes in the branch is fine and should probably stay like that, my memory-only Repo is not great. Maybe fitting that alternative behavior in the normal go-git repo is a mistake. Another way might be to extract all the go-git code into a reusable component, and then compose the normal repo and the memory-only repo from this and the other pieces around. But we should discuss what you really need.

happybeing:
Have a look at my now working InitFs/OpenFsGoGitRepo implementations which takes an optional billy.Filesystem and create the guy-bug filesystem, go-git storage, and optional worktree filesystem from it. Maybe using this approach will work better for your memory implementation.

Before I go further here I need to check that I can do what I'm hoping using those using git-bug to create and access bugs, and incorporate this into the poc. It's quite possible this won't work out is not a good approach, so I want to check it out early. I've shown it works for git up to a point, so hopefully it will be ok.

> Yes, I found that the config implementation in go-git doesn't behave the same for a memory-only storage :-(. Otherwise, everything works.

Maybe we need to fork go-git! Sounds odd as they claim to be aiming for complete compatibility so maybe worth raising an issue.

Regarding repo.isMemory, when I read the diff I didn't see it being used but I haven't checked the code.

happybeing:
Dang, I spoke too soon. I wasn't running the full tests with my InitFs..() functions and I still have the segmentation fault error. I've pushed a branch with a modified version of `CreateGoGitTestRepo` here: gobilly-test-initfs

There are some other errors which I wasn't getting before which I can probably fix, but the problem that I don't know how to look into is the following. This happens for three of the tests atm (but some may also fail when I fix the simpler bugs):

TestQueries
TestGoGitRepo/Plain/Data
TestReadBugs

Example:
```
$ go test -v ./api/graphql -run TestQueries
=== RUN   TestQueries
InitFsGoGitRepo()... at: /tmp/238722828
Building identity cache... panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x81ac6a]

goroutine 13 [running]:
github.com/go-git/go-git/v5/storage/filesystem.(*ObjectStorage).getFromUnpacked(0xc00018eb58, 0x90199a17485f542d, 0xc3d1eca3107c0286, 0xc0120d2fb1, 0x0, 0x0, 0x0, 0x0)
	/home/mrh/.gvm/pkgsets/go1.15/global/pkg/mod/github.com/go-git/go-git/v5@v5.2.0/storage/filesystem/object.go:388 +0x1ca
github.com/go-git/go-git/v5/storage/filesystem.(*ObjectStorage).EncodedObject(0xc00018eb58, 0x199a17485f542d01, 0xd1eca3107c028690, 0x120d2fb1c3, 0x0, 0x0, 0x0, 0x0)
	/home/mrh/.gvm/pkgsets/go1.15/global/pkg/mod/github.com/go-git/go-git/v5@v5.2.0/storage/filesystem/object.go:322 +0x3ec
github.com/go-git/go-git/v5/plumbing/object.GetCommit(0xba5a60, 0xc00018eb40, 0x90199a17485f542d, 0xc3d1eca3107c0286, 0xc0120d2fb1, 0x0, 0x0, 0xc0001d4e40)
	/home/mrh/.gvm/pkgsets/go1.15/global/pkg/mod/github.com/go-git/go-git/v5@v5.2.0/plumbing/object/commit.go:55 +0x4e
github.com/go-git/go-git/v5.(*Repository).CommitObject(...)
	/home/mrh/.gvm/pkgsets/go1.15/global/pkg/mod/github.com/go-git/go-git/v5@v5.2.0/repository.go:1334
github.com/MichaelMure/git-bug/repository.(*GoGitRepo).ListCommits(0xc0000218b0, 0xc0002275c0, 0x38, 0x0, 0x0, 0x2, 0xc0001c5020, 0x3)
	/home/mrh/src/go/src/github.com/happybeing/git-bug-wt1/repository/gogit.go:831 +0x111
github.com/MichaelMure/git-bug/identity.read(0x7effd58a51f0, 0xc0000218b0, 0xc0002275c0, 0x38, 0xc0000218b0, 0x8, 0x0)
	/home/mrh/src/go/src/github.com/happybeing/git-bug-wt1/identity/identity.go:129 +0xfd
github.com/MichaelMure/git-bug/identity.readAll.func1(0xc000034180, 0x7effd58a52d0, 0xc0000218b0, 0xab8a3e, 0x10)
	/home/mrh/src/go/src/github.com/happybeing/git-bug-wt1/identity/identity.go:208 +0x151
created by github.com/MichaelMure/git-bug/identity.readAll
	/home/mrh/src/go/src/github.com/happybeing/git-bug-wt1/identity/identity.go:198 +0x85
FAIL	github.com/MichaelMure/git-bug/api/graphql	0.035s
FAIL
```

MichaelMure:
Closed in favor of #510

Issue #315, opened 2020-02-07
GitHub bridge doesnt' handle mutation API rate limit


mih:
I have a GitHub project with many open issues: https://github.com/datalad/datalad

Running `bridge pull` for the first time, obtains a large chunk of them until the limit is exhausted.

```
% git bug bridge pull
...
new issue: 1b7fd677f16d7dfea56f0caad895c5e944b57c31
changed label: da246a71ce5a0ee9f4603b25385a9a2ab69af2921ebffe62b92740f9db66a0b6
import error: API rate limit exceeded
imported 1935 issues and 63 identities with default bridge
git bug bridge pull  336.95s user 60.75s system 15% cpu 43:33.15 total
```

I was expecting a resume (after a cool down period) would pick up a few more, but it doesn't:

```
% git bug bridge pull
import error: API rate limit exceeded
imported 0 issues and 0 identities with default bridge
git bug bridge pull  169.74s user 78.93s system 9% cpu 42:08.63 total
```

Similar runtime suggests that the same issues were processed, hence none were obtained in addition, before the same rate limit kicked in.

I can use `--since` to get a few more, but not all it seems. Is there a way to incrementally pull a large number of issues from github?

MichaelMure:
`--since` would be your best bet, but really the correct solution would be to deal with the rate limiting within the bridge itself.

MichaelMure:
To be more explicit, in a normal import, the bridge record the time of import when everything goes right so it can resume from there on the next run. Here the rate limiting is considered an error and the bridge does't record that time, which make the next run start from the beginning.

MichaelMure:
https://github.com/shurcooL/githubv4/issues/57

maxandersen:
i just hit the same - is there a way to query what was that last successfully date imported so to know which `--since` to use ?

maxandersen:
answered my self - `git bug webui` seem to show last updated by default so used that shown date for a `--since`

MichaelMure:
This is an annoying issue indeed. Maybe I should string match the error to implement a workaround until it's resolved upstream ...

bqv:
I hit the same while trying to use http://github.com/nixos/nixpkgs

bqv:
Not quite as simple with this one. I've managed to get 3.2k issues, but while datalad has ~4k, nixpkgs has nearly 90k

6543:
github return information about rate limit and when it's  allowed to call the api next, so look at this vaule and wait based on that ...

[...]

PR #516, opened 2020-12-12
Fix incomplete pull by git-bug bridge

rng-dynamics:
Work in progress. The github bridge contains a bug documented in issue #385.
This commit shows what is the problem. There might be more problems. The
previous GraphQL query did not only retrieve new timeline items, but additionally
it acted on the issues. I have changed the query for timeline items and there are
much less wrong imports now. (Are there any malformed imports left?) There
might be similar problems with user-content-edit queries. I would like to rework
the entire bridge/github/iterator.go file in the near future in order to create a
reliable fix for this bug.

### DIFF
--- a/bridge/github/iterator.go
+++ b/bridge/github/iterator.go
@@ -9,10 +9,20 @@ import (
 )
 
 type iterator struct {
-	gc        *githubv4.Client
-	since     time.Time
-	ctx       context.Context
-	err       error
+	// Github graphql client
+	gc *githubv4.Client
+
+	// The iterator will only query issues updated or created after the date given in
+	// the variable since.
+	since time.Time
+
+	// Shared context, which is used for all graphql queries.
+	ctx context.Context
+
+	// Sticky error
+	err error
+
+	// Issue iterator
 	issueIter issueIter
 }
 
@@ -40,8 +50,15 @@ type commentEditIter struct {
 }
 
 type iterVars struct {
-	index     int
-	capacity  int
+	// Iterator index
+	index int
+
+	// capacity is the number of elements (issues, issue edits, timeline items, or
+	// comment edits) to query at a time. More capacity = more used memory =
+	// less queries to make.
+	capacity int
+
+	// Variable assignments for graphql query
 	variables varmap
 }
 
@@ -55,6 +72,7 @@ func newIterVars(capacity int) iterVars {
 	}
 }
 
+// NewIterator creates and initialize a new iterator.
 func NewIterator(ctx context.Context, client *githubv4.Client, capacity int, owner, project string, since time.Time) *iterator {
 	i := &iterator{
 		gc:    client,
@@ -93,7 +111,6 @@ func (i *iterator) resetIssueVars() {
 	vars := &i.issueIter.variables
 	(*vars)["issueFirst"] = githubv4.Int(i.issueIter.capacity)
 	(*vars)["issueAfter"] = (*githubv4.String)(nil)
-	// Only query issues after the given date. This varaible is used in the GraphQL query.
 	(*vars)["issueSince"] = githubv4.DateTime{Time: i.since}
 	i.issueIter.query.Repository.Issues.PageInfo.HasNextPage = true
 	i.issueIter.query.Repository.Issues.PageInfo.EndCursor = ""
@@ -164,6 +181,8 @@ func (i *iterator) currIssueGqlNodeId() githubv4.ID {
 	return i.currIssueItem().Id
 }
 
+// NextIssue returns true if there exists a next issue and advances the iterator by one.
+// It is used to iterate over all issues. Queries to github are made when necessary.
 func (i *iterator) NextIssue() bool {
 	if i.HasError() {
 		return false
@@ -183,6 +202,7 @@ func (i *iterator) NextIssue() bool {
 	return nextIssue
 }
 
+// IssueValue returns the actual issue value.
 func (i *iterator) IssueValue() issue {
 	return *i.currIssueItem()
 }
@@ -208,6 +228,9 @@ func (i *iterator) queryIssue() bool {
 	return true
 }
 
+// NextIssueEdit returns true if there exists a next issue edit and advances the iterator
+// by one. It is used to iterate over all the issue edits. Queries to github are made when
+// necessary.
 func (i *iterator) NextIssueEdit() bool {
 	if i.HasError() {
 		return false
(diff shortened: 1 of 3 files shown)


rng-dynamics:
Note: this PR is a proof of concept.

rng-dynamics:
@MichaelMure: I am wondering if tests with a mock using httptest would be worth the effort. In my opinion tests which are independent of the actual github API would be very valuable. I just don't have experience with testing a web API client. Do you think it is possible to mock the GraphQL API with reasonable effort? What do you think?

MichaelMure:
Thanks for looking into that.

Reading the code doesn't help me much understand the issue but if you have a good idea of what it is, at least that's progress :)

> but additionally it acted on the issues

What do you mean by that?

> I would like to rework the entire bridge/github/iterator.go file in the near future in order to create a reliable fix for this bug.

Be my guest. FYI, the bridge initially was a monolith with import/export and data retrieval fused together, which made it fairly unreadable. I came up with this iterator structure to decouple those so now it's a bit better but I'll certainly admit it could be better. Curious what you have in mind.

> I am wondering if tests with a mock using httptest would be worth the effort. In my opinion tests which are independent of the actual github API would be very valuable.

I'm down on having more tests so if you feel like it make sense, go for it. A key thing the current tests achieve though is to monitor breakage against the real API that we don't control. I would assume the Github API being sort of stable (even though it does change) but it's a different game with other bridges (see https://github.com/MichaelMure/git-bug/blob/master/bridge/gitlab/export_test.go#L245-L248). It's a bit annoying because you need credentials and need to make actual requests, but it's useful imho.


Note: as you are working on your own fork, github actions doesn't get access to the repo secrets when running the tests, which means bridges are not tested in this PR. I added you as a collaborator of the repo so if you push your branch here that should be fixed.

rng-dynamics:
> >but additionally it acted on the issues
> 
> What do you mean by that?

I just tried to express that the bug is related to the handling of issue items. To be more specific: at the moment when git-bug needs to query more timeline items, then it uses the pagination data which refers to the next issue already.

> > I would like to rework the entire bridge/github/iterator.go file in the near future in order to create a reliable fix for this bug.
> 
> Be my guest. FYI, the bridge initially was a monolith with import/export and data retrieval fused together, which made it fairly unreadable. I came up with this iterator structure to decouple those so now it's a bit better but I'll certainly admit it could be better. Curious what you have in mind.

Yes, that's not a problem. I also did not intend to disapprove of the code. I am just trying to see what I can do and what I can improve.

At the moment for each issue there is one GraphQL query which fetches a list of timeline items and user content edits. If there exist more timeline items of user content edits, then git-bug queries for more. The code could get much simpler if we (1) first query a list of issues only (that is, multiple issues at once) and (2) for each issue query the timeline items and user content edits. This will increase the number of queries by (total-num-issues / query-capacity), which in my opinion should not be a problem. By simplifying the code running queries concurrently should become easy and we might get some performance improvements. I would keep the iterators function signatures the same but I would probably change most of the GraphQL queries and the interanals of the iterators accordingly.

>I'm down on having more tests so if you feel like it make sense, go for it. A key thing the current tests achieve though is to monitor breakage against the real API that we don't control. I would assume the Github API being sort of stable (even though it does change) but it's a different game with other bridges (see https://github.com/MichaelMure/git-bug/blob/master/bridge/gitlab/export_test.go#L245-L248). It's a bit annoying because you need credentials and need to make actual requests, but it's useful imho.

You are right! I could just try to provide some additional tests.

MichaelMure:
> I also did not intend to disapprove of the code.

I did not intend to be sarcastic or something. You can disapprove the code, that's how something better is built.

> This will increase the number of queries by (total-num-issues / query-capacity), which in my opinion should not be a problem.

Yeah, the rate-limiting for GraphQL is funny. It's not a matter of how many requests, it's pretty about much how much data you pull. Speaking of rate-limiting, the github bridge doesn't handle that and fail if it trigger: https://github.com/MichaelMure/git-bug/issues/315. If you happen to trigger it, maybe look at the response to see if we can match something and introduce back-off?

rng-dynamics:
>  #315

Good to know.

rng-dynamics:
The import works now.
Well ..., I have tried to write some useful tests, but all of them have been brittle and I did not add them to my branch.
That's what I have for now. (I might commit some improvements to the github-bridge in a future pull request.)

bridge/github/iterator.go
```diff
@@ -4,290 +4,228 @@ import (
...
 	"github.com/shurcooL/githubv4"
 )
 
-type indexer struct{ index int }
-
-type issueEditIterator struct {
-	index     int
-	query     issueEditQuery
-	variables map[string]interface{}
+type iterator struct {
```
MichaelMure, line 11: Do you mind copying the comments from the previous version if that's still valid?

bridge/github/iterator.go
```diff
@@ -300,102 +238,106 @@ func (i *iterator) nextValidIssueEdit() bool {
...
+	}
+	tli.index = 0
+	return true
+}
 
-	commentEdits := i.commentEdit.query.Repository.Issues.Nodes[0].Timeline.Nodes[0].IssueComment.UserContentEdits
-	// this is not supposed to happen
-	if len(commentEdits.Nodes) == 0 {
-		i.timeline.commentEdit.index = -1
+func (i *iterator) NextCommentEdit() bool {
```
MichaelMure, line 315: Same here and I suppose in others places.

MichaelMure (commented):
Thanks!

It's a bit PR so I'll have a more thorough look later, just giving some obvious comments.

Any idea what made your test flacky? How flacky was that?

rng-dynamics:
> Note: as you are working on your own fork, github actions doesn't get access to the repo secrets when running the tests, which means bridges are not tested in this PR. I added you as a collaborator of the repo so if you push your branch here that should be fixed.

Do the test work now? I still see

`=== RUN   Test_Importer
Building bug cache... Done.
    import_test.go:141: Env var GITHUB_TOKEN_PRIVATE missing
--- SKIP: Test_Importer (0.01s)`

bqv:
> might commit some improvements to the github-bridge in a future pull request

Please do! I'd love to use this more but the ratelimit is a dealbreaker right now :(

bqv:
I attempted to test this, I'm seeing:

```
import error: Parse error on "{" (LCURLY) at [1, 234]
```

rng-dynamics:
> I attempted to test this, I'm seeing:
> 
> ```
> import error: Parse error on "{" (LCURLY) at [1, 234]
> ```
From which repository are you importing?

bqv:
I tried multiple, same error on all!

rng-dynamics:
> Any idea what made your test flacky? How flacky was that?

I was experimenting with mocking the Github API. I would return hard-coded graphql responses to the shurcooL/githubv4 library. But the mocks are not simple and, I believe, hard to to maintain. Additionally, this approach is strongly coupled to the graphql queries (the type definitions in bridge/github/import_query.go). As a consequence, every time one changes the queries, then one also has to change the hard to understand mocks in the tests. When I looked at that, I thought the tests which you have now (calling the real github API) are actually better.

I am still experimenting with it. I am wondering if the github API needs to be kept out of the tests all together. The implementation of the iterator would probably get a bit less straight-forward. (E.g., I would like to try to internally split the NextIssue() call into functions nextIssueStep1(), maybeQueryIssue(), and nextIssueStep3() in such a way that maybeQueryIssue() is as simple as possible. Then we should be able to test the two functions nextIssueStep1() and nextIssueStep3() without coupling to a lot of implementation details.)

In any case, I would welcome suggestions.

bridge/github/iterator.go
rng-dynamics: Sure, I will add them soon!

MichaelMure:
> I thought I had accepted your invitation as a collaborator.
> It would just be nice to see them pass.

I believe you also need to push your branch to this repo.

MichaelMure:
You could store the response from github and make a mock that return those. It would reduce to a simple-ish `http.Handler`. But then we wouldn't test that the bridge *keep working*, that is, if github didn't change their API or something...

rng-dynamics:
Closed in favour of #534
(Switching the source repository)

Issue #493, opened 2020-11-01
Enhanced shell completion


MichaelMure:
Cobra, the library used to implement the CLI commands also generate the shell completion scripts, for bash, zsh, fish and powershell. This completion scripts are based on the description of each commands (living in https://github.com/MichaelMure/git-bug/tree/master/commands).

At the moment, those completion works decently, completing command names and flags. However it would be possible to do better. By teaching cobra (see https://github.com/spf13/cobra/blob/master/shell_completions.md#customizing-completions) about what is expected, it should be possible to:
- disable the default completion showing files when it's not appropriate
- complete based on git-bug specific terms (`git bug ls [tab][tab]` would also complete with [query qualifiers](https://github.com/MichaelMure/git-bug/blob/master/doc/queries.md))
- complete based on git-bug specific data (`git bug show [tab][tab]` would complete with bug Ids, possibly with the bug title)

Ogguz:
I want to work on this

MichaelMure:
Yes please!

krobelus:
As a first step, I have added bug ID completions. They work well in fish, I'm not sure how to make completions work in zsh.

TheNerdyHamster:
What is the status on this issue?
I would be interested to continue on it, if its not completed

PR #531, opened 2021-01-24
Complete bug IDs where appropriate

krobelus:
This completes bug IDs in bash and fish.  I haven't gotten completions to work in zsh.
In fish, the bug titles are shown as completion label, and can be searched for.
I don't know if/how other shells can show the completion label.
Maybe they don't use "\t" as separator but it seems to do no harm.

The interface does not yet seem ideal because sometimes subcommands and
bug IDs are both valid. Since there can be lots of bug IDs, they overshadow
the subcommands.
I wonder if instead of

	git bug status open 0123acd

we should use

	git bug status --open 0123acd

or if there's some other way to avoid ambiguity.

If that doesn't work out, fish has a -k/--keep-order flag, so subcommand
completions can be listed before bug IDs, but that's neither portable nor
a pretty solution.

Part of #493

### DIFF
--- a/commands/helper_completion.go
+++ b/commands/helper_completion.go
@@ -1,6 +1,7 @@
 package commands
 
 import (
+	"fmt"
 	"sort"
 	"strings"
 
@@ -21,9 +22,6 @@ func completionHandlerError(err error) (completions []string, directives cobra.S
 
 func completeBridge(env *Env) validArgsFunction {
 	return func(cmd *cobra.Command, args []string, toComplete string) (completions []string, directives cobra.ShellCompDirective) {
-		if len(args) > 0 {
-			return nil, cobra.ShellCompDirectiveNoFileComp
-		}
 		if err := loadBackend(env)(cmd, args); err != nil {
 			return completionHandlerError(err)
 		}
@@ -41,15 +39,12 @@ func completeBridge(env *Env) validArgsFunction {
 			completions[i] = bridge + "\t" + "Bridge"
 		}
 
-		return completions, cobra.ShellCompDirectiveDefault
+		return completions, cobra.ShellCompDirectiveNoFileComp
 	}
 }
 
 func completeBridgeAuth(env *Env) validArgsFunction {
 	return func(cmd *cobra.Command, args []string, toComplete string) (completions []string, directives cobra.ShellCompDirective) {
-		if len(args) > 0 {
-			return nil, cobra.ShellCompDirectiveNoFileComp
-		}
 		if err := loadBackend(env)(cmd, args); err != nil {
 			return completionHandlerError(err)
 		}
@@ -74,15 +69,12 @@ func completeBridgeAuth(env *Env) validArgsFunction {
 			completions[i] = cred.ID().Human() + "\t" + cred.Target() + " " + string(cred.Kind()) + " " + metaFmt
 		}
 
-		return completions, cobra.ShellCompDirectiveDefault
+		return completions, cobra.ShellCompDirectiveNoFileComp
 	}
 }
 
 func completeBug(env *Env) validArgsFunction {
 	return func(cmd *cobra.Command, args []string, toComplete string) (completions []string, directives cobra.ShellCompDirective) {
-		if len(args) > 0 {
-			return nil, cobra.ShellCompDirectiveNoFileComp
-		}
 		if err := loadBackend(env)(cmd, args); err != nil {
 			return completionHandlerError(err)
 		}
@@ -90,30 +82,32 @@ func completeBug(env *Env) validArgsFunction {
 			_ = env.backend.Close()
 		}()
 
-		allIds := env.backend.AllBugsIds()
-		bugExcerpt := make([]*cache.BugExcerpt, len(allIds))
-		for i, id := range allIds {
-			var err error
-			bugExcerpt[i], err = env.backend.ResolveBugExcerpt(id)
-			if err != nil {
-				return completionHandlerError(err)
-			}
-		}
-
-		completions = make([]string, len(allIds))
-		for i, id := range allIds {
-			completions[i] = id.Human() + "\t" + bugExcerpt[i].Title
-		}
-		return completions, cobra.ShellCompDirectiveDefault
+		return completeBugWithBackend(env.backend, toComplete)
 	}
+}
+
+func completeBugWithBackend(backend *cache.RepoCache, toComplete string) (completions []string, directives cobra.ShellCompDirective) {
+	allIds := backend.AllBugsIds()
+	bugExcerpt := make([]*cache.BugExcerpt, len(allIds))
+	for i, id := range allIds {
+		var err error
+		bugExcerpt[i], err = backend.ResolveBugExcerpt(id)
+		if err != nil {
+			return completionHandlerError(err)
+		}
+	}
+
+	for i, id := range allIds {
+		if strings.Contains(id.String(), strings.TrimSpace(toComplete)) {
+			completions = append(completions, id.Human()+"\t"+bugExcerpt[i].Title)
+		}
+	}
+
+	return completions, cobra.ShellCompDirectiveNoFileComp
 }
 
 func completeBugAndLabels(env *Env, addOrRemove bool) validArgsFunction {
 	return func(cmd *cobra.Command, args []string, toComplete string) (completions []string, directives cobra.ShellCompDirective) {
-		if len(args) == 0 {
-			return completeBug(env)(cmd, args, toComplete)
-		}
-
 		if err := loadBackend(env)(cmd, args); err != nil {
 			return completionHandlerError(err)
 		}
@@ -122,6 +116,10 @@ func completeBugAndLabels(env *Env, addOrRemove bool) validArgsFunction {
 		}()
 
 		b, args, err := _select.ResolveBug(env.backend, args)
+		if err == _select.ErrNoValidId {
+			// we need a bug first to complete labels
+			return completeBugWithBackend(env.backend, toComplete)
+		}
 		if err != nil {
 			return completionHandlerError(err)
 		}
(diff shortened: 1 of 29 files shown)


MichaelMure:
THIS IS AWESOME!

> This completes bug IDs in bash and fish. I haven't gotten completions to work in zsh.

I think the ZSH completion code in cobra does support yet this new completion framework. I think it will work eventually, without any changes required on the git-bug side.

> In fish, the bug titles are shown as completion label, and can be searched for.
> I don't know if/how other shells can show the completion label.
> Maybe they don't use "\t" as separator but it seems to do no harm.

I think the `\t` thing is to tell cobra about the label associated with the completion. If the shell doesn't support that, it will just get ignored.

> The interface does not yet seem ideal because sometimes subcommands and
> bug IDs are both valid. Since there can be lots of bug IDs, they overshadow
> the subcommands.
> I wonder if instead of
> 
> ```
> git bug status open 0123acd
> ```
> 
> we should use
> 
> ```
> git bug status --open 0123acd
> ```
> 
> or if there's some other way to avoid ambiguity.

Hmmm dunno. Even with changing `status` we'll have the problem with other commands that have subcommands. I think the `docker` completion has the same "issue" and it still works AFAIK. Now, if the completion order is not maintained between your code and the shell, maybe that's something worth reporting to https://github.com/spf13/cobra?

MichaelMure (approved):

krobelus:
(I'll push a minor fix - completions don't work yet for a command like `git bug comment add `)

commands/helper_completion.go
```diff
@@ -0,0 +1,49 @@
+package commands
+
```
krobelus, line 2: I wasn't sure about the filename, since most files inside this package should be commands, but at least `env.go` isn't.
I didn't find a good pattern, this was just what I came up with.

krobelus:
> Hmmm dunno. Even with changing status we'll have the problem with other commands that have subcommands. I think the docker completion has the same "issue" and it still works AFAIK.

Yeah, `--open` is probably no good, that was just a quick thought.
I can't think of a docker command that has the same issue. I think it's always be possible to know if the next argument is a container/image or something else.
Another way to avoid the ambiguity would be to split it into two subcommands, `comment show` and `comment add`, though that's ugly. Anyway, it's not a big deal, and I'm quite new to git-bug, so I'll have a better perspective as I use it more.

> Now, if the completion order is not maintained between your code and the shell, maybe that's something worth reporting to https://github.com/spf13/cobra?

Ok, I'll report that. Looks like there are some other fish-specific issues over there that need addressing, I'll take care of those.

krobelus:
Ok, I've extended this to all subcommands and flags. This should cover all subcommands.
The label completion is especially clever, because it also takes into account labels which are already on the commandline.

In fish, `git-bug ...` completions work fine but `git bug ...` completions need a small wrapper, which will be merged soon (https://github.com/fish-shell/fish-shell/pull/7652).. With that, fish completions can be enabled with
`ln -s $PWD/misc/fish_completion/git-bug ~/.config/fish/completions/git-bug.fish`

commands/helper_completion.go
MichaelMure: LGTM

commands/helper_completion.go
```diff
@@ -0,0 +1,355 @@
...
+		}
+		if err := loadBackend(env)(cmd, args); err != nil {
+			return completionHandlerError(err)
+		}
+		defer func() {
+			_ = closeBackend(env)(cmd, args)
+		}()
+
+		q := query.NewQuery()
+		allIds, err := env.backend.QueryBugs(q)
```
MichaelMure, line 95: `AllBugsIds` would avoid the more complex querying system.

commands/helper_completion.go
```diff
@@ -0,0 +1,355 @@
...
+
+		completions = make([]string, len(allIds))
+		for i, id := range allIds {
+			completions[i] = id.Human() + "\t" + bugExcerpt[i].Title
+		}
+		return completions, cobra.ShellCompDirectiveDefault
+	}
+}
+
+func completeBugAndLabels(env *Env, addOrRemove bool) validArgsFunction {
```
MichaelMure, line 116: you can use `RepoCache.ValidLabels()` for this one

MichaelMure: Ho wait forget that, you are doing smarter than I thought. Neat!

doc/gen_docs.go
```diff
@@ -21,22 +20,15 @@ func main() {
 		"Markdown": genMarkdown,
 	}
 
-	var wg sync.WaitGroup
```
MichaelMure, line 24: why are you removing that? Fast build is nice

krobelus: See the commit message - without this commit I usually get a data race when running `make`.
I didn't really investigate what the real fix should be. Do you think we should use global variables for the Cobra commands?

MichaelMure: It's the opposite actually. Cobra use a global map [here](https://github.com/spf13/cobra/blob/master/custom_completions.go#L21). This global map is shared even if multiple independent root command are created.

It looks like it could be easily solved in cobra though ...

MichaelMure: well, not really easy but doable

MichaelMure:
I was going to report the issue to cobra but I see you did it already ;)

I made it into its own issue though, with some extra possible solutions: https://github.com/spf13/cobra/issues/1320

MichaelMure:
Let's wait a bit to see if something is moving on the cobra side. We'll merge either way at the very last before the next version, don't worry.

MichaelMure:
The cobra issue got resolved upstream, there is hope! https://github.com/spf13/cobra/pull/1438

6543:
cobra bump is here ... need to be tested and then we can add it in ?!

6543:
ok tested and bump is in @krobelus can you rebase?

MichaelMure:
hey hello,

I finally get more and more time to work on this project, sorry for the crazy long delay.

I've been trying out this PR to see if everything was fine. The build process got fixed indeed but ... the completion doesn't work for me. It might be because of the cobra upgrade as they made some related change, but they claim to have kept backward compatibility. I've looked around and everything seems fine.

Any idea? It's probably something silly.

MichaelMure:
It's also possible that I'm just using it wrong.

krobelus:
fish and bash completions seem to work (although bash only works with `git-bug`, not `git bug`); added steps to the commit message. zsh didn't work on my brief attempt.

MichaelMure:
I tried a few things, I can't get it to work with bash (I didn't try the others):
- a simple `source misc/bash_completion/git-bug` bring the completion of the sub-command names, but not the options (`--help` and so on, which was working at some point), and `git-bug select ^I` gives filename completion.
- your command `bash --rcfile <(echo source misc/bash_completion/git-bug)` brings a new shell but any attempt at completion yield `bash: _get_comp_words_by_ref : command not found`

commands/helper_completion.go
```diff
@@ -0,0 +1,350 @@
...
+		if err != nil {
+			return completionHandlerError(err)
+		}
+
+		completions = make([]string, len(bridges))
+		for i, bridge := range bridges {
+			completions[i] = bridge + "\t" + "Bridge"
+		}
+
+		return completions, cobra.ShellCompDirectiveDefault
```
MichaelMure, line 44: Out of curiosity, why not `ShellCompDirectiveNoFileComp` here?

commands/helper_completion.go
```diff
@@ -0,0 +1,350 @@
...
+			completions[i] = cred.ID().Human() + "\t" + cred.Target() + " " + string(cred.Kind()) + " " + metaFmt
+		}
+
+		return completions, cobra.ShellCompDirectiveDefault
+	}
+}
+
+func completeBug(env *Env) validArgsFunction {
+	return func(cmd *cobra.Command, args []string, toComplete string) (completions []string, directives cobra.ShellCompDirective) {
+		if len(args) > 0 {
```
MichaelMure, line 83: It'd be nice to complete even if `len(args) == 0`. That's one of the problem I had, I tried to complete `git-bug select` which didn't work as your code returns an empty array.
If you did that to avoid returning thousands of completion proposition, maybe we should have an arbitrary limit?

commands/helper_completion.go
```diff
@@ -0,0 +1,350 @@
...
+			return completionHandlerError(err)
+		}
+		defer func() {
+			_ = env.backend.Close()
+		}()
+
+		allIds := env.backend.AllBugsIds()
+		bugExcerpt := make([]*cache.BugExcerpt, len(allIds))
+		for i, id := range allIds {
+			var err error
```
MichaelMure, line 96: what about filtering here to only output the ones with `toComplete`as prefix?

MichaelMure (commented):
I'm trying to figure out what's happening. The following actually returns something that looks ok:
```
: git-bug __complete select 68
681b740	sqdsqdqsdqsdsq
e4c9831	sqdjlksqdsdiqolz
:0
Completion ended with directive: ShellCompDirectiveDefault
````

And yet, when I try `git-bug select 68^I` nothing happens.

MichaelMure:
I found the root cause, `GenBashCompletionV2` need to be called instead of the legacy one.

krobelus:
> +func completeBug(env *Env) validArgsFunction {
> +	return func(cmd *cobra.Command, args []string, toComplete string) (completions []string, directives cobra.ShellCompDirective) {
> +		if len(args) > 0 {
> 
> It'd be nice to complete even if `len(args) == 0`.

> That's one of the problem I had, I tried to complete `git-bug select` which didn't work as your code returns an empty array.

Right, you found another bug!
I'm afraid I don't have much time these days, so if someone wants to take over go ahead.

> If you did that to avoid returning thousands of completion proposition, maybe we should have an arbitrary limit?

I think shells will already filter completions, so I don't think we need to.
In particular, fish sometimes shows all completions for which the commandline's token is a subsequence of the completion.
So if we filter, we might actually lose some valid completions, and I don't think there's a need filter unless we are returning many megabytes of data.

MichaelMure:
With @tst2005 fix for bash and a bunch of small adjustments I think we have something really cool!

Thanks @krobelus, this is awesome.

PR #534, opened 2021-01-28
Fix github bridge import

rng-dynamics:
Supersedes #516

Deals with #385

### DIFF
--- a/bridge/github/import.go
+++ b/bridge/github/import.go
@@ -100,7 +100,7 @@ func (gi *githubImporter) ImportAll(ctx context.Context, repo *cache.RepoCache, 
 	return out, nil
 }
 
-func (gi *githubImporter) ensureIssue(repo *cache.RepoCache, issue issueTimeline) (*cache.BugCache, error) {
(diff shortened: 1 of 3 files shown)


bqv:
(and potentially #315)

MichaelMure:
@rng-dynamics are you done with it now?

MichaelMure (approved):
LGTM! Looks simpler indeed.

MichaelMure:
fix #385

MichaelMure:
Based on what you have seen working on this, what's missing? I'd say maybe preloading and rate-limiting handling (#315).

rng-dynamics:
> @rng-dynamics are you done with it now?

I am done with it.

rng-dynamics:
> Based on what you have seen working on this, what's missing? I'd say maybe preloading and rate-limiting handling (#315).

Rate-limiting seems to be the biggest problem. The current implementation does not use the rate-limit points in a very economic way. I am already working on a new version which should be able to pull more issues while consuming less rate-limit points. Nevertheless, for repositories with a big number of issues we will probably have to introduce some kind of waiting when the rate-limit points are exhausted.
Preloading is also on my todo-list.

MichaelMure:
:tada: :+1:

PR #537, opened 2021-01-30
Fix #158

### LINKED ISSUE #158: WebUI - allow to create bugs
If iI did not miss anything, the Web-UI is currently read-only.
To be ablle to create bugs, comments, edit, remove and so on in a way that works for any kind of (non-technical) users, is very important for any project that might has such users.

An alternative solution coudl be, to make an alreayd existing GUI/WebUI to use git-bug as a backend, like suggeste for example in https://github.com/MichaelMure/git-bug/issues/80.

What are your thoughs about this, how hard would it be to implement it, what problems do you see?

claudioantonio:
I think its a implementation good enough to integrate into master 🤞

### DIFF
--- a/webui/package-lock.json
+++ b/webui/package-lock.json
@@ -17570,9 +17570,9 @@
       "integrity": "sha1-1PRWKwzjaW5BrFLQ4ALlemNdxtw="
     },
     "prettier": {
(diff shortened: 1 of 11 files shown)


claudioantonio:
Closing to fix simple issues

PR #538, opened 2021-01-30
Webui - Allow to create bugs #158

claudioantonio:


MichaelMure:
Could you make some screenshot of what it looks like?

webui/src/pages/list/ListQuery.css
```diff
@@ -0,0 +1,17 @@
+header {
```
MichaelMure, line 1: all the styling is done with CSS-in-JS (see https://material-ui.com/styles/basics/), so this piece should be done the same way (that is, not in pure CSS).

webui/src/pages/list/ListQuery.tsx
```diff
@@ -95,6 +98,13 @@ const useStyles = makeStyles<Theme, StylesProps>((theme) => ({
       padding: theme.spacing(2, 3),
     },
   },
+  gitbugButton: {
```
MichaelMure, line 101: it's a little weird to have the `gitbug` namespace here. Is there a reason? If you can't use `button`, maybe `newButton` or something?

MichaelMure (changes requested):
Looks good from what I know about react, but maybe some part should be more seamless with the rest of the code?

claudioantonio:
For sure!
First screenshot shows the first page where we have all bugs. New issue button was added.
image

When we click on the new issue button then new issue form is presented.
image

Submit new issue only enables when title and message are not empty.
image

After submitting, If You go back to the bug list page You will see the new bug created! 😉 
If You click in the recently created bug, You will see its details and will be able to add more comments.
image

Good enough?

webui/src/pages/list/ListQuery.css
claudioantonio: Oh I just forgot to do it. That´s simple! 😉 
May I do it when fixing #539 which is another simple issue?

webui/src/pages/list/ListQuery.tsx
claudioantonio: Well... I agree with You. If I rename to "greenButton" will It be ok for You?
I would like to use this to apply the green style (normal and hover) in all situations it is indicated, as You can see in the screenshots.
May I do it when fixing #539 which is another simple issue?

MichaelMure: I guess `greenButton` works.
If you do #539 in the same branch/PR it's fine, otherwise it's best to fix the code here so it can be merged.

MichaelMure:
Yep, looks good!

webui/src/pages/list/ListQuery.css
claudioantonio: I did a commit right now. ;-)

webui/src/pages/list/ListQuery.tsx
claudioantonio: I did a commit right now. ;-)

MichaelMure (approved):
LGTM!

Just make sure the linting is fine (`make lint` and `make fix-lint`), and you can merge.

claudioantonio:
I ran "npm run lint -- --fix" before the last commit.
I ran the same command now and no warnings neither errors were found.
I don´t have write access to merge the pull request. 😉

MichaelMure:
:+1:

PR #542, opened 2021-02-02
Commit for #535

### LINKED ISSUE #535: WebUI - Allow to change a bug title
Today when We select a bug It´s only possible to add comments but in the same page should be possible to change bug title too.

claudioantonio:
Fix #535

### DIFF
--- a/webui/src/components/BugTitleForm/BugTitleForm.tsx
+++ b/webui/src/components/BugTitleForm/BugTitleForm.tsx
@@ -126,13 +126,16 @@ function BugTitleForm({ bug }: Props) {
         />
         <div className={classes.editButtonContainer}>
           <Button
+            size="small"
             variant="contained"
             type="submit"
             disabled={issueTitle.length === 0}
           >
             Save
           </Button>
-          <Button onClick={() => cancelChange()}>Cancel</Button>
+          <Button size="small" onClick={() => cancelChange()}>
+            Cancel
+          </Button>
         </div>
       </form>
     );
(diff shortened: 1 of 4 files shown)


webui/src/components/BugTitleForm/BugTitleForm.tsx
```diff
@@ -0,0 +1,181 @@
...
+      return false;
+    }
+  }
+
+  function submitNewTitle() {
+    if (!isFormValid()) return;
+    setTitle({
+      variables: {
+        input: {
+          prefix: bug.humanId,
```
MichaelMure, line 90: you are using `humanId` here as well. It works but can cause problem if you have multiple bugs with the same prefix. Better use `id`.

claudioantonio: Fixed.
I also did a find in files and this was the only reference to humanId.

webui/src/components/BugTitleForm/SetTitle.graphql
```diff
@@ -0,0 +1,7 @@
+mutation setTitle($input: SetTitleInput!) {
+  setTitle(input: $input) {
+      bug {
+        humanId
```
MichaelMure, line 4: this should be `id`, not `humanId`

MichaelMure (changes requested):

MichaelMure:
I merged #545 which include the same changes are here if I'm not mistaken, so this one is not necessary anymore.

Your chaining of branches/PRs make things quite confusing (and probably more difficult for you). Could you try to either always branch from master or make a single branch/PR?

claudioantonio:
You are right @MichaelMure It turned things more complicated for me...
For sure. In the future I will follow Your suggestions.
Sorry for the inconvenience.

PR #545, opened 2021-02-02
Webui 544

### LINKED ISSUE #544: WebUI - Allow to create new bug in the bug detail page
Sometimes when reading a timeline we detect is necessary to create other issues. So this feature will make it more agile than force user go back to the list page to create a new issue.

claudioantonio:
Fix #544

### DIFF
--- a/webui/src/components/BugTitleForm/BugTitleForm.tsx
+++ b/webui/src/components/BugTitleForm/BugTitleForm.tsx
@@ -1,0 +1,181 @@
+import React, { useState } from 'react';
+
+import {
+  Button,
+  fade,
+  makeStyles,
+  TextField,
+  Typography,
+} from '@material-ui/core';
+
+import { TimelineDocument } from '../../pages/bug/TimelineQuery.generated';
+import Author from 'src/components/Author';
+import Date from 'src/components/Date';
+import { BugFragment } from 'src/pages/bug/Bug.generated';
+
+import { useSetTitleMutation } from './SetTitle.generated';
+
+/**
+ * Css in JS styles
+ */
+const useStyles = makeStyles((theme) => ({
+  header: {
+    display: 'flex',
+    flexDirection: 'column',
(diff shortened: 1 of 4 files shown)


webui/src/components/BugTitleForm/BugTitleForm.tsx
```diff
@@ -0,0 +1,197 @@
...
+interface Props {
+  bug: BugFragment;
+}
+
+/**
+ * Component for bug title change
+ * @param bug Selected bug in list page
+ */
+function BugTitleForm({ bug }: Props) {
+  const [bugTitleEditable, setBugTitleEditable] = useState(false);
```
MichaelMure, line 79: maybe name that one `bugTitleEdition` ? "editable" makes me think about having the right to edit or not.

MichaelMure (approved):
Just some nitpicking about a naming. Looks good!

webui/src/components/BugTitleForm/SetTitle.graphql
```diff
@@ -0,0 +1,7 @@
+mutation setTitle($input: SetTitleInput!) {
+  setTitle(input: $input) {
+      bug {
+        humanId
```
MichaelMure, line 4: you should retrieve the complete `id` here. `humanId` is a truncated Id, that's not enough to update the cache.

MichaelMure (changes requested):
actually, there is this

webui/src/components/BugTitleForm/BugTitleForm.tsx
claudioantonio: I´ll change it.

webui/src/components/BugTitleForm/SetTitle.graphql
claudioantonio: Oh! I´ll fix it!

PR #559, opened 2021-02-09
Webui 543 ifloggedin

### LINKED ISSUE #543: WebUI - Protection for all write-enabled parts of the UI with IfLoggedIn component
What considerations about permissions should I consider? Who can create or edit a bug?
I found a component that checks if the user is logged in to decide if comments can be added or not. About what kind of login are We talking about? 🤔

**MichaelMure commented:**
At the moment, anyone can create and edit any bugs (including comments they are not the author of). This should change in the future once the identity system become smarter. For now, maybe it's best to match whatever is possible in git-bug core?

Regarding the authentication, there is three different scenario:
1. the user might might not have an identity yet (that is, Repository.userIdentity can be nil in GraphQL). In that case we want to have a read-only UI if we don't have a user yet.
2. in the future, the webUI might be used for a public portal (possibly with external auth like github OAuth where identities would be imported in git-bug at first login). Again, we want a read-only view when we don't have a user logged in yet.
3. the webUI can be started with --read-only to force a read-only view.

So in the react code, the write-enabled part of the UI are guarded with the IfLoggedIn component.

claudioantonio:
fix #543

### DIFF
--- a/webui/Readme.md
+++ b/webui/Readme.md
@@ -4,23 +4,23 @@
 
 ### Run GraphQL backend
 
-1. Download a git-bug stable binary.
+1. Download a git-bug stable binary or compile your own by running `make` in the **root** directory:
 
-2. Execute git-bug binary inside directory for the git repository it will manage issues:
-   - git-bug webui -p 3001
+2. Run the git-bug binary inside your git repository. It will manage issues and start the API:
+   - `git-bug webui -p 3001`
 
 ### Run ReactJS front-end
 
-1. Clone git-bug repository.
+1. If you haven't already, clone the git-bug repository:
 
-2. Enter webui directory and install libraries needed:
-   - npm install
+2. Enter the `webui` directory and install the needed libraries:
+   - `make install` or `npm install`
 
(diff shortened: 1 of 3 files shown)


webui/Readme.md
```diff
@@ -2,13 +2,25 @@
...
+1. Clone git-bug repository.
+
+2. Enter webui directory and install libraries needed:
+   - npm install
+
+3. Generate ts code from graphql files and run webui in development mode
+   - npm start
+   - If You got compilation errors from lint, run lint command below and start again:
+      - npm run lint -- --fix
+      - npm start
```
MichaelMure, line 23:
```suggestion
   - `make start` or `npm start`
   - If you get some lint errors, run the lint command below and start again:
      - `make fix-lint` or `npm run lint -- --fix`
      - `make start` or `npm start`
```

MichaelMure (commented):
I fixed some english issues and added more informations.

MichaelMure:
:rocket:

claudioantonio:
I think It is much better! Thanks!

Issue #366, opened 2020-04-10
Multiple bugs with gitlab bridge.


dtrckd:
Hi,

Playing with git-bug on a existing gitlab repository with a few existing issues on the vendor's bug tracker,  I ran trough the following errors:

After having successfully configured the bridge I could import the gitlab issues as well as creating a couple of local issues with `git bug add`, so far so good. Then, I meet the following errors:
* when doing again `git bug bridge pull`, issues are duplicated with new hashes. 
* nothing is exported when doing `git bug bridge push`, it jus returns `exported 0 issues with myBridge bridge`.

I've tested the bridge on a freshly created repo on gitlab, again the first steps goes well, but this time when doing `git bug bridge pull` a second time to test if duplicates arise, I got this crash:

```
panic: runtime error: index out of range [-1]                                                                                                        
                                                                        
goroutine 51 [running]:       
github.com/MichaelMure/git-bug/bridge/gitlab/iterator.(*issueIterator).Value(...)
        /home/dtrckd/src/_sink/git-bug/bridge/gitlab/iterator/issue.go:38
github.com/MichaelMure/git-bug/bridge/gitlab/iterator.(*Iterator).NextIssue(0xc0010ba000, 0xc000043dd0)
        /home/dtrckd/src/_sink/git-bug/bridge/gitlab/iterator/iterator.go:86 +0x244
github.com/MichaelMure/git-bug/bridge/gitlab.(*gitlabImporter).ImportAll.func1(0xc000260100, 0xc0000aed00, 0xc0010bc000)
        /home/dtrckd/src/_sink/git-bug/bridge/gitlab/import.go:70 +0x547
created by github.com/MichaelMure/git-bug/bridge/gitlab.(*gitlabImporter).ImportAll
        /home/dtrckd/src/_sink/git-bug/bridge/gitlab/import.go:66 +0x146

```
Similar scenario on Github seems to work fine.

Version:
git-bug version: 0.7.1-dev-28c3ee3b71
github.com/xanzy/go-gitlab v0.29.0
go version go1.14.1 linux/amd64

Gitlab: 12.9.2 (self hosted)
OS: Linux 4.19.0-8-amd64 / Debian 10 x86_64

dtrckd:
New test result: 
When using git-bug bridge by pushing bug to remote first, it works. Then If we create an issue on Gitlab and import it, the same scenario as before, firts time ok, then I got the crash (+ the doublon), and bridge push doesn't push new local bug anymore.

[...]

PR #560, opened 2021-02-09
Add state changes to gitlab importer

5nord:
This PR adds the missing bug state history by using the new [GitLab resource state events API](https://docs.gitlab.com/ee/api/resource_state_events.html).

Related to #366.

### DIFF
--- a/bridge/gitlab/import.go
+++ b/bridge/gitlab/import.go
@@ -98,6 +98,16 @@ func (gi *gitlabImporter) ImportAll(ctx context.Context, repo *cache.RepoCache, 
 				}
 			}
 
+			// Loop over all status events
+			for gi.iterator.NextStateEvent() {
+				statusEvent := gi.iterator.StateEventValue()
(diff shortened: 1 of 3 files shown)


MichaelMure:
This looks good but I see 2 problems:
- state change events and other events should be interleaved and sorted based on the date because in git-bug, the date matter way less than the actual order of changes. In the timeline, you will see the events in the order they are created in git-bug. As Gitlab seems to have spread the events in multiple endpoints, that means that the iterator will need to query all of them, sort and merged them into a single stream for the importer to insert them in the correct order.
- the previous state change should probably be removed

5nord:
Yes, I was hoping the bridge-core sort events according to timestamp. It was worth a try ;)

I also think a sorting iterator is required. This additional layer could also be useful for future GitLab-API changes. I'd try that next week, if you don't mind.

MichaelMure:
> I'd try that next week, if you don't mind.

Yes please!

Issue #567, opened 2021-02-12
Search by arbitrary metadata


vmiklos:
In case a bug is imported (e.g. from github), then git-bug knows that url, e.g.:

```
$ git show refs/bugs/016b3f0c56c1724c0230919ce3218b5cfa19f8ce:root|python3 -m json.tool
{
    "version": 2,
    "ops": [
        {
            "type": 1,
            "author": {
                "id": "e256bc3329691afb0ed3f7d3b9675ef8cecd1a16"
            },
            "timestamp": 1583555863,
            "metadata": {
                "github-id": "MDU6SXNzdWU1NzcyNzY4MTY=",
                "github-url": "https://github.com/author/myproject/issues/6",
                "origin": "github"
            },
```

It would be nice if e.g. git bug ls could search in this metadata. So let's say I'm offline and I would like to look up https://github.com/author/myproject/issues/6, then that would be possible. Perhaps something like:

```
git bug ls --metadata github-url=https://github.com/author/myproject/issues/6
```

could list all matching bugs, typically just one.

vmiklos:
@MichaelMure what do you think? I don't mind taking a look at this myself, just wanted to do a sanity-check on the idea in general and on the syntax in particular before I invest time in coding this. Thanks.

MichaelMure:
I guess that make sense. The syntax is a bit problematic though (both for the query language and the CLI argument for `ls`): you need to specify both key and value, which is something both of them don't do yet.

To be able to search like that you would need to:
1. ~add the metadata in the index (see `BugExcerpt`)~ actually it's already there, `CreateMetadata` in `BugExcerpt`
2. add a new keyword in the query language (`/query`), find a way to deal with the key/value problem
3. implement the search (`/cache/filter.go`)
4. connect all that stuff in the `ls` command, deal with the key/value problem

vmiklos:
Thanks! I have a PoC (without any error handling) based on the above, will submit a PR when I don't see obvious problems with it anymore.

MichaelMure:
You can also open a draft PR for some early feedback (if I find the time of course), as you prefer.

PR #568, opened 2021-02-14
Add ability to search by arbitrary metadata

vmiklos:
Example:

~/git/git-bug/git-bug ls --metadata github-url=https://github.com/author/myproject/issues/42

Fixes <https://github.com/MichaelMure/git-bug/issues/567>.


cache/filter.go
```diff
@@ -37,6 +37,21 @@ func AuthorFilter(query string) Filter {
 		return author.Match(query)
 	}
 }
+// MetadataFilter return a Filter that match a bug metadata
```
MichaelMure, line 40: Precise that this only work with metadata **at creation time**. It will not match metadata added later in the lifetime of the bug.

cache/filter.go
```diff
@@ -37,6 +37,21 @@ func AuthorFilter(query string) Filter {
...
+func MetadataFilter(query string) Filter {
+	return func(excerpt *BugExcerpt, resolver resolver) bool {
+		tokens := strings.Split(query, "=")
```
MichaelMure, line 43: At this point of the code, you should already have your string processed into a key and value. Splitting around `=` is a matter for the CLI layer.

commands/ls.go
```diff
@@ -65,6 +66,8 @@ git bug ls status:open --by creation "foo bar" baz
 		"Filter by status. Valid values are [open,closed]")
 	flags.StringSliceVarP(&options.authorQuery, "author", "a", nil,
 		"Filter by author")
+	flags.StringSliceVarP(&options.metadataQuery, "metadata", "m", nil,
```
MichaelMure, line 69: an example might be useful as you expect a specific format

MichaelMure (commented):
Looks like it's headed the right way (:+1:) but you are missing the query language part for now.

cache/filter.go
vmiklos: Indeed, I've updated the comment to state that explicitly. My motivation is to be able to look up bug URLs offline, so that use-case works when metadata is limited to metadata at creation time.

cache/filter.go
vmiklos: Aha, and that means I can even do better error handling. :-)

commands/ls.go
vmiklos: I've added one, though I don't want to spell out a complete GH issue; hopefully the "URL" form is clear enough.

vmiklos:
> Looks like it's headed the right way (+1) but you are missing the query language part for now.

Right, that's only needed for the web part, correct? Unless you insist, I would be happy to leave that out from the current PR. I've updated the commit message to say it handles the cmdline piece.

Thanks.

MichaelMure:
> Right, that's only needed for the web part, correct? Unless you insist, I would be happy to leave that out from the current PR. I've updated the commit message to say it handles the cmdline piece.

No, it's also possible to use it with the CLI and the termui. I really don't want to have some features of the query language only supported in specific condition, so we should achieve parity.

vmiklos:
````
git-bug ls metadata:github-url:https://github.com/author/myproject/issues/42
````

works for me now, could you please take a look? Thanks.

The commit message is unchanged, as the query language is working from the cmdline, but I didn't do anything for the web UI explicitly.

**Comment by vmiklos** [comment]
@MichaelMure can I help you in some way to get this reviewed or should I just wait more? (I can wait, I'm just a bit confused if you expect something from me or not). Thanks.

File: query/lexer.go [hunk 4]
```diff
@@ -58,6 +69,20 @@ func tokenize(query string) ([]token, error) {
 			continue
 		}
 
+		if len(split) >= 2 && split[0] == "metadata" {
```
MichaelMure, line 72: The lexer should not be aware of a particular keyword meaning. Your extension to the query language means that the lexer should now be able to read tokens with one key and two values (that you called qualifier and subqualifier: I guess that make sense, I wonder if there is already an established term for that?). What those token means is the responsibility of the parser (the next step).

You likely noticed `newTokenSearch` which already has some semantic notion. This is not great, but let's try to not follow on this path. Ideally the lexer should only care about breaking a string into tokens.

Could you adjust your code with this optic?

MichaelMure: Maybe we need a `newTokenKVV` or something like that?

MichaelMure:
Sorry, things takes time sometimes :)

Don't worry, your contribution is not lost.

query/lexer.go
vmiklos: Yes, that makes sense. I did both.

query/lexer.go
```diff
@@ -58,18 +71,19 @@ func tokenize(query string) ([]token, error) {
...
 		}
 		if len(split[1]) == 0 {
 			return nil, fmt.Errorf("empty value for qualifier \"%s\"", split[0])
 		}
 
-		tokens = append(tokens, newTokenKV(split[0], removeQuote(split[1])))
+		if len(split) == 2 {
+			tokens = append(tokens, newTokenKV(split[0], removeQuote(split[1])))
+		} else {
+			rest := strings.Join(split[2:], ":")
```
MichaelMure, line 84: I'm not sure this should really accept more than two `:` in a token.

MichaelMure (commented):
Alright, I think the last step is going to be completing the lexer/parser tests with those new cases.

query/lexer.go
vmiklos: Do you have an alternative syntax in mind? I currently propose `metadata:github-url:https://github.com/author/myproject/issues/42`, but that means there will be 3 colons, even if the third is not interesting. I.e. if you ban > 2 colons, then colons are not allowed in metadata, which means URLs are not allowed in metadata, which would be sad.

Once we settle down on the syntax, I can take a look at the tests. :-) Thanks.

MichaelMure: Hu, good point. Man, those formal grammar lessons that I barely listen to are quite far now ;)

After some idea shuffling, I'm thinking that the good solution might be to use quotes. It's already possible to do that: `author:"René Descartes"`. So in your case it would be `metadata:github-url:"https://github.com/author/myproject/issues/42"`.

Also, the command line parse behave in a similar way, no? Would `git-bug ls --metadata github-url="https://github.com/author/myproject/issues/42"` be accepted? Not that it would be required as long as the shell is happy, but some uniformity would be nice.

What do you think?

vmiklos: I like your proposal, and it may work for the webui, but the trouble is that the shell eats these quotes. So if I say meta:gh:"foo", I only get meta:gh:foo in go. You may require the user to type meta:gh:\"foo\", but that would be terrible usability.

Considering these, I would argue that either the current syntax is not that bad after all, or if you really would like some different syntax, then perhaps metadata:github-url=..., so it would be more obvious that the remaining characters are a value.

Which one would do you prefer?

MichaelMure: > So if I say meta:gh:"foo", I only get meta:gh:foo in go.

Is that really the shell or [cobra](https://github.com/spf13/cobra)? What do you get with `meta:gh:"foo bar"`?

--> I tested that, it's indeed go, **but**, go only unquote that and we get the args separated properly.

--> Actually, I had this problem before and to deal with it the `ls` command put the quotes back: https://github.com/MichaelMure/git-bug/blob/master/commands/ls.go#L93-L98

So, some shenanigan happen in the process but from the CLI side and parser side, it looks and behave as we expect it to.

> then perhaps metadata:github-url=...

That could be an alternative way. The grammar would be only 1-tuples (for the full text search) and 2-tuples for the filters/sorting. No 3-tuples.

In the later case, we would still need the quotes IMHO, to make the grammar a bit more regular. Having `:` suddenly not being a separator because we already found two doesn't sounds like a great idea. What if we have 4 parameters next? It's unlikely but it would break the previously valid queries.

Considering this, this later syntax doesn't have much interest, no? What do you think?

vmiklos: Right. :-) One more thing I forgot is that the cmdline can always use `ls --metadata github-url=https://github.com/author/myproject/issues/42`, so using the query language is not a must. So in case we go with the `metadata:github-url:"https://github.com/author/myproject/issues/42"`, that would be fine: it would be a good syntax (later) for the webui, and the cmdline can use `metadata:github-url:\"https://github.com/author/myproject/issues/42\"` if it really wants to use the query language for metadata.

Is this OK? If so, then I could have a go at updating the code to do this.

vmiklos:
I've now implemented the foo:bar:"https://www.example.com/" syntax idea, also added tests. Please let me know if anything else is missing.

vmiklos:
@MichaelMure can I help with anything to get this reviewed? Thanks.

MichaelMure:
Nothing to do on your side, it's just that real life and day job don't give me much spare time.

query/lexer.go
```diff
@@ -50,26 +63,55 @@ func tokenize(query string) ([]token, error) {
 
 	var tokens []token
 	for _, field := range fields {
-		split := strings.Split(field, ":")
+		// Split using ':' as separator, but separators inside '"' don't count.
+		quoted := false
+		split := strings.FieldsFunc(field, func(r rune) bool {
```
MichaelMure, line 68: I had this exact function to handle the quotes at some point, before writing this more complex/robust lexer. I don't remember exactly why, but IIRC this left a few edges cases open (maybe the unmatched quote?). I'll dig a bit to see if there is a better way to do that. Otherwise, this can be merged, because, well, it works.

MichaelMure:
I ended up doing some refactoring/code golf to use the same split function to split on spaces and on `:`. I believe this cover more edge cases, and it's also simpler.

I think it's good to go now! Thanks for the help :)

vmiklos:
Thanks for reviewing this! :-)

Issue #566, opened 2021-02-12
WebUI - Add info in README that Go is necessary to be installed on dev machine to bundle a new release for WebUI into git-bug binary


claudioantonio:

MichaelMure:
Github wasn't smart enough to understand "fixes #..."

Issue #554, opened 2021-02-05
WebUI - Bundle the compiled js into the go binary


claudioantonio:
After You merge #548 can We bundle a new WebUI release into the go binary?

MichaelMure:
Sure thing but I'm not sure there is a point without also making a release (bug git-bug is not in a release state at the moment).
If you want to do that, you can run `make pack-webui` from the root and make a PR with the result.

claudioantonio:
Based on the article below I tried to pack webui but have no success... :(
https://dzone.com/articles/makefile-command-in-windows

Error messages

PS D:\dev\goodfirstissues\git-bug-2> make pack-webui
**process_begin: CreateProcess(NULL, uname -s, ...) failed.**
npm run --prefix webui build

> webui@0.1.0 build D:\dev\goodfirstissues\git-bug-2\webui
> npm run generate && react-scripts build


> webui@0.1.0 generate D:\dev\goodfirstissues\git-bug-2\webui
> graphql-codegen

  √ Parse configuration
  √ Generate outputs
Creating an optimized production build...
Compiled successfully.

File sizes after gzip:

  229.3 KB (-64 B)  build\static\js\2.2013acb7.chunk.js
  9.15 KB           build\static\js\main.c1c3dd52.chunk.js
  768 B             build\static\js\runtime-main.c79097da.js

The project was built assuming it is hosted at /.
You can control this with the homepage field in your package.json.

The build folder is ready to be deployed.
You may serve it with a static server:

  npm install -g serve
  serve -s build

Find out more about deployment here:

  https://cra.link/deployment

go run webui/pack_webui.go
**process_begin: CreateProcess(NULL, go run webui/pack_webui.go, ...) failed.
make (e=2): The system cannot find the file specified.
make: *** [pack-webui] Error 2**


Do You have any idea about that?

claudioantonio:
When I run make into webui folder the message is:
**PS D:\dev\goodfirstissues\git-bug-2\webui> make pack-webui                                                              make: *** No rule to make target `pack-webui'.  Stop.**

claudioantonio:
Makefile into webui folder really does not have this target pack-webui

start:
	npm start

install:
	npm install

test:
	npm run generate
	npm run lint
	CI=true npm run test

build:
	npm run build

lint:
	npm run lint

fix-lint:
	npm run lint -- --fix

claudioantonio:
[...]

PR #569, opened 2021-02-15
Commits for issues #554, #564 and #566

### LINKED ISSUE #564: WebUI - add prerequisites to README
A proposal bringing some colors. ;-)

image

claudioantonio:
fixes #564 #566 #554


MichaelMure (changes requested):
Looks good but you put the files in `/public`, which means they get packaged into the final binary (you can see them in the `packed_assets.go` file). I'm not exactly sure how the javascript build process work, but that shouldn't happen as they are just there for the readme.

claudioantonio:
What do You think about a new folder called webui/ReadmeImages, or ReadmeImgs?
It will not be considered when creating build bundle and let clear that those images are related to the Readme file.
It´s a like when You save a full website and the browser creates a folder for the site´s images.

MichaelMure:
That could work, but I'm starting to think that it's not worth it just to have some pretty images in a developer readme. Licensing could also be a problem as those images are covered by their own copyright. To have them in this repo we would need to explicitly mention exception to the general GPL licensing...

claudioantonio:
No problem, @MichaelMure ! I'll remove the images from Readme. ;-)

claudioantonio:
Pull request ready to go!

MichaelMure:
:+1: :+1:

Issue #435, opened 2020-07-25
[gitlab] Status change are not imported anymore


MichaelMure:
I've been investigating the CI break on master. It seems that the gitlab bridge is now unable to import the status changes of an issue (open/close).

This is not due to a broken code in git-bug, it's due to a gitlab API change. It seems that the "Notes" API is not listing those events anymore. More debugging is required to see if this information appears in another way. If not, we'll have to checks the issue status at the time of the import and create a status change if necessary. Very less than ideal.

I won't rant to much here but there is reasons I should (do they even have a changelog??).

MichaelMure:
Confirmed that the notes just don't have this event anymore. @A-Hilaly I'm sad now :|

This is not a nice API.

MichaelMure:
Closing as this is likely fixed.

PR #571, opened 2021-02-15
[gitlab] Add new iterator with state change events

5nord:
GitLab issues events are spread across various GitLab API calls. This causes various problems for gitlab bridge. 

This PR replaces the previous iterator with a channel/Go-routine based implementation. The new implementation is easier to adapt, provides the long-lost issue state changes ( #435, #560) and resolves various importing and exporting issues (#366).

Kindly note this PR is still a work in progress and it's not functioning correctly, yet.


5nord:
I think this PR is not ready to merge yet, because three issues are still observable:
* changed description events seem to be missing
* calling `git bug bridge pull` twice tails with an commit-error, when upstream has changed.
* `git bug bridge push` still not working.

I indent to continue in a couple of weeks, when I have more spare time available.

MichaelMure:
@5nord I merged in master https://github.com/MichaelMure/git-bug/pull/570 which update go-gitlab with your changes.

MichaelMure:
Hey @5nord, git-bug is finally converging towards a stable version. It'd be great to have your changes as well!

Don't worry there is no rush either, it's going to take at the very least a few weeks.

5nord:
Hey Michael,

I rebased and aligned my changes with your latest master. I think, it's mergeable, despite its minor issues:

* all edit events (except for title change) are missing. But this is probably due to git-bug architecture?
* changed description event is missing. I cannot explain why, do you have an idea?
* local changes are not pushed by `git bridge push`. I suggest investigating and fixing this issue in another PR.

I would be really great, if someone with more experience in git-but could verify those issues are acceptable.
BTW I used [this issue](https://gitlab.com/5nord/test01/-/issues/3) to verify the PR work sufficiently.

MichaelMure:
Hey @5nord, I'll need your help to understand:

>     * all edit events (except for title change) are missing. But this is probably due to git-bug architecture?

What sort of edit events are you talking about? Events from gitlab? git-bug support title change, comment edition ... so that should just work. Maybe have a look at the github bridge for inspiration (it recently got a nice upgrade by @rng-dynamics to make it more robust) ?

>     * changed description event is missing. I cannot explain why, do you have an idea?

Also not sure which event you are talking about.

>     * local changes are not pushed by `git bridge push`. I suggest investigating and fixing this issue in another PR.

So the exporter is not working? What I don't get is that there is a test that make sure this keep working.

Speaking of tests, are we able to re-enable https://github.com/MichaelMure/git-bug/blob/master/bridge/gitlab/export_test.go#L245-L248 with your changes?

5nord:
I'll explain remaining issues with some screenshots:

gitlab-issue-history

On the left side of the picture you'll see a detailed view of my [gitlab test issue](https://gitlab.com/5nord/test01/-/issues/3). On the right side you'll see the same issue/bug viewed from `git bug tui`

> edit events (except for title change) are missing

I edited comment 1. A note, that the comment has been edited is displayed in gitlab, but not in `tui`.

>  changed description event is missing

Also edited the description. And there's [code to import](https://github.com/5nord/git-bug/blob/add-new-gitlab-iterator/bridge/gitlab/import.go#L197-L218) this to git-bug. But it seems to have no effect.

> So the exporter is not working? What I don't get is that there is a test that make sure this keep working.

It seems this way.  
I focused on fixing the import, I just noticed by accident that expert does not work for me.

6543:
what's the current status?

GlancingMind:
Related Issue is: https://github.com/MichaelMure/git-bug/issues/653

Some vague and short description:
The edit operations work, but the UI's currently don't display them. This is due to, that Comments and CommentEdits use a new ID scheme (CommentID := BugID interleaved with CommentID). The problem is, that currently the UIs still use the old IDs and by doing so, all Comment look to them as unedited Comments.

5nord:
> what's the current status?

This PR fixes the pulling issues from Gitlab and fixes the lost state-change-events.

I believe this PR does not introduce any new issues. But I am still waiting for review from someone with more experience in this code base.

Note, this PR does not fix all Gitlab-bridge issues. For example, exporting bugs to Gitlab seems unreliable still.

MichaelMure:
@5nord let's keep that open if it's still relevant? I promise this project is not dead :-S

bridge/gitlab/import.go
```diff
@@ -59,42 +55,30 @@ func (gi *gitlabImporter) Init(_ context.Context, repo *cache.RepoCache, conf co
...
 			}
 
-			// Loop over all notes
-			for gi.iterator.NextNote() {
-				note := gi.iterator.NoteValue()
-				if err := gi.ensureNote(repo, b, note); err != nil {
-					err := fmt.Errorf("note creation: %v", err)
-					out <- core.NewImportError(err, entity.Id(strconv.Itoa(note.ID)))
-					return
+			for _, e := range SortedEvents(IssueEvents(ctx, gi.client, issue)) {
```
MichaelMure, line 74: If I understand that right, this force pulling all the events (in memory on top of that) to sort them, and only actually import after that. Is there not a way to pull them already ordered?
If the problem is merging the 3 sources of event together, this could be done "on the fly" by picking the earliest of the 3 head, assuming each of those source are ordered already.

MichaelMure (commented):
Thanks, that looks way simpler and more robust.
Before merging, I have a question about the event ordering.

bridge/gitlab/import.go
5nord: You are right! I would make git-bug feel snapier, I like it. However I am not sure when I am able to give it a try.

MichaelMure: I went ahead and implemented that. From a code perspective that works, but I have no idea it that actually works with gitlab. I guess the CI will tell.

MichaelMure: Ho well, the CI will not tell because this PR is in your repo. It has to be pushed in mine (which you will be able to do if you accept the invite) to have the proper tokens. In the mean time, I'll push that myself into a new PR.

MichaelMure:
I also re-enabled the test marked as broken before: https://github.com/MichaelMure/git-bug/pull/571/commits/e888391b36307d0d4a1e12ba3a57b602c8b1528a

MichaelMure:
Well, everything looks great, what do you think?

5nord:
Am I supposed to merge? I don't know project policy and I don't want to be rude :)

MichaelMure:
I just wanted to know your opinion about my changes. Let's merge!
Very nice work, thank you :)

PR #576, opened 2021-02-27
Add container setup

GlancingMind:
This PR will introduce a container setup to start the (dev)-WebUI without the need of installing git-bug, go or node.
The only requirement is docker and docker-compose or some other compatible tool - like podman.
To use it, check out the updated WebUI-Readme.

- It should work flawlessly on Linux.
- Windows should work, too. Though **it is highly recommended to use docker-compose via WSL**.  As one service will copy the node_modules from the container to the host, which might be very slow. Additionally this  node_modules might contain for Windows incompatible Linux binaries. 
- Mac and Bsd wasn't tested.

## Some considerations or required changes before merging:
- [ ] I introduced a .gitattributes and .editorconfig file to normalize linefeeds to unix format. I'm not sure, if this might break some users diffs, if they are on Windows. I recommend to clone the repository again to make sure, that .gitattributes changes are applied correctly. Additional care is required with binaries. If git cannot correctly detect a file as binary, linefeeds might be wrongly inserted, leading to corruption. I recommend to explicitly declare binary files as binaries in the .gitattributes file.
- [ ] Current a [docker image](https://hub.docker.com/r/glancingmind/git-bug) is build automatically **out of my fork**. I recommend to create an official account on dockerhub, which should host the official container images. Be aware, that Dockerhub will use the ReadMe of the repository when automatic builds are used. Which contains to much unnecessary information as a container description. This could be resolved, by building the container image with a github workflow and then push the image with the dedicated [ReadMe to dockerhub](https://github.com/christian-korneck/docker-pushrm). Or the Dockerfile is hosted with it's dedicated ReadMe in a separate repository.
- [ ] Currently the ReadMe and the Dockerfile contains links to my fork, this links should be changed to the official Repository after the images are officially hosted somewhere.


Dockerfile
```diff
@@ -0,0 +1,83 @@
...
+
+# NOTE: Git-Bug currently only supports the interactive creation of an user.
+# To still create a user, the username, email and avatar is piped to the
+# git-bug process.
+# As git-bug input-prompt reads a whole 4096 byte buffer at ones, every value
+# must be to the length. Otherwise the first prompt (username) will eat up all
+# other given values, which will lead to an EOF fpor the second prompt (email).
+# See the answere to the question: reading-input-from-stdin-in-golang
+# over at https://stackoverflow.com/a/45914294
+RUN printf "%-4096.4096s\n%-4096.4096s\n%-4096.4096s\n" \
```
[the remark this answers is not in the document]
MichaelMure: https://github.com/MichaelMure/git-bug/issues/577

.editorconfig
```diff
@@ -0,0 +1,25 @@
+# This file defines a common editor configuration for a consistent code base.
```
MichaelMure, line 1: The go code formatting is done with `go fmt` so adding a default formatting might just create a conflict.
For the js code there is prettier which I think does the same thing?
The remaining is not going to be very useful as most of the editor you can use already use sensible default values.

commands/webui.go
```diff
@@ -54,6 +57,7 @@ Available git config:
 	flags := cmd.Flags()
 	flags.SortFlags = false
 
+	flags.StringVar(&options.host, "host", "127.0.0.1", "Network address or hostname to listen to (default 127.0.0.1)")
```
MichaelMure, line 60: This is nice. Could you have it in a separate PR?

MichaelMure (commented):
hey @GlancingMind

I'm very sorry about this but I'll have to say no about most of those changes. Hopefully you can understand why.

I can see how a docker setup could help you and your teammates, but as you have seen this require a quite complicated setup to be useful for the webUI development. It's not a bad thing but it's not something I'd like to keep maintaining during the lifetime of this project.

See also my other comments inline.

.gitattributes
```diff
@@ -0,0 +1,8 @@
+# Force linux line ending for all as "text" file recognized files.
```
MichaelMure, line 1: This might be useful, but before introducing such a thing I'd prefer to actually have a problem to fix. As far as I know this fine very rarely exist in open source project, which let me thing that git must have sensible default.

commands/webui.go
GlancingMind: Sure. :)

.gitattributes
GlancingMind: I understand. To be honest, I only introduced this after I noticed problems with the container setup on windows.
Prettier complained about wrong line endings. Forcing a common linefeed wasn't my preferred solution, but nothing else seemed to work properly.

GlancingMind:
Well that's unfortunate. Maybe my solution is a bit over-engineered, too. :-)
Nonetheless I can relate to what an overhead of supporting another "environment" might add.

commands/webui.go
GlancingMind: Here it is: https://github.com/MichaelMure/git-bug/pull/583

PR #584, opened 2021-02-28
Add option to specify host address

MichaelMure:
'--host'-cmdline-option is added to the webui command.

Previously, the WebUI couldn't be hosted inside of a container. As the
WebUI-server only listend per default to localhost and there was no
option to change the address, the server should listend to. This means,
that the WebUI was only reachable from localhost. So only from inside of
the container but never from outside.
The '--host'-option allows to set the IP address or a hostname
which the WebUI-server should listen to. E.g. by setting 0.0.0.0 or ::
as address.

Update documentation for new option.

Update shell completion for new option.

Compilation seems to add another go-gitlab version.


MichaelMure:
Merged! Thanks @GlancingMind !

PR #599, opened 2021-03-14
WebUI: Fix hover-color of green buttons and caption

GlancingMind:
Why were the buttons captions changed, too? 
 See Michaels reply: https://github.com/GlancingMind/git-bug/pull/31#issuecomment-798921866


MichaelMure:
:rocket:

PR #600, opened 2021-03-15
WebUI: Improve navigation

GlancingMind:
Changes:

- Add 404 Page for invalid URLs
- - URLs with invalid BugID will also redirect to 404 page
- Add a BackToList button on "NewBugPage" and "BugDetailPage"
- The "NewBugPage" will redirect to the BugsDetailPage after create of a bug


return-404-page.patch
```diff
@@ -0,0 +1,37 @@
+diff --git a/webui/src/App.tsx b/webui/src/App.tsx
```
MichaelMure, line 1: I just had a quick look, but this should not be here.

GlancingMind: Oh totaly missed that one. Removed it.

MichaelMure:
So this is nice but I'm not 100% about the "back to list button".

My first complain is that it breaks the HTML layout. As you can see below, it removes width from the main container, so if the button is translated someday, it can seriously break the layout. Ideally, the width of the button would not change the layout.

image

Maybe just having the arrow would be enough? Also, the string color make it look like a primary button while it's only there as a secondary option in the navigation.

Now, ideally this button would not exist to begin with. If you look at Github, there is no button like that and the (secondary) navigation is done with the fixed header:
image

We don't have yet enough categories for that, but we will likely have it at some point (at least when we have pull-requests, a code viewer would make sense as well). Maybe we should take the leap already even if that means that we either have only one tab or we have more but disabled?

GlancingMind:
> My first complain is that it breaks the HTML layout. As you can see below, it removes width from the main container, so if the button is translated someday, it can seriously break the layout. Ideally, the width of the button would not change the layout.

I agree, didn't think about that one.

> Maybe just having the arrow would be enough? Also, the string color make it look like a primary button while it's only there as a secondary option in the navigation.

Only showing an arrow would be fine too. Regarding the color choice. It was intentionally made in the same color as the Navbar/AppBar to convey a navigational meaning.

> We don't have yet enough categories for that, but we will likely have it at some point (at least when we have pull-requests, a code viewer would make sense as well).

Yeah, the "back to list"-button is compromise for the above mentioned reason.

GlancingMind:
@MichaelMure Would rather like a navigation similar to GitHub (always visible navigation bar) or more like GitLab with a collapsible side navigation.
E.g. from the Material UI demos:
[Github like](https://codesandbox.io/s/ubt9g)
[GitLab like](https://codesandbox.io/s/h0l7d)

PS: I think the collapsible one might not work with the git-bug logo.

MichaelMure:
Both is fine to me, but I agree that the materialUI collapside one is not great. 

It can be good though:

image image

The horizontal tab one is OK-ish but not so great either. Maybe it should be done without pre-made MaterialUI components?

GlancingMind:
Here is a mock up of the top-Navigation.

With full width.
image
Without full width.
image

I'm currently not aware of other ways of navigation. Besides of the sidebar, which then would look like your posted pictures.

MichaelMure:
This look quite nice!

I'd say:
- not full width (it won't work well with very wide screen)
- don't include projects and wiki, it's really too far in the future

On the disabled tabs, could you have a tooltip like "This feature doesn't exist yet, come help us build it." ?

GlancingMind:
> - maybe centered?
>
> On the disabled tabs, could you have a tooltip like "This feature doesn't exist yet, come help us build it." ?

Should be possible. I'll have a look.

GlancingMind:
Done. Took me way to long due an React error but I got it!
Unfortunatly I could not create a screenshot with a displayed tooltip.

image

Notice the bug navigation item will only be highlighted on the BugListPage.
image

This is how it looks like when visiting "/code" from the URL (which is not implemented).
image

GlancingMind:
Btw. the eslint errors just don't won't to go away. If I fix them locally, they break on the pipeline.
Fixing the pipeline, will fail locally.

MichaelMure:
I applied the lint command (+ `--lint`) from `package.json` which fixed the error locally, and the CI seems happy now.

MichaelMure:
This looks great, I'll try it when I get a minute and we can probably merge.

MichaelMure:
Alright, I fixed two issues:
- I used a `Link` component as base for the tabs to redirect within the SPA instead of reloading the whole page
- two leftover `console.log`

MichaelMure:
Nice work :)

PR #602, opened 2021-03-19
Edit comments and inspect edit history

GlancingMind:
It is now possible to edit a comment and inspect its previous edit history.


MichaelMure:
Awesome!

The only problem I could find is the first version missing in the history:
image

To reproduce:
- add a new comment
- edit it
- inspect the history

GlancingMind:
Hmm for me it seems to work fine. Did you miss to click on the "#2 Edit"?
(The edits are accordion elements :-) )

Maybe I should change the caption for the first comment to be "Initial Mesage" instead of "#2 Edit".
Also I should add an unfold-indicator to the elements.

GlancingMind:
Done.
- Add Icon to indicate fold status
- Change order of edit number. The first edit should have "#1"
- Add text for first issue.
I think this might reduce confusion.
Per default, the most recent edit is shown.
image

On click of another edit "tab". The state of the message at this point in time is shown.
image

MichaelMure:
Ha yeah, I completely missed the fold behavior. The icon for the fold status does help but I wonder if having all the sections when the popup open would help by forcing the user to click on something?

Regarding the messages, what about simplifying?
- `(most recent edit)` --> `(most recent)`
- `(initial description)` --> dropped entirely

GlancingMind:
> I wonder if having all the sections when the popup open would help by forcing the user to click on something?

I don't understand. Do you mean, that all Edits should be initially in a folded state?

> (most recent edit) --> (most recent)

Sure that is easy.

> (initial description) --> dropped entirely

Well I added it after i noticed that the first "edit" isn't an edit but the initial description. 
Not that a user things the initial description is left out of the history and only the following changes are shown.

MichaelMure:
> I don't understand. Do you mean, that all Edits should be initially in a folded state?

Yeah, I'm considering that, but I'm not sure that the right way. My thoughts in favor of that are:
- the user is not especially interested to see the *last* version, as it's the same as the one already displayed in the normal UI
- he's also not necessarily interested by the *first* version. At least there is not much reason to show that one especially.
- if everything is folded at first, it force the user to learn to click on the elements, and thus learn that it's foldable sections

On the minus side, it means always an extra click, but Github seems happy with that so ...

> Well I added it after i noticed that the first "edit" isn't an edit but the initial description.
> Not that a user things the initial description is left out of the history and only the following changes are shown.

The confusion start when you have "most recent edit" instead of "most recent", because you named those things "edit". The first version doesn't fit anymore as it's not an edit. If you just have "most recent", you don't have that confusion. What's implied is "version" instead of "edit". And that's enough information IMHO.

GlancingMind:
I changed some part of it. The initial description is no longer treated as edit.
Means the initial description will now be displayed as "Created on" and the title will only count the real edits.

Using "#x Version on <Date>" seems to be elegant but their is one drawback that I could think of.
If markdown might be supported, the history could show a diff to the previous edit. 
But when "Version" is used as description the expectation could be to see the is-state, instead of the changes to the previous one state.

image

MichaelMure:
That sounds like a good compromise. I'll give it a try and we can merge.

GlancingMind:
Oh yeah, if I should squash some commits let me know.
I just don't want to force push on another repo without notice. :-)

GlancingMind:
Ops, totally forgot to push the commit for the above mentioned changes.

MichaelMure:
LGTM!

> Oh btw, if I should squash some commits let me know.
> I just don't want to force push on another repo without notice. :-)

Squash as much as you want in your branch.

PR #603, opened 2021-03-19
Display count of comment for each bug on bug list

GlancingMind:
image

### DIFF
--- a/webui/src/pages/list/BugRow.graphql
+++ b/webui/src/pages/list/BugRow.graphql
@@ -9,5 +9,8 @@ fragment BugRow on Bug {
   labels {
     ...Label
   }
+  comments {
+    totalCount
+  }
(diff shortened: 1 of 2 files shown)


MichaelMure:
Looks good but for some reason on my side the icon is way more spaced from the border than on you screenshot. Does that make sense to you? I'm using Firefox.

image

GlancingMind:
Hmm, I'm using firefox too.
I'll play around with it. Maybe I'll find the culprit.

GlancingMind:
Found the culprit. Working on a fix :construction_worker_man:

GlancingMind:
Solved.
For short title.
image

With long titles.
image

MichaelMure:
Nice, it even works fairly well with a phone size screen.

PR #605, opened 2021-03-26
WebUI: Add user profile

GlancingMind:
Adds a profile page for identities.
Clicking on the user name in the AppBar will display a menu.
Author component will link to the profile page of the respective identity.

Requires:

- [x] email fix 
- [x] author dark mode color fix
- [x] discuss, what should happen with the bug listing on profil page


api/graphql/schema/repository.graphql
```diff
@@ -17,6 +17,10 @@ type Repository {
         query: String
     ): BugConnection!
 
+    allBugsDetail(
```
MichaelMure, line 20: What is that for?

GlancingMind: I think this might be from trying something out.
At least it looks like it is not used anywhere.
Will remove it.

GlancingMind: Removed https://github.com/MichaelMure/git-bug/pull/605/commits/580534f6829bd11007b744362c038edba422ae99

GlancingMind:
I think I got everything. Will remove draft status from this PR.
There are two minor things.

1. The links currently make a full site refresh. I'll fix this with another PR.
2. What should happen to the authored bug listing on the profile page?
2.1 Remove it?
2.2 Keep it, but create another issue to improve upon it?

GlancingMind:
Rebased

GlancingMind:
Converted back to draft, as this PR https://github.com/MichaelMure/git-bug/pull/623 should be merged first.
Otherwise a conflict might happen due to some directory changes. At least, I hope that this will prevent a conflict.

MichaelMure:
> * What should happen to the authored bug listing on the profile page?
>   2.1 Remove it?
>   2.2 Keep it, but create another issue to improve upon it?

Yeah about that, I finally realized that my idea of "user timeline" is not exactly easy as the data we have is not in this form. We would have to pre-process it another way, which is really complex and probably not worth it.

In that case ... I'm not sure if it's really useful in this form.

It could be useful to see the bugs (and PRs, ...) opened by someone, or participated in ... but that is really the job of the bug list page.

We could add links to the bug list pre-filtered, but then again, if the bug list is well done that shouldn't bring so much. A bit, but not much. It's quite easy to do though, so it might still be worth it.

What do you think? It's up to you :)

GlancingMind:
Rebased. Maybe this fixes the pipeline fail. Let's see...

MichaelMure:
@GlancingMind is that still a draft?

GlancingMind:
Unfortunately, yes.
I personally would remove the bug list from the profile page and then this could be merged, but as @LenaNoctua also worked on this, I thought I might get her feedback first. Let's wait until tomorrow for an response.

LenaNoctua:
Then remove it *sniff*

MichaelMure:
Sorry @LenaNoctua, I know that sucks :-|

GlancingMind:
I've an idea. Maybe we could replace the bug listing with some status field, how many bugs the user has authored, participated etc.
Each one of these statistic could then have a link to the bug list page with the proper query-string. But then the profile page has to construct the proper URL with parameters.

MichaelMure:
This is the sort of things I was hinting at in https://github.com/MichaelMure/git-bug/pull/605#issuecomment-815167114

Works for me!

GlancingMind:
Hmm the GraphQL query to get a the counts is quite bad. 

```GraphQL
query GetBugsByQuery($authoredQuery: String, $participantQuery: String, $actionQuery: String) {
  repository {
    authored: allBugs(query: $authoredQuery) {
      totalCount
    },
    participated: allBugs(query: $participantQuery) {
      totalCount
    }
    actions: allBugs(query: $actionQuery) {
      totalCount
    }
  }
}
```
Query arguments:
```GraphQL
{
  "authoredQuery": "author:28845ba",
  "participantQuery": "participant:28845ba",
  "actionQuery": "actor:28845ba"
}
```

As GraphQL won't support string interpolation.
This means I can't do following :disappointed:  :
```GraphQL
authored: allBugs(query: `author:$id`) {
   totalCount
},
participated: allBugs(query: `participant:$id`) {
   totalCount
},
...
```

But at least it works. So only one thing left todo: construct query string from identity page. Which I think isn't so nice, as I the query logic would then exist in two places (bug list page and profile page).

MichaelMure:
GraphQL has this concept of variable, maybe you can use that? https://graphql.org/learn/queries/#variables

GlancingMind:
How does this look like?
image

Every statistic line is a link to the proper filter.
- http://localhost:3000/?q=author%3A0e75d29+sort%3Acreation
- http://localhost:3000/?q=participant%3A0e75d29+sort%3Acreation
- http://localhost:3000/?q=actor%3A28845ba+sort%3Acreation

MichaelMure:
Looks great but:
- Id (truncated) --> ID
   - ^ assuming that the tooltip give the full Id 
- Participations --> Participated

Also, it's not necessarily "your account", right? That page could also work to inspect other people's profile, no? Then "Your account" could just be the name of the identity.

MichaelMure:
Also, the public keys (if any) could be displayed here.

MichaelMure:
And even the history of the public keys.

GlancingMind:
>     * Id (truncated) --> ID
>     * ^ assuming that the tooltip give the full Id

Yes the tooltip displays the full ID when hovered with the mouse.
>     * Participations --> Participated

Missed that one. Will fix.

> Also, it's not necessarily "your account", right? That page could also work to inspect other people's profile, no? Then "Your account" could just be the name of the identity.

Good catch. I would rather remove the "Your" from "Your Account" as the name is already displayed in another line.
Further "Account" matches "Statistics" so perfectly :D
But I could also remove it if you prefer.

> Also, the public keys (if any) could be displayed here.
> And even the history of the public keys.

I'm not sure where I can get these details. The graphql schema seem to not expose them.

GlancingMind:
Okey, I replace "Account" with the user name. Looks better.

MichaelMure:
> I'm not sure where I can get these details. The graphql schema seem to not expose them.

Yeah correct, that would need to be done.

GlancingMind:
> Yeah correct, that would need to be done.

Okey, should this be done in another issue?

MichaelMure:
> Okey, should this be done in another issue?

yes

MichaelMure:
Let me know when you think this is ready to merge.

GlancingMind:
You have my go. :rocket:

MichaelMure:
I'm not sure why but "Participated in" and "Interacted with" is just not working for me:
image

Any idea why?

webui/src/pages/identity/Identity.tsx
```diff
@@ -0,0 +1,147 @@
...
+type Props = {
+  identity: IdentityFragment;
+};
+const Identity = ({ identity }: Props) => {
+  const classes = useStyles();
+  const user = identity;
+
+  const { loading, error, data } = useGetUserStatisticQuery({
+    variables: {
+      authorQuery: 'author:' + user?.humanId,
```
MichaelMure, line 54: as much as possible, you should make the queries with the full ID (especially if it's not visible to the user) to avoid an ID collision

webui/src/pages/identity/GetUserStatistic.graphql
```diff
@@ -0,0 +1,13 @@
+query GetUserStatistic($authorQuery: String!, $participantQuery: String!, $actionQuery: String!) {
```
MichaelMure, line 1: You could pass just the ID and use the variable multiple time, no?

webui/src/pages/identity/Identity.tsx
GlancingMind: Oh yeah, my fault. Will fix.

webui/src/pages/identity/GetUserStatistic.graphql
GlancingMind: Unfortunatlly not. I tried, but I cannot pass the ID variable to the query string in GraphQL.
I would like to have something like this.

```GraphQL
# Define the author query by concatenating the `author:` with the ID.
authored: allBugs(query: `author:$id`) {
   totalCount
},
participated: allBugs(query: `participant:$id`) {
   totalCount
},
```
Unfortunatly this doesn't work, as GraphQL does not support string concatenation. At least, that was the result of my research. That's why I have to construct the query in Typescript and pass the query as variable to graphql.
Not beautiful, but it works.

MichaelMure: hu, weird limitation indeed. Works for me!

webui/src/pages/identity/Identity.tsx
GlancingMind: This should work. https://github.com/MichaelMure/git-bug/pull/605/commits/774bae6d432c13cfa0de3ddc2f111927743243fe

GlancingMind:
> I'm not sure why but "Participated in" and "Interacted with" is just not working for me:
> image
> 
> Any idea why?

Interesting. If I add a new bug, all three statistic field are bumped by 1.

About the statistics mentioned here https://github.com/MichaelMure/git-bug/pull/605#issuecomment-818861652
Those are a result of importing the issues of my git-bug fork. https://github.com/GlancingMind/git-bug

MichaelMure:
Can you try to rebase and test again with an up to date go binary?

GlancingMind:
Sure. But first I will try to use a new identity.

MichaelMure:
Somehow it seems to be happening in the go code as the API indeed return zero results. Puzzling ...

MichaelMure:
@GlancingMind https://github.com/MichaelMure/git-bug/pull/644 should fix the issue

I suppose you are using an old-ish version of git-bug ?

GlancingMind:
Yeah I think so. I'm currently building the new version.

MichaelMure:
Running `go mod download` before packaging should fix this problem.

I merge my other PR.

GlancingMind:
> Running `go mod download` before packaging should fix this problem.
>
> I merged my other PR.

Okey, I still couldn't reproduce the faulty behavior. I wanted to report back earlier, but I have to try once more :)

MichaelMure:
Let's merge then!

PR #624, opened 2021-04-07
WebUI: Fix and improve rendering of markdown elements

### LINKED ISSUE #579: webui: Emoji support
Currently emojis in the webui are rendered as `:expressionless:`. It would be nice to render the actual emojis :slightly_smiling_face: 
I'm not sure how Markdown is rendered in the webui. Maybe a javascript lib could do the trick.

GlancingMind:
Closes #579 

Add/Fix support of markdown rendering for:

- Bug description
- Bug comment
- Comment history

Introduces following (changes) custom markdown elements:

- Add custom BlockQuoteTag to style quotes.
- Add custom AnchorTag to style anchors and support linking within the SPA.
- - E.g. following links should be equivalent `/bug/abc`, `localhost:3000/bug/abc`, both will link withing the SPA.
-  Always add line break after an image tag to prevent text to be inline with an image.

### DIFF
--- a/webui/src/components/Content/AnchorTag.tsx
+++ b/webui/src/components/Content/AnchorTag.tsx
@@ -1,0 +1,38 @@
+import React from 'react';
+import { Link } from 'react-router-dom';
+
+import { makeStyles } from '@material-ui/core/styles';
+
+const useStyles = makeStyles((theme) => ({
+  tag: {
+    color: theme.palette.text.secondary,
+  },
+}));
+
+const AnchorTag = ({ children, href }: React.HTMLProps<HTMLAnchorElement>) => {
+  const classes = useStyles();
+  const origin = window.location.origin;
+  const destination = href === undefined ? '' : href;
+  const isInternalLink =
+    destination.startsWith('/') || destination.startsWith(origin);
+  const internalDestination = destination.replace(origin, '');
+  const internalLink = (
+    <Link className={classes.tag} to={internalDestination}>
+      {children}
+    </Link>
+  );
+  const externalLink = (
+    <a
+      className={classes.tag}
+      href={destination}
+      target="_blank"
+      rel="noopener noreferrer"
+    >
+      {children}
+    </a>
+  );
+
+  return isInternalLink ? internalLink : externalLink;
+};
+
+export default AnchorTag;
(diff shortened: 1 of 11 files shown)


GlancingMind:
No sure, if the `nofollow` attribute should be set for [external links](https://github.com/MichaelMure/git-bug/pull/624/files#diff-220cc699c35486b4fcba7dd0b0775d68438d372daefae579dcb98298f009757fR29).

webui/src/components/Content/AnchorTag.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+import { makeStyles } from '@material-ui/core/styles';
+
+const useStyles = makeStyles((theme) => ({
+  tag: {
+    color: theme.palette.text.secondary,
+  },
+}));
+
+const AnchorTag = ({ children, href }: React.HTMLProps<HTMLAnchorElement>) => {
```
MichaelMure, line 12: I'm not sure what happening here but it looks like you are doing something more complicated than necessary.
What about something like https://github.com/MichaelMure/git-bug/blob/master/webui/src/pages/list/ListQuery.tsx#L393 ?

GlancingMind: Not sure, if I understand you correctly, but that is whats done in the last line.
`return isInternalLink ? internalLink : externalLink;`
Of course I could also inline internalLink and externalLink. But I thought, that the naming them might be better for understanding.
I might be wrong, but I think the link component didn't work when linking outside of git-bug (e.g. a link to github.com).
That's why I use an anchor for external links and the link component for internal links. This will result in fast page transitions within the SPA and when an external link is clicked, a "you are leaving git-bug... Are you sure?"-Page could be displayed. Though, currently not implemented. :)

---

Also let me describe following piece of code and why it might be necessary:
```
const origin = window.location.origin;
const isInternalLink = destination.startsWith('/') || destination.startsWith(origin);
const internalDestination = destination.replace(origin, '');
```

Suppose I want to link to a bug within a comment (via markdown).
I have following options:
```
[link to bug](/bug/f312ee9)
[full link to bug](http://localhost:3000/bug/f312ee9)
```
The first link works with the Link-Component, the second doesn't as the link component will prepend `http://localhost:3000`. Resulting in following link `http://localhost:3000/http://localhost:3000/bug/f312ee9`.
With above code, the second link will be converted to the first link. By doing this, both ways of linking are supported.
Which I find important, as the second link is way more convenient for users (they just have to copy the URL).

MichaelMure: sounds good to me :+1:

PR #625, opened 2021-04-08
Minor bugfixes

GlancingMind:
- Fix a console error which occurs when a route is unknown to the navigation bar.
  E.g. users opens the detail page of a bug.

- When editing the title of an bug, only set it if the title is different form the old title.

### DIFF
--- a/webui/src/components/Header/Header.tsx
+++ b/webui/src/components/Header/Header.tsx
@@ -69,20 +69,11 @@ function Header() {
   const location = useLocation();
 
   // Prevents error of invalid tab selection in <Tabs>
+  // Will return a valid tab path or false if path is unkown.
   function highlightTab() {
-    switch (location.pathname) {
-      case '/':
-        return '/';
-      case '/code':
-        return '/code';
-      case '/pulls':
-        return '/pulls';
-      case '/settings':
-        return '/settings';
-      default:
-        // using false as value for tabs will result in no selected tab
-        return false;
-    }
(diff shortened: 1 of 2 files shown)


webui/src/components/Header/Header.tsx
```diff
@@ -67,14 +67,23 @@ const DisabledTabWithTooltip = (props: TabProps) => {
...
+        return '/code';
+      case '/pulls':
+        return '/pulls';
+      case '/settings':
+        return '/settings';
+      default:
+        // using false as value for tabs will result in no selected tab
+        return false;
+    }
+  }
```
MichaelMure, line 86:
```suggestion
  function highlightTab() {
    switch (location.pathname) {
      case '/':
      case '/code':
      case '/pulls':
      case '/settings':
        return location.pathname;
      default:
        // using false as value for tabs will result in no selected tab
        return false;
    }
  }
```

What about this?

GlancingMind: Makes sense. I think I could even remove the switch... Will be back soon.

GlancingMind: This is even more [compact](https://github.com/MichaelMure/git-bug/pull/625/commits/f980d9487ed1e764a364d5a1f385a09835276962#diff-65c829cfa57d5761b963b25ffac663098388927e2ac8abc286e261c8693f1be5R74-R76) :)
What do you think?

MichaelMure:
Merged!

PR #631, opened 2021-04-11
Fix ID string in order to find correct bug instance on comment edit operation

### LINKED ISSUE #629: Import: issue edits and comment edits incorrect
I am importing the issues from https://github.com/rng-dynamics/gb-test-two.

When I use git-bug from the current master branch (bc5f618e), then the import is not correct: The issue https://github.com/rng-dynamics/gb-test-two/issues/2 is missing issue edits and comment edits:

```
 [open] Alexander Scharinger opened this bug on Apr 9 2021

     comment 1


 Alexander Scharinger commented on Apr 9 2021

     comment 3
```

Some time ago it used to work. E.g., on 01b9490 the import is correct:
```
 [open] Alexander Scharinger opened this bug on Apr 9 2021 (edited)

     comment 1, edit 2


 Alexander Scharinger commented on Apr 9 2021 (edited)

     comment 3, edit 4
```

rng-dynamics:
Fixes #629

### DIFF
--- a/bug/op_edit_comment.go
+++ b/bug/op_edit_comment.go
@@ -34,12 +34,12 @@ func (op *EditCommentOperation) Apply(snapshot *Snapshot) {
 	// Todo: currently any message can be edited, even by a different author
 	// crypto signature are needed.
 
-	snapshot.addActor(op.Author_)
+	// Recreate the Comment Id to match on
+	commentId := entity.CombineIds(snapshot.Id(), op.Target)
 
 	var target TimelineItem
-
 	for i, item := range snapshot.Timeline {
-		if item.Id() == op.Target {
+		if item.Id() == commentId {
 			target = snapshot.Timeline[i]
 			break
 		}
@@ -51,7 +51,7 @@ func (op *EditCommentOperation) Apply(snapshot *Snapshot) {
 	}
 
 	comment := Comment{
-		id:       op.Target,
+		id:       commentId,
 		Message:  op.Message,
 		Files:    op.Files,
 		UnixTime: timestamp.Timestamp(op.UnixTime),
@@ -62,12 +62,18 @@ func (op *EditCommentOperation) Apply(snapshot *Snapshot) {
 		target.Append(comment)
 	case *AddCommentTimelineItem:
 		target.Append(comment)
+	default:
+		// somehow, the target matched on something that is not a comment
+		// we make the op a no-op
+		return
 	}
+
+	snapshot.addActor(op.Author_)
 
 	// Updating the corresponding comment
 
 	for i := range snapshot.Comments {
-		if snapshot.Comments[i].Id() == op.Target {
+		if snapshot.Comments[i].Id() == commentId {
 			snapshot.Comments[i].Message = op.Message
 			snapshot.Comments[i].Files = op.Files
 			break
--- a/bug/comment.go
+++ b/bug/comment.go
@@ -11,6 +11,8 @@ import (
 
 // Comment represent a comment in a Bug
 type Comment struct {
+	// id should be the result of entity.CombineIds with the Bug id and the id
+	// of the Operation that created the comment
 	id      entity.Id
 	Author  identity.Interface
 	Message string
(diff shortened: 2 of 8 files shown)


rng-dynamics:
Well, the tests fail, but it shows the problem: in `EditCommentOperation.Apply()` the `op.Target` is not a combined ID and no target is found.

MichaelMure:
Hmm that's an overlook on my part indeed. I'm not sure that's the right fix though.

Since the introduction of the combined Id, there is two ways to address an operation:
- `Operation.Id()` returns an Id in the context of a Bug (or other future Entity). This Id is decoupled from the storage (it's not the result of a commit as before) and is decoupled from the Entity (the Id doesn't include the Entity's Id). This last property is interesting IMHO as that removes a quite annoying coupling. 
- `Entity.CombineIds()` provide a way to combine Entity and Operation's Id to give a way to efficiently locate an Operation without having to iterate over all of them to find which Entity has a Operation matching the Id [1].

The second addressing scheme is only relevant in some case like comments where you want to avoid having to reference both Entity's Id and comment's Id. Otherwise, almost everything should be done with the first scheme. As an example, the target of an `EditCommentOperation` should be done with the first scheme as we already know in which Entity to look into when resolving.

TL;DR: I believe combining the Ids should be done in the bridge, closer to the few places where this second scheme make sense.  It makes sense to me that `ResolveOperationWithMetadata` would return the normal Id.

What I'm not clear on right now is where do you need this combined Id in the bridge?

[1]: ... that makes me think that the kind of entity should probably be encoded in that combined Id as well

rng-dynamics:
> Hmm that's an overlook on my part indeed. I'm not sure that's the right fix though.

Sure, we can just fix it in another way.

> What I'm not clear on right now is where do you need this combined Id in the bridge?

What we do in the bridge when there is an issue edit or a comment edit:

1. Given the Github Id `ghTargetId` find the comment or issue which should be modified:
https://github.com/MichaelMure/git-bug/blob/c71d26d513e4e5e36e6983e05dee4ad28ec664c7/bridge/github/import.go#L409
2. Pass it to `EditCommentRaw()`:
https://github.com/MichaelMure/git-bug/blob/c71d26d513e4e5e36e6983e05dee4ad28ec664c7/bridge/github/import.go#L438-L446
3. `target` from above ends up as `op.Target` in `EditCommentOperation::Apply()` and it is compared with the elements in `snapshot.Timeline` which are combined IDs:
https://github.com/MichaelMure/git-bug/blob/c71d26d513e4e5e36e6983e05dee4ad28ec664c7/bug/op_edit_comment.go#L39-L46
4. If `target` is not of the same kind as the IDs in `snapshot.Timeline`, then obviously the target is not found and `EditCommentOperation::Apply()` returns without doing anything:
https://github.com/MichaelMure/git-bug/blob/c71d26d513e4e5e36e6983e05dee4ad28ec664c7/bug/op_edit_comment.go#L48-L51

That's what is happening in #629.

MichaelMure:
Alright so in this case (3) is incorrect. `EditCommentOperation::Apply()` should combine the Ids to create the correct comment Id to search in the timeline, and below in the comment list as well. See `CreateOperation::Apply()` where the mirror operation is done. My bad!

MichaelMure:
Thanks for the break down, very helpful!

MichaelMure:
.. this also means that the tests for `CreateOperation` are incorrect :-|

bug/op_edit_comment.go
```diff
@@ -51,7 +52,7 @@ func (op *EditCommentOperation) Apply(snapshot *Snapshot) {
 	}
 
 	comment := Comment{
-		id:       op.Target,
+		id:       entity.CombineIds(snapshot.Id(), op.Id()),
```
rng-dynamics, line 55: @MichaelMure: Is that the correct Id for the new `Comment`?

MichaelMure: That's incorrect actually. A comment Id is the same Id as the operation that created it (only combined with the Bug Id).

Here you combined the Bug Id with the *edit* operation Id.

MichaelMure: I'll fix it and try to make things a bit more clear in the code.

MichaelMure:
@rng-dynamics I think it should be good now. Does that looks correct to you?
I added some comments, is that clear enough?

rng-dynamics:
Great! It looks good and I can confirm that it fixes #629.

MichaelMure:
🎉

PR #632, opened 2021-04-17
make sure every text input is safe and validated

### LINKED ISSUE #630: Import error: title is not fully printable
I am importing all the the issues from https://github.com/golang/go. (Currently I am using git-bug from PR #585).

I get the following error:
```
import error: issue creation: title is not fully printable
```
Here is the issue which causes that: https://github.com/golang/go/issues/43562.
The JSON from the GraphQL API looks as follows:
```
"issue": {
        "title": "src/net/http/header when header have\u0002 ",
        "body": "<!--\r\nPlease answer these questions before submitting your issue. Thanks!\r\nFor questions please use one of our forums: https://github.com/golang/go/wiki/Questions\r\n-->\r\n\r\n### What version of Go are you using (`go version`)?\r\n\r\n<pre>\r\n$ go version\r\ngo version go1.15.2 linux/amd64\r\n</pre>\r\n\r\nwhen i post header like (202012300001\u0002sometext\u0002timestampvalue)  to go http server   \r\ni got error  400 bad request  invalid header value  \r\n\r\n\r\n"
}
```

MichaelMure:
fix #630

### DIFF
--- a/api/graphql/resolvers/mutation.go
+++ b/api/graphql/resolvers/mutation.go
@@ -5,11 +5,12 @@ import (
 	"time"
 
 	"github.com/MichaelMure/git-bug/api/auth"
-	"github.com/MichaelMure/git-bug/entity"
 	"github.com/MichaelMure/git-bug/api/graphql/graph"
 	"github.com/MichaelMure/git-bug/api/graphql/models"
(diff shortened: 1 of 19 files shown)


MichaelMure:
@rng-dynamics I believe this fix your issue.

PR #636, opened 2021-04-19
Support new GitHub token formats

ellsclytn:
GitHub have introduced a new format for their access tokens, which does
not fit within the rules of the previous regex. For the time being, the
previous token format is still being supported by GitHub, so it makes
sense to continue allowing legacy tokens.

https://github.blog/changelog/2021-03-04-authentication-token-format-updates/


ellsclytn:
I went for the approach of two separate regexes because I think it's a little more readable, and a little easier to maintain (dropping support for legacy tokens is as simple as deleting the old regex, and its use in the `if` statement)

MichaelMure:
LGTM, thanks!

PR #651, opened 2021-04-25
CLI: Add non-interactive option to interactive commands

GlancingMind:
Using an empty string for the avatar cli flag e.g. `git-bug user create -a ""` will still result in a prompt. As the avatar URL is an optional option, it should be possible to skip asking for it entirely.
Otherwise automated user creation via a script must make use of pipe hacks.

### DIFF
--- a/commands/root.go
+++ b/commands/root.go
@@ -15,12 +15,6 @@ var GitCommit string
 var GitLastTag string
 var GitExactTag string
 
-type GlobalOptions struct {
-	nonInteractive bool
-}
-
-var rootOptions GlobalOptions
-
 func NewRootCommand() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   rootCommandName,
@@ -66,7 +60,6 @@ _git_bug() {
 `,
 	}
 
-	cmd.PersistentFlags().BoolVar(&rootOptions.nonInteractive, "non-interactive", false, "Do not ask for user input")
 	cmd.AddCommand(newAddCommand())
 	cmd.AddCommand(newBridgeCommand())
 	cmd.AddCommand(newCommandsCommand())
--- a/commands/root.go
+++ b/commands/root.go
@@ -15,6 +15,12 @@ var GitCommit string
 var GitLastTag string
 var GitExactTag string
 
+type GlobalOptions struct {
+	nonInteractive bool
+}
+
+var rootOptions GlobalOptions
+
 func NewRootCommand() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   rootCommandName,
@@ -60,6 +66,7 @@ _git_bug() {
 `,
 	}
 
+	cmd.PersistentFlags().BoolVar(&rootOptions.nonInteractive, "non-interactive", false, "Do not ask for user input")
 	cmd.AddCommand(newAddCommand())
 	cmd.AddCommand(newBridgeCommand())
 	cmd.AddCommand(newCommandsCommand())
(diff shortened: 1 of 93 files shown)


MichaelMure:
That's indeed something I didn't think about. Definitely need to be addressed but after thinking about it for a while, maybe we need a more generic solution that we could apply to all commands?

Maybe something like `--no-interactive` that would disable all the prompts?

GlancingMind:
Sounds like a better solution. I'll take a look.

GlancingMind:
Found some inconsistency. The `user create` command requires email to be set (uses `input.required`), but `Validate()` in `identity/version.go` doesn't check for an empty email. This means, that currently a user can be created (non-interactively) without specifying an email. So I'm not sure, where I should check for the missing email or if I check at all.

GlancingMind:
I've added a global option `non-interactive` which can be used in every git-bug (sub-)command.
The `user create` command won't fail if `--non-interactive` is set and no email is given.

GlancingMind:
Went through all commands and checked for those with interactive prompts, the interactive mode.
Except the bridges itself. Don't know, if I should alter them or the respective bridge author/commiter.
Otherwise, I hope that I didn't miss one.

MichaelMure:
> Don't know, if I should alter them or the respective bridge author/commiter.

Yes please!

commands/root.go
```diff
@@ -15,6 +15,12 @@ var GitCommit string
 var GitLastTag string
 var GitExactTag string
 
+type GlobalOptions struct {
+	nonInteractive bool
+}
+
+var rootOptions GlobalOptions
```
MichaelMure, line 22: I get why you did  that but we really should avoid global variables like that. At the beginning of git-bug each flags were done as global variables like that (following the cobra standard) but it ended up being a major pain. Following that came this new "functional" structure without any globals.

Now to get out of this, there still might be a way to define the flag at the root level (say, with `cmd.PersistentFlags().Bool("non-interactive", ....)`, but it looks to me that we could just avoid that entirely and define the flag only on the commands where it makes sense (that is, those that have prompts).

What do you think?

MichaelMure (changes requested):
As explained in  the comment, we should avoid the global variable.

commands/root.go
GlancingMind: I have the same feeling about the current solution. Good that we are on the same wavelength.
I will remove the global option and instead add a dedicated non-interactive option on each of the interactive commands.
Shouldn't be to much work.

GlancingMind:
> > Don't know, if I should alter them or the respective bridge author/commiter.
> 
> Yes please!

I can think about two ways to implement this.
1. Pass the non-interactive flag to Bridge::Configure(). Then the bridge has to check for the flag itself, before making a prompt.
2. Make the input.go/prompt.go an interface. Then create a *non-interactive*-implementation which basically prints an error instead of prompting for user input. The `bridge configure` command will then pass this *non-interactive* input implementation to Bridge::Configure(). Now the specific bridges won't have to deal with the non-interactive option, as long as they are using the git-bug/input package. (Not sure how to implement this in go, have to read myself into it.)

I think the 2. option is the better one, but wanted your (@MichaelMure) feedback first.

MichaelMure:
I would go for (1), as an extra boolean flag to `Configure`. Let's keep the prompts being prompts instead of adding a new feature there.

GlancingMind:
- [x] GitHub bridge
- [ ] GitLab bridge
- [ ] Jira
- [ ] launchpad

GlancingMind:
So fare I'm done. Hope that I got everything right.

@MichaelMure are you able to test the JIRA bridge authentification?

MichaelMure:
I changed the logic in the bridges from `isNonInteractive` --> `interactive`, good to go now! Thanks!

Issue #655, opened 2021-04-26
Label naming


artumi-richard:
Hi. Thanks for working on git bug.

When creating a label containing a colon ":" character, the search seems to become unusable in the terminal user interface. After adding the label "st:wontfix" to a bug I tried to search with each of these: 

- label:st:wontfix
- label:"st:wontfix"
- label:'st:wontfix'

Each threw the error "can't tokenize". 

I would expect that either the colon character should be made illegal, or some other escape sequence be adopted to allow the search to cope with colons

artumi-richard:
This seems to be a problem on the web interface also.

GlancingMind:
Can confirm for latest release version (7.2).
The current master version seems to allow the following search strings:
- label:"abc:abc"
- label:'abc:abc'
Though `label:abc:abc` fails with *unkown qualifier "label:abc"*.
Also the WebUI seems to not show the open and closed bug count when using above filters. Instead only ??? is displayed as count.

MichaelMure:
Very likely related to https://github.com/MichaelMure/git-bug/pull/568. It's a voluntary change in the query language where we allow to search with a triple `foo:bar:baz`. Before that, `label:st:wontfix` was interpreted (not by design) as `label` + `st:wontfix`. The correct way should be `label:"st:wontfix"` but maybe there is something wrong in the parser or in each UI?

GlancingMind:
The CLI and TermUI work by quoting the label string. The WebUI will display the matching bugs too, but WebUI will still recieves
[Error: unknown qualifier "label:abc"](https://github.com/MichaelMure/git-bug/blob/master/query/parser.go#L77) when using `label:"abc:abc"` as query string. Which is quite strange, as the graphql-playground won't give me an error...

GlancingMind:
I think I've got the error. Let's see if I can fix it.

GlancingMind:
@MichaelMure I've got an interesting idea. Why don't we use go's URLs package for query handling. The only syntax difference should be = instead of : for parameters. This could simplify the code significantly. Parsing would already be implemented and the WebUI can also use URLSearchParams API.

MichaelMure:
Well, this query language is a way for the user to interact with git-bug. The main goal has to be how well it does that job, not how easy it is to implement (as long as bugs are not an issue).

[...]

PR #658, opened 2021-04-29
WebUI: Fix quoted filter strings

GlancingMind:
Fix https://github.com/MichaelMure/git-bug/issues/655#issuecomment-829140370
The parse function dropped user given quotes. This resulted into a wrong formated query to the backend, which further lead to an error. This error then prevented the WebUI from displaying the proper bug count.

- [x] Fix query
- [x] Update unit tests
- [x] Make sure, that labels containing a ':' are quoted. (Must convert user quotes to one specific quoting style, which then can also be used by label value. Otherwise label must match 'value' as well as "value".)

### DIFF
--- a/webui/src/__tests__/query.ts
+++ b/webui/src/__tests__/query.ts
@@ -24,7 +24,6 @@ it('parses a query with multiple filters', () => {
     in: [''],
     freetext: [''],
   });
-
 });
 
 it('parses a quoted query', () => {
@@ -55,9 +54,23 @@ it('parses a quoted query', () => {
   expect(parse(`label:"with:quoated:colon"`)).toEqual({
     label: [`"with:quoated:colon"`],
   });
+});
 
-  expect(parse(`foo:'escaped\\' quotes'`)).toEqual({
-    foo: [`'escaped\\' quotes'`],
+it('should not escape nested quotes', () => {
+  expect(parse(`foo:'do not escape this ->'<- quote'`)).toEqual({
+    foo: [`'do not escape this ->'<- quote'`],
+  });
+
+  expect(parse(`foo:'do not escape this ->"<- quote'`)).toEqual({
+    foo: [`'do not escape this ->"<- quote'`],
+  });
+
+  expect(parse(`foo:"do not escape this ->"<- quote"`)).toEqual({
+    foo: [`"do not escape this ->"<- quote"`],
+  });
+
+  expect(parse(`foo:"do not escape this ->'<- quote"`)).toEqual({
+    foo: [`"do not escape this ->'<- quote"`],
   });
 });
 
@@ -76,9 +89,11 @@ it('parses a complex query', () => {
 });
 
 it('parses a key:value:value query', () => {
-  expect(parse(`meta:github:"https://github.com/MichaelMure/git-bug"`)).toEqual({
-    meta: [`github:"https://github.com/MichaelMure/git-bug"`],
-  });
+  expect(parse(`meta:github:"https://github.com/MichaelMure/git-bug"`)).toEqual(
+    {
+      meta: [`github:"https://github.com/MichaelMure/git-bug"`],
+    }
+  );
 });
 
 it('quotes values', () => {
--- a/webui/src/__tests__/query.ts
+++ b/webui/src/__tests__/query.ts
@@ -7,49 +7,76 @@ it('parses a simple query', () => {
 });
 
 it('parses a query with multiple filters', () => {
-  expect(parse('foo:bar baz:foo-bar')).toEqual({
+  expect(parse(`foo:bar baz:foo-bar`)).toEqual({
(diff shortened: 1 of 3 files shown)


GlancingMind:
Okey the counter on the open/close filters is fixed.
For the time being, this should work. But for the future, the WebUI must find a better solution to the current query handling.

webui/src/__tests__/query.ts
```diff
@@ -7,49 +7,76 @@ it('parses a simple query', () => {
...
 it('parses a complex query', () => {
-  expect(parse('foo:bar foo:baz baz:"foobar" idont:\'know\'')).toEqual({
+  expect(parse(`foo:bar foo:baz baz:"foobar" idont:'know'`)).toEqual({
     foo: ['bar', 'baz'],
-    baz: ['foobar'],
-    idont: ['know'],
+    baz: [`"foobar"`],
+    idont: [`'know'`],
   });
 });
```
MichaelMure, line 71: it would be good imho to add tests for `key:value:value` (ex: `metadata:key:"https://www.example.com/"`) and `key:value` with quoted `:` (ex: `label:"foo:bar"`)

MichaelMure (changes requested):
Looks good but could you add the two extra tests?

> But for the future, the WebUI must find a better solution to the current query handling.

Agree on that. The lexer/parser in go are 100 (simple) lines each. Re-implementing that in js would 1) ensure feature parity and 2) allow for simpler query manipulation (dedup, reordering, simplification, quote transformation ...).

webui/src/__tests__/query.ts
```diff
@@ -7,49 +7,76 @@ it('parses a simple query', () => {
...
 
-  expect(parse('foo:\'bar "nested" quotes\'')).toEqual({
-    foo: ['bar "nested" quotes'],
+  expect(parse(`label:"multi word label with 'nested' quotes"`)).toEqual({
+    label: [`"multi word label with 'nested' quotes"`],
   });
 
-  expect(parse("foo:'escaped\\' quotes'")).toEqual({
-    foo: ["escaped' quotes"],
+  expect(parse(`foo:'escaped\\' quotes'`)).toEqual({
```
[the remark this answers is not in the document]
GlancingMind: I'm pretty sure, that this test was already there and I'm not sure what todo next. Remove it?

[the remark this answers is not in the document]
MichaelMure: This test was written before the proper lexer/parser existed in go. The right test would actually do the opposite, but then you would have to change the parser. Change the test and comment it until the code is changed?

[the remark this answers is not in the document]
GlancingMind: Hope that this does the right thing. https://github.com/MichaelMure/git-bug/pull/658/commits/1ad5f6b78333dd19986fa0a149bf2ec85082f852

webui/src/__tests__/query.ts
GlancingMind: Added test cases here https://github.com/MichaelMure/git-bug/pull/658/commits/1987ace79d872e8acf8bab7b9ec22af976bebcca . Though not sure about the proper expected value of the k:v:v test.

MichaelMure: I guess that's good enough until the whole thing is reworked.

GlancingMind:
@MichaelMure I have nothing to add. If you are fine with this changes, this can be merged. :-)

webui/src/__tests__/query.ts
```diff
@@ -7,49 +7,102 @@ it('parses a simple query', () => {
...
 
-  expect(parse("foo:'escaped\\' quotes'")).toEqual({
-    foo: ["escaped' quotes"],
+  expect(parse(`label:"with:quoated:colon"`)).toEqual({
+    label: [`"with:quoated:colon"`],
+  });
+});
+
+it('should not escape nested quotes', () => {
+  expect(parse(`foo:'do not escape this ->'<- quote'`)).toEqual({
```
MichaelMure, line 60: This behavior is also different in the go parse I believe. The first corresponding quote would terminate the quoted section. It looks like the regex based parser is too greedy and try to match as much as possible?

MichaelMure: It's fine for now if the JS parser is not matching exactly the go one, just write the correct test and comment the ones that don't pass. That way it's explicit were the mismatch are and we don't have tests that mislead the reader into thinking that the code is correct.

GlancingMind: Should be fixed. Added also additional test cases to catch greediness.
Let me now if something is missing and before you merge, so that I can get rid of the eslint hotfixes. :-)

GlancingMind:
Rebased, to get rid of the fix-eslint commits and a fix-typo commit.

MichaelMure:
good to go!

Issue #553, opened 2021-02-05
WebUI - Allow to Close and Comment an issue


claudioantonio:
At moment the CloseButton is only enabled when there is no comment typed.

When You enter in bug details page, Close Button is enabled (because the textfield for comments is empty) and Comment button is disabled. When You type the first key in the textfield for comments, Close button become disabled and Comment button become enabled.

Github shows a "Close and Comment" button when the user starts to type a comment but our mutation for closeBug don´t accept a comment as argument. 😉 

```graphql
input CloseBugInput {
    """A unique identifier for the client performing the mutation."""
    clientMutationId: String
    """"The name of the repository. If not set, the default repository is used."""
    repoRef: String
    """The bug ID's prefix."""
    prefix: String!
}
```

**Michael**
> You actually can do multiple mutation in one query in GraphQL, but I'm not sure how that translate in the tsx code: https://hasura.io/docs/1.0/graphql/core/mutations/multiple-mutations.html

claudioantonio:
Can You tell me what use cases You think will be possible with multiple queries in one request?

MichaelMure:
You got me, I actually don't have much example for something else that would require multiple mutations. I would expect that the same need arise elsewhere when we add more features (pull-request ...).
In any case and more importantly, the UX/UI should not be made worse because of a technical limitation if we can.

claudioantonio:
Specifically in this case, is It difficult ( not a goodfirstissue :) ) to change the mutation to close bug to add an optional parameter for the comment?
Or even create a new mutation CloseAndCommentBug where bug prefix and comment are required?

claudioantonio:
Hi @MichaelMure, are the options that I mentioned in the last comment viable/acceptable?

MichaelMure:
Hey, sorry for the delay. I guess having a `CloseAndCommentBug` mutation and avoiding entirely the multi-mutations requests make sense. If you want to do that you'll have to:
1. update the graphQL schema files and add the mutation: https://github.com/MichaelMure/git-bug/blob/master/api/graphql/schema/mutations.graphql
2. regenerate the go code of the graphQL server: either through go generate (https://github.com/MichaelMure/git-bug/blob/master/api/graphql/handler.go#L1) or run `go run gen_graphql.go`
3. implement the mutation (https://github.com/MichaelMure/git-bug/blob/master/api/graphql/resolvers/mutation.go)

This should be not too difficult as you have examples of very similar code.

[...]

PR #665, opened 2021-05-05
GraphQL: Implement AddCommentAndCloseBug mutation

GlancingMind:
Should implement #553 ~Still required:~

~- [ ] Marshalling~
~- [ ] Unit Tests~
~- [ ] Test if this operation actually works this way~

### DIFF
--- a/api/graphql/graph/gen_graph.go
+++ b/api/graphql/graph/gen_graph.go
@@ -67,6 +67,13 @@ type DirectiveRoot struct {
 }
 
 type ComplexityRoot struct {
+	AddCommentAndCloseBugPayload struct {
+		Bug              func(childComplexity int) int
+		ClientMutationID func(childComplexity int) int
+		CommentOperation func(childComplexity int) int
+		StatusOperation  func(childComplexity int) int
+	}
+
 	AddCommentOperation struct {
 		Author  func(childComplexity int) int
 		Date    func(childComplexity int) int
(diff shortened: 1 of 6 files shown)


bug/op_comment_and_close.go
```diff
@@ -0,0 +1,61 @@
...
+
+import (
+	"github.com/MichaelMure/git-bug/entity"
+	"github.com/MichaelMure/git-bug/identity"
+	"github.com/MichaelMure/git-bug/repository"
+)
+
+var _ Operation = &CommentAndCloseOperation{}
+
+type CommentAndCloseOperation struct {
```
MichaelMure, line 11: Hold on, hold on :)

You don't need a new operation for that, just to trigger both `AddComment` and `Close` in the implementation of the graphQL mutation.

GlancingMind: I'm at a dead end.
1. I updated [mutations.graphql](https://github.com/MichaelMure/git-bug/blob/5f79e19e0e7211170e0d43924d27692dee2ad1c0/api/graphql/schema/mutations.graphql#L144-L162) and add the Operation type to [operations.graphql](https://github.com/MichaelMure/git-bug/blob/5f79e19e0e7211170e0d43924d27692dee2ad1c0/api/graphql/schema/operations.graphql#L79-L86). But running `go generate` will then result into following error: 
```merging type systems failed: unable to bind to interface: github.com/MichaelMure/git-bug/api/graphql/models.CommentAndCloseBugOperation does not satisfy the interface github.com/MichaelMure/git-bug/bug.Operation```
2. Fix above error by adding the new operation type to graphql-models in [gqlgen.yml](https://github.com/MichaelMure/git-bug/blob/5f79e19e0e7211170e0d43924d27692dee2ad1c0/api/graphql/gqlgen.yml#L45-L46). Running `go generate` again will result in ```modelgen: unable to find type github.com/MichaelMure/git-bug/bug.CommentAndCloseBugOperation
exit status 3```

But does above error not require me to implement this new operation?

MichaelMure: The root cause of your issue is that you modeled the graphQL mutation the same way as the others (good idea, but that made you fall in the trap) with a single returned `operation`. Later, you had to define this operation, which led to writing an actual operation.

Instead, you should consider that this mutation apply two pre-existing operation and ... return both. Then the code naturally devolve into not creating a new operation.

I pushed the code for that. I haven't tested it but it might be ok anyway. Let me know if that solve the issue for you.

MichaelMure: There is still some cleanup to do at least.

GlancingMind:
- Removed the left overs of the dedicated operation.
- Remove unnecessary quotes from comments 
- Rephrased the [AddComment::Message](https://github.com/MichaelMure/git-bug/pull/665/commits/f2c4c0bccba0797291f8a2a728647cffbcb0e559#diff-da956e8a68f3136ce6ee817a0a771f1922a0375f654a255d686ddc9859721c72L30-R30) to match the one of [AddCommentAndCloseBug](https://github.com/MichaelMure/git-bug/pull/665/commits/f2c4c0bccba0797291f8a2a728647cffbcb0e559#diff-da956e8a68f3136ce6ee817a0a771f1922a0375f654a255d686ddc9859721c72R52).
- Tested mutation via GraphQL playground. Works.

I think this PR is good to go. Implementing the AddCommentAndReopenBug and adjusting the WebUI could be done in a dedicated PR.

GlancingMind:
Consolidated all commits into one and rebased.

MichaelMure:
:tada:

PR #680, opened 2021-05-25
WebUI: Support comment-and-close of a bug in one step

GlancingMind:
When a user has some text entered in the comment field, the close bug button will change to "Close bug with comment". 
This way the user does not have to comment and close a bug in two separate steps.

### DIFF
--- a/webui/src/pages/bug/CommentForm.tsx
+++ b/webui/src/pages/bug/CommentForm.tsx
@@ -6,6 +6,7 @@ import { makeStyles, Theme } from '@material-ui/core/styles';
 
 import CommentInput from '../../components/CommentInput/CommentInput';
 import CloseBugButton from 'src/components/CloseBugButton/CloseBugButton';
+import CloseBugWithCommentButton from 'src/components/CloseBugWithCommentButton/CloseBugWithCommentButton';
 import ReopenBugButton from 'src/components/ReopenBugButton/ReopenBugButton';
 
 import { BugFragment } from './Bug.generated';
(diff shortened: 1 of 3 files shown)


webui/src/pages/bug/CommentForm.tsx
```diff
@@ -77,12 +78,14 @@ function CommentForm({ bug }: Props) {
...
-  function getReopenButton() {
-    return <ReopenBugButton bug={bug} disabled={issueComment.length > 0} />;
+  function getBugStatusButton() {
+    if (bug.status === 'OPEN' && issueComment.length > 0) {
+      return <CloseBugWithCommentButton bug={bug} comment={issueComment} />;
+    } else if (bug.status === 'OPEN') {
+      return <CloseBugButton bug={bug} />;
+    } else {
+      return <ReopenBugButton bug={bug} disabled={issueComment.length > 0} />;
+    }
```
MichaelMure, line 88:
```suggestion
    if (bug.status === 'OPEN' && issueComment.length > 0) {
      return <CloseBugWithCommentButton bug={bug} comment={issueComment} />;
    }
    if (bug.status === 'OPEN') {
      return <CloseBugButton bug={bug} />;
    }
    return <ReopenBugButton bug={bug} disabled={issueComment.length > 0} />;
```

MichaelMure: Really just a question of style, but you can avoid the `else` if you return anyway. It often makes a cleaner code.

MichaelMure (approved):

MichaelMure:
I'm starting to run out of emojis to say good job. Good job!

PR #682, opened 2021-05-25
WebUI: Support comment-and-reopen a bug in one step

GlancingMind:
**NOTE**: This PR includes also the changes of following PR: https://github.com/MichaelMure/git-bug/pull/680

When a user has some text entered in the comment field and the bug is closed, then the  reopen button will change to "Close bug with comment" instead of only "reopen bug". This way the user does not have to comment and reopen a bug in two separate steps.


GlancingMind:
Added missing GraphQL commit.

MichaelMure:
I think that need a good rebase now? But otherwise, looks good.

GlancingMind:
Rebased should be good to go. I didn't squash the commits as they are quite atomic.

MichaelMure:
Neat :)

PR #705, opened 2021-08-17
Command ls: Add compact format

### LINKED ISSUE #699: Output of `git bug ls` doesn't fit in an 80-column terminal
When using an 80-column terminal, the output of `git bug ls` wraps around.  (termui works fine, thanks.)

GlancingMind:
This PR resolves #699 by introducing another `compact` format to the ls command.
The `compact` format should print a bug per line while never exceeding the 79 column to prevent wrapping in smaller terminals.

An image  with description of different `ls` outputs can be found [here](https://github.com/MichaelMure/git-bug/issues/699#issuecomment-899599628)

- [ ] A request was made to make this `compact` format the default one.

  >  I would like to most humbly request that this format should be the default, it is compact and informative.

  _Originally posted by @jech in https://github.com/MichaelMure/git-bug/issues/699#issuecomment-899771039_

I would add a follow up commit if nothing speaks against it. @MichaelMure

PS: I tried to port the default format over to "go templates". This could be an interesting solution for the future, as users could then define their own formats by supplying a template string or file without compilation. Also the bugs could be display by rendering a markdown table to the console (when the MD-rendering will work). So if interest exists, I would try to implement this.

### DIFF
--- a/commands/ls.go
+++ b/commands/ls.go
@@ -133,6 +133,8 @@ func runLs(env *Env, opts lsOptions, args []string) error {
 		return lsPlainFormatter(env, bugExcerpt)
 	case "json":
 		return lsJsonFormatter(env, bugExcerpt)
+	case "compact":
+		return lsCompactFormatter(env, bugExcerpt)
 	case "default":
 		return lsDefaultFormatter(env, bugExcerpt)
 	default:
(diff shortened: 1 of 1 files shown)


MichaelMure:
Thanks. We might want to choose one format as default, or detect the terminal size to select which one make sense, but that's already an improvement. Sorry for the delay.

PR #726, opened 2021-12-03
Fix: github bridge: push then pull without duplication

### LINKED ISSUE #690: pushing a new bug to github via git-bug-bridge then pulling back down duplicates the bug
after setting up a bridge to github using a authtoken (i believe it was option number 3 in bridge-configure) i ran the following commands

    git bug bridge push
    git bug bridge pull
    git bug ls

outputs the following  

    70864ca open	Mirror Select list exits unexpectedly             	Jacob Stannix  	   1 💬
    8009d53 open	program will erase drive with existing partitions…	Jacob Stannix  	   1 💬
    a74b02b open	Mirror Select list exits unexpectedly             	Jacob Stannix  	   1 💬
    5076442 open	program will erase drive with existing partitions…	Jacob Stannix  	   1 💬     

issue only appears when a bug created by git-bug is pushed to github and pulled back down. Does not  occur when the issue is created on github's side

rng-dynamics:
Fixes #690

Rationale: When we export the bug instance, we only set `githubID` and `githubURL` (bridge/github/export.go:469)

### DIFF
--- a/bridge/github/import.go
+++ b/bridge/github/import.go
@@ -184,7 +184,7 @@ func (gi *githubImporter) ensureIssue(ctx context.Context, repo *cache.RepoCache
 
 	// resolve bug
 	b, err := repo.ResolveBugMatcher(func(excerpt *cache.BugExcerpt) bool {
-		return excerpt.CreateMetadata[core.MetaKeyOrigin] == target &&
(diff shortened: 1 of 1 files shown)


MichaelMure:
:+1:

PR #760, opened 2022-03-09
Improve GitHub bridge

### LINKED ISSUE #749: problems importing from github.com/nixos/nixpkgs
Ten minutes ago, trying to `git bug bridge pull` the [nixpkgs](https://github.com/nixos/nixpkgs) project's bugs (>5k bugs, >3k PRs -- a very very large set of bugs):

```
import error: Something went wrong while executing your query. Please include `A8DE:54BE:87BAE:9F6E7:62140406` when reporting this issue.
```

I have never been able to get a pull of nixpkgs' bugs to complete, and I've been trying for weeks now (see also #740).

I am using 05d73e1b5321c97cd05133b5ae49d1798bc2fe5d, which has the fix for #585.

Is there any chance that git-bug could keep fetching other bugs when it encounters a problem like this?  It appears that it just aborts as soon as any API call produces an error.  The nixpkgs bugset is ginormous, so it's going to take me at least a week (due to ratelimits) anyways.  But right now it keeps failing after 108 bugs out of 5000+.

Is there an environment variable I can set to cause git-bug to print a backtrace or other context when it gives up like this?

rng-dynamics:
Fix #749. The errors in #749 are seemingly arbitrary errors by github. I think we cannot really avoid them.

On this branch the bridge tries to import the issues one by one and keeps track of the progress in a file. Errors don't terminate the import and the import can be restarted. The details are up to discussion, but for now git-bug writes a text file `git-bug-import` with one issue-number per line.

### DIFF
--- a/bridge/github/client.go
+++ b/bridge/github/client.go
@@ -150,6 +150,7 @@ func (c *rateLimitHandlerClient) callAPIDealWithLimit(apiCall func(context.Conte
 		}
 		// Get the time when Github will reset the rate limit of their API.
 		resetTime := limitQuery.RateLimit.ResetAt.Time
+		resetTime = resetTime.Add(time.Second * 16)
 		msg := fmt.Sprintf(
 			"Github GraphQL API rate limit. This process will sleep until %s.",
 			resetTime.String(),
--- a/bridge/github/import_mediator.go
+++ b/bridge/github/import_mediator.go
@@ -2,6 +2,7 @@ package github
 
 import (
 	"context"
+	"fmt"
 	"time"
 
 	"github.com/shurcooL/githubv4"
@@ -9,10 +10,10 @@ import (
 
 const (
 	// These values influence how fast the github graphql rate limit is exhausted.
-	NumIssues        = 40
-	NumIssueEdits    = 100
-	NumTimelineItems = 100
-	NumCommentEdits  = 100
+	NumIssues        = 100
+	NumIssueEdits    = 50
+	NumTimelineItems = 50
+	NumCommentEdits  = 50
 
 	ChanCapacity = 128
 )
@@ -74,6 +75,13 @@ type CommentEditEvent struct {
 
 func (CommentEditEvent) isImportEvent() {}
 
+type ErrorEvent struct {
+	issueId githubv4.ID
+	err     error
+}
+
+func (ErrorEvent) isImportEvent() {}
+
 func (mm *importMediator) NextImportEvent() ImportEvent {
 	return <-mm.importEvents
 }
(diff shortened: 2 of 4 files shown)


MichaelMure:
Hmmm, I'm not sure this fully makes sense.

A bridge should probably not write it's own file. I understand you are doing that to keep track of the import progress? There is already a restart mechanism, although just storing the timestamp of the last successfully imported issue: https://github.com/MichaelMure/git-bug/blob/1939949fcdd5d540fd20ead79b9c5d27e85e8f9d/bridge/core/bridge.go#L378-L382

Would it be reasonable to piggyback on this mechanism or to extend it if a more granular progress tracking is necessary?

Thank you for working on that.

rng-dynamics:
>  I understand you are doing that to keep track of the import progress?

I used the file to keep track of the progress and then restart the import from where it has left off. 

> A bridge should probably not write it's own file.                                               
                                                                                                                                                                                                                                                                                                          
Sure, the bridge should not write its own file. The solution with the file should be viewed as a proof of concept.

---

The GitHub API is more fragile these days. It's seems to me that we should have  a good story to restart/continue an import, or to import only specific issues.                                                    
          
> Would it be reasonable to piggyback on this mechanism or to extend it if a more granular progress tracking is necessary?

nixpkgs has about 24000 issues. Tracking the issue numbers would mean to just write the failed issue numbers somewhere (into the local config). I am not sure if that makes sense. The list could be quite long. Is the local config supposed to be also edited by humans?   
          
                                                                                                                                                     
What do you think about having a command line argument for importing particular issues? Then I could imagine having a simple message at the end of the import. Imagine something like the following.
                                                                                                                                                     
<pre>
$ git-bug bridge pull
...
new issue: e0612737cebde57cb49b3f596ce045235a260258f70037d0f9860759ceba3e04 
changed label: 682fcc468648af894d56930f714d36124db5ad5348e250ddb6ad42ce1b5b7bd0
<b>import error: couldn't import issues 1254,174098,47786,50032</b>
imported 24645 issues and 6644 identities with default bridge

$ git-bug bridge pull <b>--issues 1254,174098,47786,50032</b>
...
new issue: 3719580985bde516469b3f596ce045235a260258f70037d0f451451435b00002
import error: couldn't import issues 47786
imported 3 issues and 1 identities with default bridge

</pre>

MichaelMure:
> nixpkgs has about 24000 issues. Tracking the issue numbers would mean to just write the failed issue numbers somewhere (into the local config). I am not sure if that makes sense. The list could be quite long. Is the local config supposed to be also edited by humans?

The local config is not supposed to be edited by humans.

I was thinking, in addition (or even replacement) of `git-bug.bridge.%s.lastImportTime` (that record up to what time the importer succeeded), maybe we could have something like `git-bug.bridge.%s.importStatus`, which would hold an opaque string. This string would be up for each bridge to store whatever they want in to track the import progress and continue later. Typically something like `base64(struct{IssueId string, CommentId string}{"1", "2"})`. That way, bridge are free to do what make sense for them, while we still have a common process and code to drive them.

Would that make sense?

> What do you think about having a command line argument for importing particular issues? Then I could imagine having a simple message at the end of the import. Imagine something like the following.

Yeah, importing a subset of issue is definitely something useful. The tricky part is to figure out a way to make that generic for all bridges.

In https://github.com/MichaelMure/git-bug/pull/241, there was a discussion about reusing git-bug's query language as such a bridge neutral way to define a selection. Each bridge would then apply the parsed query into their own filtering, either directly when querying the remote tracker or filtering before actually importing.

That said, this is a bit different than importing specific issue number, but we could extend the query language to support that, it kinda make sense.

MichaelMure:
To be clear about the last part, it could also make sense to have two different ways to configure the import: one for specific issues, one with a query. I'm just thinking out loud.

milahu:
for the nixpkgs repo (#749) i only get 4966 of (9150 + 32277 = 41427) issues = 12%
then `git-bug bridge pull` keeps saying `imported 0 issues and 0 identities`

milahu:
> i only get 4966 of (9150 + 32277 = 41427) issues = 12%

this looks very much like a limitation of the github API
which gives access only to the first 5000 items of a collection

also the github webinterface for [NixOS/nixpkgs](https://github.com/NixOS/nixpkgs) says

> Issues 5k+
> Pull requests 5k+

so we can use the github API to fetch the first 5000 items
and then we must fallback to webscraping to fetch the remaining items

possible workaround: https://github.com/settings/admin

> Export account data
> Export all repositories and profile metadata. Exports will be available for 7 days.

but this requires cooperation from the repo admins

> Migration started.
> Some repositories were too large to be included in the export.
> Please visit the [Migration API](https://docs.github.com/en/rest/migrations) to export: nixpkgs

so the `nixpkgs` repo is too large for "Export account data"

fix: [fetch-github-repo-data.sh](https://github.com/milahu/milahu-nixpkgs-github-data/blob/scripts/fetch-github-repo-data.sh)
example result: [milahu-nixpkgs-github-data](https://github.com/milahu/milahu-nixpkgs-github-data)

sudoforge:
it seems clear to me that one way `git-bug` is used is for archival of external projects' issues. i think this is an important use case to address across the board. @milahu brings up a great point: we need to have a process for cleanly handling large issue counts from the bridges.

that said, i'm going to close this for now to help keep the active board clean. if you wish to work on this tree in the future, you can always check it out like so:

```
$ git fetch https://github.com/git-bug/git-bug.git refs/pull/492/head
$ git checkout FETCH_HEAD
```

PR #788, opened 2022-05-04
Model touchup

hoijui:
Lots of small improvements and some minor fixes to the model docu

### DIFF
--- a/doc/model.md
+++ b/doc/model.md
@@ -19,7 +19,7 @@ An `Operation` is a piece of data including:
 
 - a type identifier
 - an author (a reference to another entity)
-- a timestamp (there is also 1 or 2 Lamport time that we will describe later)
+- a timestamp (there is also 1 or 2 [Lamport time](#time-is-unreliable))
 - all the data required by that operation type (a message, a status ...)
 - a random [nonce](https://en.wikipedia.org/wiki/Cryptographic_nonce) to ensure we have enough entropy, as the operation identifier is a hash of that data (more on that later)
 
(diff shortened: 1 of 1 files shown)


MichaelMure:
Thanks! Did you find that document understandable?

hoijui:
yes! :-)
I liked the new example images, and in general, it is much easier to understand then the first version, from a few years ago. :-)

thoughts:

* `id = hash(json(op))`  - the `json(...)` part: Isn't that language/library/settings dependent? meaning: If I'd re-implement git-bug in an other language, and used that languages standard serialize-to-JSON library, I might get a slightly different JSON (order of properties, indents/general white-space, header-data, ...) then your git-bug/Go implementation.
  I would not know how to best mitigate this.
* Maybe a JSON Schema, documenting your JSON format, would help in understanding/adoption?

MichaelMure:
>     * `id = hash(json(op))`  - the `json(...)` part: Isn't that language/library/settings dependent? meaning: If I'd re-implement git-bug in an other language, and used that languages standard serialize-to-JSON library, I might get a slightly different JSON (order of properties, indents/general white-space, header-data, ...) then your git-bug/Go implementation.
>       I would not know how to best mitigate this.

That's correct but that also wouldn't matter. Once the json is written in a git blob, we use that serialized data to compute the hash. At this point it doesn't matter how exactly is formatted that JSON, as long as it's valid and readable.

>     * Maybe a JSON Schema, documenting your JSON format, would help in understanding/adoption?

Yeah possibly but I feel like there is a long road before having alternative implementation. It wouldn't hurt to document that though.

Issue #797, opened 2022-05-25
Allow `LocalStorage` configuration in the creation of a `GoGitRepo`


smoyer64:
As noted in https://github.com/MichaelMure/git-bug/issues/704#issuecomment-1134822739, `GoGitRepo` has been refactored to provide a reusable data structure.  The repository's `LocalStorage` is hard-coded to the value "git-bug".  This is probably appropriate for applications that extend git-bug such as distributed Kanban boards, distributed code review, etc.  For applications that are not in any way related to git-bug, it's probably better to use a different location for `LocalStorage`.  While most of the rationale for these changes helps avoid the confusion of several applications sharing one `LocalStorage` directory, accidental reuse of the Lamport clocks could result in data consistency issues.

It's therefore recommended that the three functions that return a `GoGitRepo` be altered to accept and additional `applicationName` parameter.  Excluding test and example code, this change only requires that the environment setup in `commands/env.go` be changed to pass "git-bug" as the `applicationName`.

Issue #704, opened 2021-08-12
new release?


jedahan:
Anything stopping a new release from happening?

Having trouble with things like github bridge #581 , and want to use git-bug for a shared project, but feel like its a big ask for people to clone, build and install git-bug instead of using the packaged version.

MichaelMure:
Yeah, sorry about that.

The situation is that I'm buying and refurbishing an apartment. This has kept me extremely busy in the last few months and might last a bit more.

git-bug is close to a new release, but there is still some stuff to do:
- finish/control/merge a bunch of PRs
- finish a bunch of stuff in flux, in particular regarding the data format
- make sure everything is fine
- write doc and communicate around the new data format (make your own distributed datastructure in git easily!)
- write release notes and actually release

jedahan:
Good luck / congrats on the new apartment!

If any of those things are something we could help out on - testing, documentation, etc, lmk.

pka:
Any chance for a new release? The GH bridge seems not working with new tokens in the latest release.

6543:
@MichaelMure should I push this forward?

MichaelMure:
A new release is not so much the problem, it's having the code in a releasable state. Unfortunately some very extensive changes in the core have been amde,that need to be stabilized and exercised up to a point where I/we are confident about it. This is especially important as the next release is likely to be the 1.0, which comes with some form of compatibility guarantee regarding the data stored.

MichaelMure:
Thanks, that's already useful. On my side, the apartment refurbish should finish this month, so hopefully I finally get some free time soon-ish.

MichaelMure:
Let's go back to my initial list:
- ~finish/control/merge a bunch of PRs~
- ~finish a bunch of stuff in flux, in particular regarding the data format~
- make sure everything is fine
- ~write doc and~ communicate around the new data format (make your own distributed datastructure in git easily!)
- write release notes and actually release

I think the last PR I want to have before 1.0 is https://github.com/MichaelMure/git-bug/pull/664. After that, it's "just" a question of making sure everything is ready for prime time, and in particular that git-bug is ready to commit to a data format.

smoyer64:
> write doc and

I'm trying to use the current implementation from the master branch and was trying to "make my own distributed datastructure in git".  Can you point me to the above referenced document?  And one of the problems that I have is that `LocalStorage` always writes to `.git/git-bug/xxx`.  Thanks!

MichaelMure:
[...]

Issue #778, opened 2022-04-10
"git bug rm" broken on master branch


DarthRubik:
It appears that the "git bug rm" command hangs on my machine when I try to remove a test bug that I created....                                                                                                                                                                                                                                                                                                                       I think I have tracked it down to the lock taking on line 502 of "cache/repo_cache_bug.go".                                                                                                                                                                                                                                                                                                                                           I don't know enough about the code to find the root cause at the moment......                                                                                                                                                                                                                                                                                                                                                         Side question:  In order to remove the bug I removed the branch from the "refs/bugs" folder and then deleted the git-bug                                                                                           cache....Should that work?

MichaelMure:
Yes that's enough to do that manually.

What error did you get?

DarthRubik:
I don't get an error it just hangs forever

smoyer64:
@MichaelMure - I've confirmed this issue on my local machine and will have a failing test for it tomorrow.  If you'd like, you can assign this to me.

smoyer64:
Running this locally (Debian Bullseye Linux) results in the hanging CLI as noted above and CTRL-C doesn't terminate the hung state.  Running this using Go test running and the new `testEnv` created as part of https://github.com/selesy/git-bug/tree/fix/778-git-bug-rm-broken allows the supervising Go routine to terminate the operation on timeout and return each Go routine's state.  Here's the output of the test runner when that timeout occurs:

```
smoyer1@smoyer1-carbon:~/git/git-bug$ go test -coverprofile=/tmp/junk ./commands -run TestRm
^C^C^C^CBuilding identity cache... Done.
Building bug cache... Done.

coverage: 24.5% of statements
panic: test timed out after 10m0s

goroutine 36 [running]:
testing.(*M).startAlarm.func1()
        /opt/go/1.18.2/src/testing/testing.go:2029 +0x8e
created by time.goFunc
        /opt/go/1.18.2/src/time/sleep.go:176 +0x32

goroutine 1 [chan receive, 10 minutes]:
testing.(*T).Run(0xc00030d1e0, {0x11906b8?, 0x51b005?}, 0x1242530)
        /opt/go/1.18.2/src/testing/testing.go:1487 +0x37a
testing.runTests.func1(0xc00030d1e0?)
        /opt/go/1.18.2/src/testing/testing.go:1839 +0x6e
testing.tRunner(0xc00030d1e0, 0xc0000d7ca0)
[...]

PR #798, opened 2022-05-25
feat: enable configurable `LocalStorage` in `GoGitRepo`

smoyer64:
As noted in https://github.com/MichaelMure/git-bug/issues/704#issuecomment-1134822739 and further described in https://github.com/MichaelMure/git-bug/issues/797, this PR changes the functions that return a `GoGitRepo` to accept an additional `applicationName` parameter so that the directory containing the repository's `LocalStorage` folder is configurable.

### DIFF
--- a/repository/gogit_test.go
+++ b/repository/gogit_test.go
@@ -17,7 +17,7 @@ func TestNewGoGitRepo(t *testing.T) {
 	require.NoError(t, err)
 	defer os.RemoveAll(plainRoot)
 
-	_, err = InitGoGitRepo(plainRoot)
+	_, err = InitGoGitRepo(plainRoot, testApplicationName)
 	require.NoError(t, err)
 	plainGitDir := filepath.Join(plainRoot, ".git")
 
@@ -26,7 +26,7 @@ func TestNewGoGitRepo(t *testing.T) {
 	require.NoError(t, err)
 	defer os.RemoveAll(bareRoot)
 
-	_, err = InitBareGoGitRepo(bareRoot)
+	_, err = InitBareGoGitRepo(bareRoot, testApplicationName)
 	require.NoError(t, err)
 	bareGitDir := bareRoot
 
@@ -52,7 +52,7 @@ func TestNewGoGitRepo(t *testing.T) {
 	}
 
 	for i, tc := range tests {
-		r, err := OpenGoGitRepo(tc.inPath, nil)
+		r, err := OpenGoGitRepo(tc.inPath, testApplicationName, nil)
 
 		if tc.err {
 			require.Error(t, err, i)
--- a/repository/gogit_test.go
+++ b/repository/gogit_test.go
@@ -72,10 +72,15 @@ func TestGoGitRepo_Indexes(t *testing.T) {
 
 	plainRoot, err := ioutil.TempDir("", "")
 	require.NoError(t, err)
-	// defer os.RemoveAll(plainRoot)
+	t.Cleanup(func() {
+		require.NoError(t, os.RemoveAll(plainRoot))
+	})
 
 	repo, err := InitGoGitRepo(plainRoot, namespace)
 	require.NoError(t, err)
+	t.Cleanup(func() {
+		require.NoError(t, repo.Close())
+	})
 
 	// Can create indices
 	indexA, err := repo.GetBleveIndex("a")
--- a/repository/gogit_test.go
+++ b/repository/gogit_test.go
@@ -17,7 +17,7 @@ func TestNewGoGitRepo(t *testing.T) {
 	require.NoError(t, err)
 	defer os.RemoveAll(plainRoot)
 
-	_, err = InitGoGitRepo(plainRoot, testApplicationName)
+	_, err = InitGoGitRepo(plainRoot, namespace)
 	require.NoError(t, err)
 	plainGitDir := filepath.Join(plainRoot, ".git")
 
@@ -26,7 +26,7 @@ func TestNewGoGitRepo(t *testing.T) {
 	require.NoError(t, err)
 	defer os.RemoveAll(bareRoot)
 
-	_, err = InitBareGoGitRepo(bareRoot, testApplicationName)
+	_, err = InitBareGoGitRepo(bareRoot, namespace)
 	require.NoError(t, err)
 	bareGitDir := bareRoot
 
@@ -52,7 +52,7 @@ func TestNewGoGitRepo(t *testing.T) {
 	}
 
 	for i, tc := range tests {
-		r, err := OpenGoGitRepo(tc.inPath, testApplicationName, nil)
+		r, err := OpenGoGitRepo(tc.inPath, namespace, nil)
 
 		if tc.err {
 			require.Error(t, err, i)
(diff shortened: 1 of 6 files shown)


MichaelMure:
That seems reasonable, yet I wonder a few things:
- how many more parameters are going to be added to the Repo constructors? Should we add a config struct or variadic parameters? I suppose not but from your point of view do you see more required changes if that code is used outside of git-bug?
- instead of `applicationName`, should that be named `tempPath`or something generic like that? There is not really the concept of application at that level.
- should that value have a default and a way to override it for other app instead of making things more complicated here?

smoyer64:
I wondered about the same things but thought it would be easier to have something concrete to discuss!  Following your points above:

- I'm a fan of variadic options (as a simpler form of Rob Pike's [Self-referential functions and the design of options](https://commandcenter.blogspot.com/2014/01/self-referential-functions-and-design.html), so we could create options with a signature of (e.g.) `func(*GoGitRepo) error`.  I'm not sure I'm qualified to decide how many more options there might be - this was the only configuration I needed to use `GoGitRepo` completely separately from git-bug.  I can see how we could provide an option and default for the `path` parameter (try using `cwd` no option is present) but what would the default be for `applicationName`?  If the default is "git-bug", we haven't really protected that directory from misuse.  Could there be a default for `[]ClockLoader`?
- I don't like `application` either but needed to start with something.  Naming is hard!  If we want to be specific, we could use something like `localStoragePath` and `tempPath` certainly works more generally.  Maybe the key is to document what the user might want it for?
- See above - perhaps combining your comments, the default could be "temp-dir" and an option would override that (git-bug specifically in `commands/env.go`.

Thanks for the review ... I'll think this over a bit and try to improve on it tomorrow morning before work!

smoyer64:
Looking at the `GoGitRepo` struct again, I'm not sure what other configuration we might want to pass in the various constructors:

```
type GoGitRepo struct {
	// Unfortunately, some parts of go-git are not thread-safe so we have to cover them with a big fat mutex here.
	// See https://github.com/go-git/go-git/issues/48
	// See https://github.com/go-git/go-git/issues/208
	// See https://github.com/go-git/go-git/pull/186
	rMutex sync.Mutex
	r      *gogit.Repository
	path   string

	clocksMutex sync.Mutex
	clocks      map[string]lamport.Clock

	indexesMutex sync.Mutex
	indexes      map[string]bleve.Index

	keyring      Keyring
	localStorage billy.Filesystem
}
```

The value of `r` itself is based on the `path`, the three mutexes shouldn't be externally accessible and the path for the `clocks` and `indexes` is based on the value of `localStorage`.  So given the current make-up of this type, I'd say there's not anything else we'd want to make configurable.  What's your vision for the future of this type?  This seems like an infrastructure layer that's not as likely to change but if you'd rather see a`GoGitRepoOption` type or a `GoGitRepoConfig` type, I'm happy to do adapt this PR to either of those.

Regarding the idea of which might be optional and if so, what the associated defaults might be, here's how I see the three parameters:

- repoPath: required but could default to `cwd` (my opinion is that this is better handled the way it is in `git-bug`.

- tempPath: required but without a sensible default.  I started on this particular journey because I wanted to use `GoGitRepo` for another application and noticed that my new `dag.Entity`'s clocks were mixed into the `git-bug` local storage.  As a `git-bug` user (and given the importance of issue tracking), my intuition is to protect `git-bug` from that being the accidental behavior.

- clockLoaders: optional as the caller might have an implementation that doesn't use Lamport clocks, indexes, etc.

So, without a good vision of the future I'd suggest that `application` or `applicationName` is a horrible name for this parameter.  Perhaps just change the two string parameters to `repoPath` and `tempPath` so they're clearly differentiated?  What would you like me to do?

MichaelMure:
Thinking a bit more about this I think we can stay with the simple thing of adding a parameter. We can revisit that later if it grows too complicated as we don't have to keep compatibility too much here.

As for the naming, maybe keep `path` and use `namespace` instead of `applicationName` ?

smoyer64:
`namespace` is a great choice! ... will submit an update early tomorrow.

repository/gogit_test.go
```diff
@@ -66,3 +66,36 @@ func TestNewGoGitRepo(t *testing.T) {
 func TestGoGitRepo(t *testing.T) {
 	RepoTest(t, CreateGoGitTestRepo, CleanupTestRepos)
 }
+
+func TestGoGitRepo_Indexes(t *testing.T) {
+	t.Parallel()
+
+	plainRoot, err := ioutil.TempDir("", "")
+	require.NoError(t, err)
+	// defer os.RemoveAll(plainRoot)
```
MichaelMure, line 75:
```suggestion
	defer os.RemoveAll(plainRoot)
```

I think this should be enabled.

MichaelMure:
It might be unrelated to your changes but there is an odd CI  failure on windows:
```
 === CONT  TestGoGitRepo_Indexes
    gogit_test.go:99: 
        	Error Trace:	gogit_test.go:99
        	Error:      	Received unexpected error:
        	            	remove C:\Users\RUNNER~1\AppData\Local\Temp\314289431\.git\git-bug\indexes\a\store: The process cannot access the file because it is being used by another process.
        	Test:       	TestGoGitRepo_Indexes
--- FAIL: TestGoGitRepo_Indexes (0.02s)
```

Not sure what's up really ...

smoyer64:
That is weird and I don't have a Windows machine to chase it down.  Sorry about the commented-out defer statement (I wanted to see the files produced instead of having them deleted).  I'll probably replace that `defer` statement with `t.Cleanup`.  I'll push another version with the fix and after checking that there's not an OS-dependent file separator sneaking into the path.

repository/gogit_test.go
```diff
@@ -66,3 +66,36 @@ func TestNewGoGitRepo(t *testing.T) {
 func TestGoGitRepo(t *testing.T) {
 	RepoTest(t, CreateGoGitTestRepo, CleanupTestRepos)
 }
+
+func TestGoGitRepo_Indexes(t *testing.T) {
+	t.Parallel()
```
smoyer64, line 71: @MichaelMure - I'm pretty sure the issue on Windows is related to this `t.Parallel`.  Is there any interest in running tests concurrently or should I just remove it?

MichaelMure: I suspect the problems comes from Bleve (the indexing thing) not being closed before attempting to remove the repo from disk. It seems to be a problem in other tests as well.

MichaelMure: That is, `repo.Close()` should be called before removing the folder on disk. `EraseFromDisk()` would also work.

smoyer64: I'm writing tests for the `commands` package and found that this morning too!  I don't have a PR ready (this is to diagnose #778) but the branch is https://github.com/selesy/git-bug/tree/fix/778-git-bug-rm-broken.

MichaelMure:
Hey @smoyer64 I added a similar close-repo fix for another test, unfortunately I couldn't push in you branch so I opened https://github.com/MichaelMure/git-bug/pull/803 from the branch pushed to this repo. I invited you as a contributor so feel free to push here directly.

MichaelMure:
You will also get the CI without needing me to push the button each time.

PR #803, opened 2022-05-31
Feat/configurable local storage

MichaelMure:
Replace https://github.com/MichaelMure/git-bug/pull/798 to sneak in a commit.

### DIFF
--- a/commands/env.go
+++ b/commands/env.go
@@ -14,7 +14,7 @@ import (
 	"github.com/MichaelMure/git-bug/util/interrupt"
 )
 
-const gitBugApplicationName = "git-bug"
+const gitBugNamespace = "git-bug"
 
 // Env is the environment of a command
 type Env struct {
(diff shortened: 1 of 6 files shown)


MichaelMure:
@smoyer64 looks like there is another similar issue to fix :-|

smoyer64:
I purposely didn't change the `t.Cleanup()` functions needed and the `t.Parallel()` at the same time, I'm going to remove the latter and see if it builds as I suspect this is due to tests running that are not concurrency-safe.

smoyer64:
@MichaelMure - that was a bit painful but at least I learned something - it's perhaps a bit scary that something in the `TestGoGitRepo_Indexes` test is not concurrency safe only on Windows.  I'd recommend that, for now, we just remember NOT to mark tests parallel since there are three or four packages that use the underlying OS calls - if this is acceptable, we can merge this change.  I can also create a tracking issue as a reminder that the problem with `t.Parallel()` is present if you'd like.

smoyer64:
@MichaelMure

> Or it's just because you added another test that uncovered that problem?

I'm beginning to think this might be the case.  To avoid polluting this PR too much, I created #803.  Adding `t.Cleanup()` functions that close the indexes individually helps (sometimes the tests pass).  I had originally commented out the code that deletes `indexA` but adding that code back in makes the tests fail unpredictably.  Unless I'm missing something, the tests that report failure vary each time the test is executed.  I'm going to try putting the test back to the way it (more failures) and repeatedly run tests to see if a) it occasionally passes and b) whether there's any consistency to the reported failures.  Should we move this conversation to that PR?

smoyer64:
@MichaelMure - I've created #809 to track the problem with flaky tests on Windows builds and have removed the call to `t.Parallel()` from the test I added.  Since you've created the PR, I can't request you as a reviewer in the right side-bar but if you'd be willing to take a look, I'll wait for your approval before we merge.

smoyer64:
@MichaelMure - Without `t.Parallel()`, I think this is ready to merge ... do you have any final comments?

MichaelMure:
Looks good, thank you :-)

PR #804, opened 2022-05-31
fix: try t.Parallel() on existing test

smoyer64:
There appears to be a problem with concurrency on Windows machines.  This PR is experimental (just for testing  on Windows) but may be used if a problem is actually discovered.


sudoforge:

```
$ git fetch https://github.com/git-bug/git-bug.git refs/pull/804/head
$ git checkout FETCH_HEAD
```

Issue #810, opened 2022-06-04
Race conditions in `graphql` and `bridge/github` packages (or package tests)


smoyer64:
While trying to track down the problem with concurrent tests on Windows as described by #809, I ran the tests (Debian Bullseye and Go 1.18.3) with race detection enabled.  Two different issues were found.  The following command was run against the `master` branch at commit 96327c33:

```
go test -race ./...
```

In the `graphql` package, the following race condition was reported:

```
Building identity cache... Done.
Building bug cache... Done.
==================
WARNING: DATA RACE
Write at 0x00c0001b4998 by goroutine 30:
  github.com/MichaelMure/git-bug/api/graphql/models.(*lazyBug).load()
      /home/smoyer1/git/git-bug/api/graphql/models/lazy_bug.go:64 +0x144
  github.com/MichaelMure/git-bug/api/graphql/models.(*lazyBug).Operations()
      /home/smoyer1/git/git-bug/api/graphql/models/lazy_bug.go:148 +0x30
  github.com/MichaelMure/git-bug/api/graphql/resolvers.bugResolver.Operations()
      /home/smoyer1/git/git-bug/api/graphql/resolvers/bug.go:88 +0x72
  github.com/MichaelMure/git-bug/api/graphql/resolvers.(*bugResolver).Operations()
      <autogenerated>:1 +0xb7
  github.com/MichaelMure/git-bug/api/graphql/graph.(*executionContext)._Bug_operations.func2()
      /home/smoyer1/git/git-bug/api/graphql/graph/gen_graph.go:4563 +0x28d
  github.com/99designs/gqlgen/graphql/executor.processExtensions.func4()
      /home/smoyer1/go/pkg/mod/github.com/99designs/gqlgen@v0.17.1/graphql/executor/extensions.go:72 +0x41
  github.com/MichaelMure/git-bug/api/graphql/graph.(*executionContext)._Bug_operations()
      /home/smoyer1/git/git-bug/api/graphql/graph/gen_graph.go:4561 +0x6cd
  github.com/MichaelMure/git-bug/api/graphql/graph.(*executionContext)._Bug.func20()
      /home/smoyer1/git/git-bug/api/graphql/graph/gen_graph.go:12159 +0x159
  github.com/MichaelMure/git-bug/api/graphql/graph.(*executionContext)._Bug.func21()
      /home/smoyer1/git/git-bug/api/graphql/graph/gen_graph.go:12167 +0x4c
[...]

PR #811, opened 2022-06-05
graphql: fix two invalid mutex lock leading to data races

MichaelMure:
Fix one the issue mentioned in https://github.com/MichaelMure/git-bug/issues/810


smoyer64 (approved):
Pull the branch and reran the tests with the `-race` flag and can confirm that the described problem has been resolved.

PR #812, opened 2022-06-05
fix(810): close context to exit loop blocked by race in bridge/github package

smoyer64:
This solution to closing the otherwise blocked thread is to call `Done()` on it's `Context` which allows blocked go routine to end cleanly.

### DIFF
--- a/bridge/github/import_mediator.go
+++ b/bridge/github/import_mediator.go
@@ -89,6 +89,7 @@ func NewImportMediator(ctx context.Context, client *rateLimitHandlerClient, owne
 	}
 	go func() {
 		mm.fillImportEvents(ctx)
+		ctx.Done()
 		close(mm.importEvents)
 	}()
 	return &mm


bridge/github/import_mediator.go
```diff
@@ -89,6 +89,7 @@ func NewImportMediator(ctx context.Context, client *rateLimitHandlerClient, owne
 	}
 	go func() {
 		mm.fillImportEvents(ctx)
+		ctx.Done()
```
MichaelMure, line 92: Hmm, I don't think `Done()` is supposed to be used like that. You normally need to wait on the returned channel within a `select`. Your logic make sense but `Done` doesn't actually terminate the context, it just allows you to wait that it is the case.

smoyer64: Yes ... in this case it's simply used to wait and it could be wrapped in a select without an issue.  There's nothing special about `Done()` though it might more properly be written as `_ = ctx.Done()`.  I'm not married to this solution but figured since I had a working fix for this race, I'd at least start a discussion.

MichaelMure: I tried your branch and the race detector still trigger for me. I think that if you don't wrap `Done()` in a select, you are not actually waiting on anything. It gives you a channel that get ignored and the function terminate.
In any case, this problem is solved now. Thanks for finding it and looking at it :)

MichaelMure: Ha yes, that make sense with `<- ctx.Done()` :)

Issue #809, opened 2022-06-01
Tests that include `t.Parallel()` causes concurrency issues on Windows test runner


smoyer64:
As discussed in #803, including a test that calls `t.Parallel()` and uses a temporary directory causes unpredictable failures on the Windows build server.  Rerunning a failed check often succeeds but failures report that many different tests randomly fail.  This is a pretty clear sign that there's a concurrency issue.  Here's the code that originally caused the problem:

```go
func TestGoGitRepo_Indexes(t *testing.T) {
	t.Parallel()

	plainRoot, err := ioutil.TempDir("", "")
	require.NoError(t, err)
	t.Cleanup(func() {
		require.NoError(t, os.RemoveAll(plainRoot))
	})

	repo, err := InitGoGitRepo(plainRoot, namespace)
	require.NoError(t, err)
	t.Cleanup(func() {
		require.NoError(t, repo.Close())
	})

	// Can create indices
	indexA, err := repo.GetBleveIndex("a")
	require.NoError(t, err)
	require.NotZero(t, indexA)
	require.FileExists(t, filepath.Join(plainRoot, ".git", namespace, "indexes", "a", "index_meta.json"))
	require.FileExists(t, filepath.Join(plainRoot, ".git", namespace, "indexes", "a", "store"))

	indexB, err := repo.GetBleveIndex("b")
	require.NoError(t, err)
	require.NotZero(t, indexB)

	// Can get an existing index
	indexA, err = repo.GetBleveIndex("a")
	require.NoError(t, err)
	require.NotZero(t, indexA)

	// Can delete an index
	err = repo.ClearBleveIndex("a")
	require.NoError(t, err)
	require.NoDirExists(t, filepath.Join(plainRoot, ".git", namespace, "indexes", "a"))
}
```

Linux and MacOS filesystems have a different behavior than Windows - if you delete a file that's in use, it will remove the directory entry but the inodes won't be deleted until all the related file handles have been closed.  Changing the code above to the following makes the tests fail less frequently:

```go
func TestGoGitRepo_Indexes(t *testing.T) {
	t.Parallel()

	plainRoot := t.TempDir()

	repo, err := InitGoGitRepo(plainRoot, namespace)
	require.NoError(t, err)
	t.Cleanup(func() {
		require.NoError(t, repo.Close())
	})

	// Can create indices
	indexA, err := repo.GetBleveIndex("a")
	require.NoError(t, err)
	require.NotZero(t, indexA)
[...]

PR #817, opened 2022-06-16
refactor(809): guarantee test cleanup

smoyer64:
In the current test code, when a test repository is created using `CreateGoGitTestRepo(...)`, the test write has to remember to include `defer CleanupTestRepos(...)` so that resources are properly deleted and file handles are closed.  This is missing in at least a few instances.

This PR attempts to guarantee proper test hygiene by making the following changes:

- Make `CreateGoGitTestRepo(...)` a `t.Helper()`
- Use `t.TempDir()` to create underlying temporary directories (as they're guaranteed to be deleted when the test completes)
- Use `t.Cleanup()` to execute `repo.Close() so it's guaranteed to be performed when the test completes.

Related to #809 and may supercede #804.


MichaelMure:
Love the smell of cleaner code.

Are you planning to deal with the interupt catcher in tests as well?

smoyer64:
I actually deleted the interrupt catcher from the tests in `commands/` (for now).  I marked this as a draft as I'm also thinking through the best way to use `CreateGoGitTestRepo` in the (new) tests in `commands` (probably by adding a way to get the test repos' temporary directory via the `repoTest` interface.)

I guess that brings up a question ... do you think we need to catch signals during testing?  I'll have to go reread how `t.Cleanup()` (and the related `t.TempDir()`) are handled if the test runner receives a signal - my recollection is that tests try to stop with `SIGINT` and `SIGTERM` but have no guarantees with `SIGKILL`.

smoyer64:
@MichaelMure - TLDR: I don't think the interrupt catcher should ever be included in tests.

This took both research and experimentation but what I believe I've found is that if you catch a signal in even one test, it disables the default behavior (the test runner immediately kills all child tests and exits with a message).  So in essence, signal handling (for the specified signals) is completely disabled.  The `CleanerFunc` will indeed be executed in this case but the test continues to run since it's in a different go routine.  My initial thought was to call `t.Fatal()` with a message indicating that the test had been signaled to quit but the [comment](https://pkg.go.dev/testing#T.FailNow) for `t.FailNow()` includes the following:

> FailNow marks the function as having failed and stops its execution by calling runtime.Goexit (which then runs all deferred calls in the current goroutine). Execution will continue at the next test or benchmark. FailNow must be called from the goroutine running the test or benchmark function, not from other goroutines created during the test. Calling FailNow does not stop those other goroutines.

The result is that the `CleanupFunc` will have no (positive) effect on the running test, though it might close resources that would cause the test to fail.  My opinion is that it's probably better to let the OS cancel the parent test runner, release any file handles in use and then eliminate unused files in `/tmp` the next time the OS starts.

I've removed this PR's "draft" status - I think it's ready for review.  This PR will not solve the problem described in #809 but knowing that all resources are being closed will reduce the amount of code that needs to be debugged.

repository/gogit_testing.go
```diff
@@ -1,21 +1,28 @@
...
-	"io/ioutil"
 	"log"
+	"path/filepath"
+	"strings"
+	"testing"
 
 	"github.com/99designs/keyring"
 )
 
+type TestingT interface {
```
MichaelMure, line 12: Hmm, why defining this interface? If you don't want to bind to `testing.T`, maybe use `testing.TB`?

repository/gogit_testing.go
```diff
@@ -62,3 +75,18 @@ func SetupGoGitReposAndRemote() (repoA, repoB, remote TestedRepo) {
 
 	return repoA, repoB, remote
 }
+
+func RepoDir(t *testing.T, repo TestedRepo) string {
```
MichaelMure, line 79: Hmmm, I'm not sure about this.

My first concern is that if you really need to access the repo path for testing, it might be better to put an extra function in [repoTest](https://github.com/MichaelMure/git-bug/blob/44f61b66775e77fd8b42ba3ea57b2c323463b725/repository/repo.go#L205).

But then ... I'm not sure it's correct to test actual files, or even assume there *is* files. For example, `mockRepo` is purely in memory. In the future, we could have other alternate implementation for a Repo: in the LocalStorage of a browser, in a database ... `Repo` is a very good interface to transplant git-bug onto another system.

Which means ... [TestGoGitRepo_Indexes](https://github.com/MichaelMure/git-bug/blob/f3d316d16cc9fadd90faaaf77becbb82f1e09367/repository/gogit_test.go#L63) should likely be implemented as a reusable `func RepoBleveTest(t *testing.T, repo RepoBleve)` in `repo_testing.go`, and ignore files entirely. If the indexing works, do we really care about the files?

MichaelMure:
> TLDR: I don't think the interrupt catcher should ever be included in tests.

Yes, let's remove that from tests.

repository/gogit_testing.go
MichaelMure: > The TestedRepo itself includes a function called EraseFromDisk so to some degree, we're expecting that the repo used for testing is disk backed.

Not really, at least in my head. `TestedRepo` is a low quality catch-all to add testing specific capabilities. If you look at the implementation of `mockRepoTest`, `EraseFromDisk` just do nothing. Admittedly, the naming could be improved, or even better this should be removed entirely as it's now possible with this PR.

1. I see that one as an extra test, specific for `GoGitRepo`. In this case it's fine to bypass the interface boundary and test internal specific things that are specially error prone (where the files end up in different scenario).

2. A similar case could be made about this one the need for a generic test for the index remains.

smoyer64:
Ready for re-review ... items 1. and 2. above still use an unexported function to retrieve the GoGit repository's directory.  Tests in the `commands` package no longer rely on a repo being disk-based.

smoyer64:
Thanks ... this will make it a lot easier to chase down #809 by making small changes from `master`!

PR #835, opened 2022-07-25
entity/dag: proper base operation for simplified implementation

MichaelMure:
- reduce boilerplace necessary to implement an operation
- consolidate what an operation is in the core, which in turn pave the way for a generic cache layer mechanism
- avoid the previously complex unmarshalling process
- support operation metadata from the core
- simplified testing

@smoyer64 FYI


MichaelMure:
> $ git diff --shortstat  origin/master -- bug
> 23 files changed, 205 insertions(+), 1099 deletions(-)

Writing an entity just became much easier!

Issue #836, opened 2022-08-01
Update the output of `ls` so that it's machine parseable


smoyer64:
While working on #828, I found that the output of the `ls` command could be improved.  The `default` value of `OutputFormat` produces the following output (only one bug shown):

```
7daec5d open\tthis is a bug title                               \tJohn Doe       \t\n
```

This output is almost TSV compatible but would require that the space following the hash prefix be replaced by a tab.  For a human, any shell with a tab size of 2, 4 or 8 characters would still result in the output being displayed with a single space after when viewed.  It would however make machine processing of the output much simpler.

This issue would involve analyzing (and updating where needed) the following `OutputFormat`s:

- [ ] default
- [ ] ~org-mode~
- [ ] ~plain~
- [ ] ~json~
- [ ] ~compact~

smoyer64:
@MichaelMure - If this makes sense, please assign it to me and I'll take care of it before finishing #828.

smoyer64:
After looking through the code for the `ls` formatting functions, I realized that a) the `org-mode` and `json` formatters already provide machine parseable output and b) we really only need one of the three text output formatters to be "regular".  Since the `default` formatter is already using tabs for most of its field separators, changing that format makes the most sense.

Using the `default` formatter for field-oriented operations has an additional side effect as you can simple type `git but ls` without having to specify a formatter - e.g. `git bug ls -f compact`.  One caveat is that the label indicators are still right-aligned in the same field as the title.  We may want to change this in the future.

PR #842, opened 2022-08-11
Added QuietFormatter output

sioutisc:
QuietFormatter prints only the bug ID, similarly to how 'docker ls --quiet' works.
This mode is very useful for making batch changes to bugs using bash loops.
You can use the output of any query to easily batch add/remove labels, comments etc.

For example, the bash line below adds the label 'newLabel' to all bugs.
``for item in `git bug ls -f quiet`; do git bug label add $item newLabel; done``

### DIFF
--- a/commands/ls.go
+++ b/commands/ls.go
@@ -91,7 +91,7 @@ git bug ls status:open --by creation "foo bar" baz
 		"Select the sorting direction. Valid values are [asc,desc]")
 	cmd.RegisterFlagCompletionFunc("direction", completeFrom([]string{"asc", "desc"}))
 	flags.StringVarP(&options.outputFormat, "format", "f", "default",
-		"Select the output formatting style. Valid values are [default,plain,json,org-mode]")
+		"Select the output formatting style. Valid values are [default,plain,quiet,json,org-mode]")
 	cmd.RegisterFlagCompletionFunc("format",
 		completeFrom([]string{"default", "plain", "json", "org-mode"}))
 
@@ -138,6 +138,8 @@ func runLs(env *Env, opts lsOptions, args []string) error {
 		return lsOrgmodeFormatter(env, bugExcerpt)
 	case "plain":
 		return lsPlainFormatter(env, bugExcerpt)
+	case "quiet":
+		return lsQuietFormatter(env, bugExcerpt)
 	case "json":
 		return lsJsonFormatter(env, bugExcerpt)
 	case "compact":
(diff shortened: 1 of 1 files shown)


MichaelMure:
Hmmm, that might make sense but I'm not sure:
- there is already the `ls-id` command, but that list all the bugs without filtering. Your usage is in between. Maybe `ls-id` is odd and should be removed in favor of your idea. Maybe it should instead learn how to filter like `ls` does.
- not sure if "quiet" is appropriate. Maybe it works for docker but that doesn't seem very descriptive to me.

@smoyer64 thoughts?

smoyer64:
Curiously, I'm just working on `ls.go` (and the associated `ls_test.go`) as part of #831.  As part of that work, the default format would be changed slightly to be strictly TSV with the TSV content going to `stdout` and the `headings` going to `stderr`.  At that point, you could accomplish the above with:

```
for item in `git bug ls`; do cut -f1 | git bug label add $item newLabel; done
```

It's about what I would expect from *nix (leaving it to a different tool to translate the line).  That being said, I like the idea of combining `ls` and `ls-id` so that filtering is done in one place and simply adding a formatter to do that seems like a win.  A bigger win would be to simply eliminate `ls-id` with the expectation that people could `cut` fields based on `tab` separators.

As an aside, the undocumented `compact` format reduces the number of fields but wouldn't eliminate the need to use `cut`.  I will be submitting a PR to add `compact` to the usage and completion early next week.

smoyer64:
PR #851, which resolves #836, changes the `ls` command's default formatter to output TSV (see the specifics in #836).  With this change, the following simplifies the above *nix command a bit:

```
smoyer1@xps-15-7:~/git/git-bug$ for item in `git bug ls | cut -f1`; do echo $item; done | head
f509b69
8310415
0e2428a
cd6935a
3dbe3bc
73441f2
a2e9271
ef31717
bc0bf1d
1dafa62
```

MichaelMure:
@sioutisc is that enough for you?

sioutisc:
Absolutely, many thanks!

MichaelMure:
Reflecting a bit on this, it make sense to have a formatter spewing only IDs, it avoids the need to do pipe handling, and it's a kinda more natural concept: "just format ls the way you want", rather than polluting the entity namespace.

Not a fan of `-f quiet` though, let's call it `-f id`.

smoyer64:
@sioutisc - Now that #863 is merged, `git-bug ls --format id` should eliminate the need to pipe the output to `cut`.  Note that instead of printing human readable hashes (seven characters) it provides the full Git hash (64 characters).

PR #843, opened 2022-08-13
project board

MichaelMure:


smoyer64:
I started looking this over yesterday but didn't finish.  This morning, I realized that the "research" I'd promised which looked at how the GitHub and GitLab APIs represented Kanban is STILL on the computer at home.

The essence is:

```
Board <-- 1:n --> Columns <-- 1:n --> Bug (issue)
```

There shouldn't be a ton of operations changing columns on a board once, but there will be a ton of adding and moving bugs around the columns, so caching should really help there.  I had also looked at Github Projects and that's a whole different data structure (I'm assuming out-of-scope WRT Kanban).

In any case, `Board` looks like I'd expect.  My only suggestion is that `Board` might want an `open`/`closed` status like bug.

MichaelMure:
>  I had also looked at Github Projects and that's a whole different data structure (I'm assuming out-of-scope WRT Kanban).

Isn't that just a board with some form of transition rules?

> My only suggestion is that Board might want an open/closed status like bug.

You mean that the board itself should support being "closed" ?

smoyer64:
Projects can be cross-repository, so I think it's structurally similar to a board but how would the importer handle this?  I also got the impression that it was possible to add cards to columns that weren't issues but it's been a few months and I'd have to look again.  I'd be happy to be proven wrong too!

And yes, at least on GitHub you can mark boards as closed.  I can take a look at the JSON I saved when I get back to my other computer on Monday.  Better yet, I can attach the JSON to this repository for reference.

MichaelMure:
> Projects can be cross-repository, so I think it's structurally similar to a board but how would the importer handle this?

Import multiple remote repo within the same local git repo?

catskul:
Any status on this one?

fmv1992:
Is there any chance this will be done?

PR #844, opened 2022-08-13
Generalized resolvers

MichaelMure:


MichaelMure:
@smoyer64 FYI, little achievement over here:
- instead of an "identity resolver", there is now a generic resolver type, able to resolve any kind of entity. That means that entities like Board can now link to many other types easily
- one step further generalizing the cache layer
- there is now very few that `entity` knows about `Identity`, so with a bit of massaging to avoid the circular import, it should be possible to implement `Identity` with the generalized `entity/dag` framework, simplify that code and have them truly supporting concurrent edition

smoyer64:
Makes sense (now that I've noticed that the generic type is Interface - with a capital I) - If every entity is going to have a `SimpleResolver` are we also expecting that some entities will have a `ComplexResolver`?

And now that we're using generics, we'll have to be careful to differentiate between "real" generics and generalizations ;)

MichaelMure:
Not necessarily `ComplexResolver` but some resolvers are doing caching for example. And really at this point, nothing prevent from resolvers pulling from a remote storage, or even generating data on the fly. Although I'm not sure what that would be useful for :-D

smoyer64:
I was trying to be funny (unsuccessfully!) but was really trying to get a sense of what the `AlternateResolver` might be.  Fortunately, you answered my question;)

MichaelMure:
Haha yeah, humor is hard on the internet :-)

entity/dag/clock.go
```diff
@@ -18,21 +19,13 @@ func ClockLoader(defs ...Definition) repository.ClockLoader {
...
 			for _, def := range defs {
-				// we actually just need to read all entities,
-				// as that will create and update the clocks
-				// TODO: concurrent loading to be faster?
-				for b := range ReadAll(def, repo, resolver) {
-					if b.Err != nil {
-						return b.Err
-					}
-				}
+				errG.Go(func() error {
```
MichaelMure, line 24: Bonus here: parallelize and only read the necessary data to build the clocks. On my machine, speed up this operation from 210ms to 120ms.

smoyer64: 🚀

PR #852, opened 2022-08-18
move {bug,identity} to /entities, move input to /commands

MichaelMure:
De-clutter the root, make space for more entities.


MichaelMure:
Also, gqlgen exploded for some reason, so I fixed it and simplified the config. Not sure why, but gqlgen now bind to enums, but they now need to implement the specific (un)marshaler.

Issue #850, opened 2022-08-18
`git-bug comment edit` has no effect


sioutisc:
`git-bug comment edit` command seems to succeed (no error) but has no effect. This is also reflected with in termui whereby comments can be edited but there is no effect on completion of edit.

smoyer64:
I'll take a look at this tomorrow since I'm working on tests for the `commands` package.  From the description, it sounds like it might be in core (or maybe we're not replacing the cache contents?).

GlancingMind:
Sound like the same bug as in issue: #653 
A solution is currently still being worked on: #664

smoyer64:
I've created a draft PR (#855) which thus far only includes tests to help debug this issue.  A bit of test instrumentation shows the following:

```shell
--- FAIL: TestCommentEdit (0.05s)
    /home/smoyer1/git/git-bug/commands/comment_edit_test.go:29: Operation:  &{{8483adacda89defe875e59d8b34ae56575ea0cd58127863f78d502d22c4da45c 0xc0000aa810 1 1660925526 [134 222 207 70 204 193 119 53 189 1 99 168 239 163 141 220 249 82 33 16] map[] map[]} this is a bug title this is a bug message []}
    /home/smoyer1/git/git-bug/commands/comment_edit_test.go:29: Operation:  &{{61c39170f0022b695bdcc680c6dc9a34a127771fe1efdedac677f2a9b8262f9e 0xc0000aa870 3 1660925526 [111 188 226 202 118 147 68 141 182 109 58 155 187 164 95 212 76 56 235 212] map[] map[]} this is a bug comment []}
    /home/smoyer1/git/git-bug/commands/comment_edit_test.go:29: Operation:  &{{b299a64f8947595b92915c0ea9226d38fe4c204990d78019f3242e8cded1b0d0 0xc0000aa030 6 1660925526 [205 169 251 57 184 129 185 101 218 47 73 234 83 95 122 19 40 66 70 102] map[] map[]} 86418c3ad3acda989de1fe8775e590d8b3f4ae5065750ea0c2d58122786b3f78 this is an altered bug comment []}
[...]

Issue #849, opened 2022-08-17
Currate `stdout` and `stderr` in "core"


smoyer64:
As noted [a comment](https://github.com/MichaelMure/git-bug/issues/828#issuecomment-1185356084) in #828, the "core" of `git-bug` should support the CLI's need for parsable text output.  For the purposes of this issue, "core" comprises the following packages: `cache`, `entities` and `repository`.  In order to enable this behavior, the "core" should have the following traits:

- [ ] There should be no output to `stdout`
- [ ] Minimize the amount of output to `stderr`
- [ ] ~During "core" initialization, both `stdout` and `stderr` should be redirectable.~
- [ ] During "core" initialization, `stderr` should be redirectable.
- [ ] All error conditions should be returned to callers via Go `error`s without output to either `stdout` or `stderr`.

PR #855, opened 2022-08-19
test(850): add tests for `git-bug comment edit` has no effect

smoyer64:
This PR continues the work of adding unit tests to the CLI commands with the aim of solving the issue with the `git-bug comment edit` command not being recognized (#850).  These tests have a bit of logging that should be removed before this PR is merged.


MichaelMure:
>  These tests have a bit of logging that should be removed before this PR is merged.

Reminds me, I think your previous PR had some logging as well. Should we remove those? It makes the test output a bit hard to parse.

smoyer64:
> Reminds me, I think your previous PR had some logging as well. Should we remove those? It makes the test output a bit hard to parse.

The tests do indeed result in output to `stdout`/`stderr` but it's a side-effect of the tests.  Right now, you can't redirect either of those `io.Writer`s so each time I create repositories for the tests, there are two new lines written (I'm not reusing the same repository for all the tests as that will make it hard to use `t.Parallel()` in the future.

Ignoring the extra logging I mentioned above and the failed test's output, running the CLI tests results in output like this:

```shell
Building bug cache... Done.
Building bug cache... Done.
Building bug cache... Done.
Building bug cache... Done.
Building bug cache... Done.
Building bug cache... Done.
Building bug cache... Done.
Building bug cache... Done.
Building bug cache... Done.
Building bug cache... Done.
Building bug cache... Done.
```

Every time I add a test, there are two more lines and so it's been getting more annoying as more of the CLI has test coverage.  I haven't really started covering the error paths yet!  A couple days ago, I created #849 to address this in the "core" - that will result quiet (passing) tests when there's a failure or when the `-v` flag is used.

MichaelMure:
> The tests do indeed result in output to `stdout`/`stderr` but it's a side-effect of the tests.

Now that I am on a computer ... I was actually thinking about this one: https://github.com/MichaelMure/git-bug/blob/master/commands/ls_test.go#L80

Regarding this specific issue, it **might** be something I noticed a while back in #664 but hasn't been merged yet: https://github.com/MichaelMure/git-bug/pull/664/files?show-deleted-files=true&show-viewed-files=true&file-filters%5B%5D=#diff-8ad691e03338c49c8ea10285bf2a59172fc5b230f161292b92931a66c06b5f6aR275-R282

I don't remember exactly why I made that change, but it looks like the combined Id need to be resolved to a comment/operation Id first. If the Id is incorrect, the operation will silently do nothing when applying as it's considered invalid.

smoyer64:
Ugh ... you're right!  That logging should have been removed but since it's a `t.Log()` I needed to use the `v` flag to see it (unless the test fails).  Fixed in 2c2c449.

smoyer64:
From what you've described, I think the following code (to be removed from the test) should be working:

```go
	// TODO: remove this comment and everything between the "snip"
	//       comments when issue #850 is resolved.

	// ***** snip *****
	cache, err := env.env.backend.ResolveBugPrefix(bugID)
	require.NoError(t, err)
	bu, err := bug.Read(env.env.repo, cache.Id())
	require.NoError(t, err)
	for _, op := range bu.Operations() {
		t.Log("Operation: ", op)
	}
	t.Log("Compiled comments: ", bu.Compile().Comments)
	// ***** snip *****
```

Let me know if you see something that could be improved for debugging purposes!

smoyer64:
@MichaelMure - The problem described in #850 (`git-bug comment edit` has no effect) was closed by #664 so I've updated this PR to simply provide tests for the CLI `comment add` and `comment edit` commands.

MichaelMure (approved):

PR #868, opened 2022-09-08
feat: check security via `make` recipe and Github check

smoyer64:
This PR adds security checks to the Go workflow so that packages with known vulnerabilities (that are compiled into the binary) and insecure coding practices (e.g. unsafe user input) result in failed checks when commits are pushed to PRs.

This functionality can be run locally by the developer by running:

```
make secure
```

We don't really have a "developer" section in the `README.md` file (or a `DEVELOPER.md` file).  Is there somewhere we should put instructions for running this locally?


MichaelMure:
Interesting. Would you know how that compare with the github vulnerability detection?

As for the developer's doc, there is the doc folder. Improvements welcome :-)

smoyer64:
The vulnerability checks should (eventually?) be better than Github's ... there's an official Go vulnerability database and as of a few days ago, there's an official tool to check the code that's actually compiled into your binary against the database ([govulncheck](https://go.dev/blog/vuln)).

I don't believe there's an equivalent to the "insecure coding practice" checks in the Github vulnerability checker.  It's pretty specific to looking for places that unsafe user input is used without "safing".  ([GoKart](https://github.com/praetorian-inc/gokart)).

I'm not sure there's a need to replace the Github tool ... I do like being able to run checks before I commit though.

MichaelMure (approved):
FYI, there is also CodeQL (already in the CI).

smoyer64:
Yes ... I saw that CodeQL is in the workflow.  This was something I did over 0x since the `govulncheck` application was released.  I'll let you decide if we want to merge this - it might be interesting to see if the results from the three tools ever conflict.

MichaelMure:
Make sense to me :-)

PR #870, opened 2022-09-10
commands: reorg into different packages

MichaelMure:
This PR does a lot of things but can be summarized as:
- reorganize commands following a [new and consistent convention](https://github.com/MichaelMure/git-bug/pull/870/files#diff-2ebf410267425300638abdd79771ca430b935e6905901475c5da59b02bc20a73)
- move bug's command under a dedicated `bug` command to make space for more entities


MichaelMure:
TODO: we need to decide once and for all if those commands remains integrated in git (`git bug new`, `git bridge push` ...), if they are distributed as a standalone binary (for example `gb`), or some other solution.

Once decided, some commands description might need to be adjusted.

PR #873, opened 2022-09-12
feat: add check for allowed OSS licenses

smoyer64:
This addition to the `Makefile` and Github workflow breaks the build if a package is added that doesn't have an approved OSS license.  The (current) approved list can be found in `.lichen.yaml` which was populated from the list of licenses of packages that are already included in the library.

The license checks can also be run locally using `make legal`.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -56,3 +56,6 @@ jobs:
 
       - name: Check Security (vulnerable dependencies and insecure practices)
         run: make secure
+
+      - name: Check that all included packages have acceptable OSS licenses
+        run: make lint
--- a/.lichen.yaml
+++ b/.lichen.yaml
@@ -1,0 +1,21 @@
+---
+# Configuration for the Lichen software license scanner.  The list below
+# represents the licenses that are currently compiled into the git-bug
+# binary (with the exception of the GPL license which is git-bug's own
+# license and is therefore compatible.)  Licenses can be added to the
+# "allow" list using the official identifiers from the SPDX License
+# List which can be found at https://spdx.org/licenses/.
+#
+# The Lichen configuration file format allows overrides (for packages
+# where the license can't be automatically discovered) and exceptions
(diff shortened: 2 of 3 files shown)


.github/workflows/go.yml
```diff
@@ -56,3 +56,6 @@ jobs:
 
       - name: Check Security (vulnerable dependencies and insecure practices)
         run: make secure
+
+      - name: Check that all included packages have acceptable OSS licenses
```
MichaelMure, line 60: arguably that could be a separate action workflow, to run only once, cleanly separated

MichaelMure:
On a side note, the vulnerability workflow is problematic as it does block everything right now, even if the tool recognize that the vulnerable code is not called.

.github/workflows/go.yml
smoyer64: That's true for the `security` and `lint` as well (I was actually surprised the matrix build and lint steps were in one workflow.)

smoyer64:
Hmm ... I thought both `GoKart` and `govulncheck` specifically only broke the build if code could (in some path) be called.

MichaelMure:
For the vulnerabilty checks, it seems like https://github.com/kitabisa/gokart-action and https://github.com/Templum/govulncheck-action would be more appropriate as they would run independently from a PR and push problems into the vulnerability tooling of github.

smoyer64:
That sounds good ... and it doesn't preclude leaving the `make` recipes in place so you can check locally before committing.

sudoforge (changes requested):
@smoyer64 i like this idea being part of the standard `presubmit` pipeline, however, this should be refactored to check all binaries created for all platforms. would you be interested in picking this back up?

i know this is pretty old, so if you're not interested in taking action here please let me know and i'll add it to my queue.

sudoforge:

```
$ git fetch https://github.com/git-bug/git-bug.git refs/pull/873/head
$ git checkout FETCH_HEAD
```

PR #881, opened 2022-09-20
add gokart and govulncheck as periodic vulnerability detection

MichaelMure:


MichaelMure:
@smoyer64 I tried to do that but somehow gokart fail with a go module issue. No idea how to fix that.

PR #883, opened 2022-09-25
refactor(863): merge ls-id into ls

smoyer64:
This PR adds an `id` formatter to the `ls` command which effectively replaces the need for `ls-id` and has the additional benefit the `ls` command's querying and sorting features.

The `ls-id` command has been marked deprecated and slated for removal in v1.0 but in some ways it's effectively deprecated now.  The documentation for the `Command.Deprecated()` method states:

> Deprecated defines, if this command is deprecated and should print this string when used.

The output from the `ls-id` command now looks something like:

```
Command "ls-id" is deprecated, and will be removed in v1.0.  

Please use the "ls" command which allows filtering and sorting of the resulting
list of ids.  The following example would print a new-line separated list containing
the ids of all open bugs:
git-bug ls --format id --status open

d4cd95a6c51b71b97ad5462a508a72d81fc6f4930eabcb1a23ee89e8f082a1fc
c92067b364307ac29226dd35edc6c34f85994e20462c654d8e121a4ed1d1762f
d2f70381f1cf43f54237f05f5e8525f7e4c433dc70cef41e39f4591c0ec7ff20
32b57762dc08312a3ed78805c2684a0d7b296f81b5c72a330f543e3aed1ba8f5
741418352a82f78eecc49c42e0e2752a9dd57f321ef408321eacf38131b4dc0e
4f67b27bc126ba121b033d1fada7cb85a9eeea5f61bd63ce71b6f62ee6ed7e76
6dcee255ca2094c9912c5f8af31967e85cde6a405c89347ba41fe6be24f4c859
95edba1e2cb2e931e7f2f9f9124581555ee7f366979e92d54ae6773bee020beb
6dfd375043813cecbacf21d65bd43d57957a1d2f09ec7be84eea7b009445f9fc
a2e92713e225ff6350ece00503f1a6ba714cd7282ecf895650af8067a849b915
f75ff97963691db16d17ad664365740fc0fa254645d10572bf959ab01d8f5296
17b6bc144bf5ea577ef7023e16fef51bfb6747106f490516f772c0adfacbc04b
b42b4a5a8f0a69d831d81c6ec2302fae8f1dcd4a1e44bec9d25a9aa2b5b519ea
1dafa62af6e0bd0263ed77a46b1086af9d20f8d321470f0c2a11222bbd0a8f8e
6c356c962dc5a3569e3445f2a30fb0384a83e907b06a07f04cbada00c75ea7d6
```

Anyone using this command as input to an automation will therefore need to strip the first seven lines of the output before processing.  If the user has to change their automation, it probably makes sense to just switch to using `ls`.


MichaelMure (approved):
Awesome, thanks!

smoyer64:
Don't thank me yet! ... all the recent builds are failing due to the security checks.  The vulnerability that's blocking the build is GO-2022-0969 which in normal `git-bug` operation, would be hard to exploit (the `webui` runs on localhost by default).

This problem is fixed in Go 1.19.1 and in Go 1.18.6 but the Github action is declares Go 1.18.x and is executing Go 1.18.5 - any idea why the action isn't using the latest version of Go 1.18?  If not, I'll dig into it after church (it's raining here today - no yardwork.)

MichaelMure:
Yeah ... At this point I think we should remove those vuln checks from the CI, and move to the periodic github actions thing that push to Github's security thing. And do so if we actually get a good signal/noise (that is, if it does ignore the vulnerabilities that are not compiled in the final binary).

smoyer64:
@MichaelMure I think we agree - it's been bugging me that a vulnerability could be found well before we actually discover it.  Does #884 properly describe what we want?

Issue #530, opened 2021-01-24
CLI namespace revamp


MichaelMure:
I've been thinking about dropping out of the git porcelain CLI integration (meaning  that `git bug xxx` would stop working, in favor of directly using `foo xxx`). This integration is fun but actually doesn't bring much positive to the table. It does though bring possible confusion and, more importantly, way longer commands to type.

Additionally, when future new entities get added (say, pull-requests or config), we'll need to move all bug related command to their own namespace. That will make the commands even longer and repetitive (`git bug bug add`).

I've been thinking about the following organization:

 ```
 IDENTITY
     gb user [USER-ID] [flags]
     gb user adopt USER-ID
     gb user create
     gb user ls [flags]
 
 BUG
     gb bug add [flags]
     gb bug comment [ID]
     gb bug comment add [ID] [flags]
     gb bug label [ID]
     gb bug label add [ID] LABEL...
     gb bug label ls
     gb bug label rm [ID] LABEL...
     gb bug ls [QUERY] [flags]
     gb bug ls-id [PREFIX]
     gb bug rm ID
     gb bug show [ID] [flags]
     gb bug status [ID]
     gb bug status close [ID]
     gb bug status open [ID]
     gb bug title [ID]
     gb bug title edit [ID] [flags]
 
     gb bug select ID
     gb bug deselect    
 
 PR
     gb pr add
     gb pr show
     gb pr status
     ...
 
     gb pr select ID
     gb pr deselect    
 
 BRIDGE
     gb bridge
     gb bridge auth
     gb bridge auth add-token [TOKEN] [flags]
     gb bridge auth rm ID
     gb bridge auth show
     gb bridge configure [flags]
     gb bridge pull [NAME] [flags]
     gb bridge push [NAME]
     gb bridge rm NAME
 
 REMOTE
     gb pull [REMOTE]
     gb push [REMOTE]
 
 UIs
     gb termui
     gb webui [flags]
 
 UTILITY
     gb commands [flags]
     gb version [flags]
 ```

GlancingMind:
:+1: from me. I would also propose a slightly different version.
Currently the call format is `Binary - Entity - Operation`

MichaelMure:
@GlancingMind what's your proposal?

GlancingMind:
Accidentally hit `comment` without everything written. :)

GlancingMind:
[...]

PR #885, opened 2022-09-25
feat(530): make label commands consistent

smoyer64:
The discussion in #530 surrounding reorganizing the `git-bug` commands around the various entities that are defined (bug, board, pr, user, etc) isn't finished but the idea of changing the command to `git-bug label ls`list the valid labels to `git-bug label ls` still has merit:

This form is consistent with how the other sub-commands are structure and eliminates the last hyphenated command.  Whether or not, #530 moves forward, I think this syntax makes sense.

Note that the `git-bug ls-label` command has been deprecated in the same way as `ls-id` (with a notice that it will be removed for v1.0.

### DIFF
--- a/commands/label.go
+++ b/commands/label.go
@@ -20,13 +20,14 @@ func newLabelCommand() *cobra.Command {
 	}
 
 	cmd.AddCommand(newLabelAddCommand())
+	cmd.AddCommand(newLabelLsCommand())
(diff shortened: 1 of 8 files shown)


MichaelMure (approved):
Make a lot of sense.

smoyer64:
And if we move forward with #530, we can move the `root` commands to the `bug` sub-command!

PR #891, opened 2022-10-01
feat(849): currate stdout and stderr in "core"

### LINKED ISSUE #849: Currate `stdout` and `stderr` in "core"

smoyer64:
Resolves #849
References #828

### DIFF
--- a/api/graphql/graphql_test.go
+++ b/api/graphql/graphql_test.go
@@ -1,6 +1,7 @@
 package graphql
 
 import (
+	"bytes"
 	"testing"
 
 	"github.com/99designs/gqlgen/client"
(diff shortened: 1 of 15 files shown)


smoyer64:
What should we do with this line?

https://github.com/MichaelMure/git-bug/blob/aa23d8de5d8b3e044ef050163c8ad5da604ae90d/repository/gogit.go#L902

Right now, there is no other output to `stdout` and `stderr` in the `repository` package.  As such, `stderr` is redirectable in the `cache` package but not in the `repository` package.  If this line should be output to the user, I'll need to change the constructors in the `GoGitRepo` type to finish this PR.

MichaelMure:
> What should we do with this line?
> 
> https://github.com/MichaelMure/git-bug/blob/aa23d8de5d8b3e044ef050163c8ad5da604ae90d/repository/gogit.go#L902

I think we can just ignore that entirely. `EraseFromDisk()` is part of the private `repoTest` interface, which is only in tests.

Issue #676, opened 2021-05-19
Command to remove git-bug data from git repository


wavexx:
At one point I was using an old version of git-bug on a repository that I stopped updating for a while. Today I wanted to use it again, and git-bug complained about the incompatible metadata format.

However I didn't want to upgrade the existing data, I just wanted to re-create the metadata from scratch.

I actually had an hard time finding out all the places/refs that one needed to remove before git-bug was happy to initialize a new repository.

I don't think a "remove from repository" command is necessary, but it would be nice to have an overview and/or list of commands that can be used to remove all refs, dirs and config entries used by git-bug from an existing repository.

To account for all past versions, so far these seem to be the following refs: ``refs/bugs/ refs/identities/ refs/remotes/*/bugs/``, the "git-bug[.*]" entries from ``.git/config``, the ``.git/git-bug directory``.. (did I miss anything?)

GlancingMind:
The Makefile does contain two targets ([clean local bugs](https://github.com/MichaelMure/git-bug/blob/master/Makefile#L45-L48) and [clean local identities](https://github.com/MichaelMure/git-bug/blob/master/Makefile#L53-L56)) which seem to remove the "not so obvious" directories. Though, they don't cleanup the `.git/config` and the `.git/git-bug` directory and a makefile is not particularly user friendly. So yeah, this is more a workaround then a proper solution.

stefnotch:
> I don't think a "remove from repository" command is necessary

Today I was searching for this after the migration, since not everything worked as intended. I'd greatly appreciate having such a command that lets me completely remove git-bug from a repository and then I can safely try again.

MichaelMure:
Arguably, that could be a dedicated CLI subcommand.

zinderic:
I see that the make targets are using `git` commands to clean remote bugs and identities which might be non-trivial to implement without shelling out the commands or printing them for user consideration.

I can give this a shot if I know how that should be handled.

MichaelMure:
@zinderic sure, it seems like there is a consensus that it makes sense. From what I've seen, it's also a very common question/need, especially for newcomers.

I'm not so sure about the naming though, `eject`? `cleanup`? It doesn't fully feels right.

Now for the design:
- in [`entity_actions.go`](https://github.com/MichaelMure/git-bug/blob/5511c230b678a181cc596238bf6669428d1b1902/entity/dag/entity_actions.go) there is already a `Remove()`, we can have a `RemoveAll()`
[...]

PR #933, opened 2022-11-26
add cleanup sub-command that remove local bugs and identities

zinderic:
Related to https://github.com/MichaelMure/git-bug/issues/676

I've added subcommand called "cleanup" which removes local bugs and identities.

### DIFF
--- a/entity/dag/entity_actions.go
+++ b/entity/dag/entity_actions.go
@@ -258,3 +258,19 @@ func Remove(def Definition, repo repository.ClockedRepo, id entity.Id) error {
 
 	return nil
 }
+
+// RemoveAll delete all Entity matching the Definition.
+// RemoveAll is idempotent.
+func RemoveAll(def Definition, repo repository.ClockedRepo) error {
+	localIds, err := ListLocalIds(def, repo)
+	if err != nil {
+		return err
+	}
+	for _, id := range localIds {
+		err = Remove(def, repo, id)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
--- a/entity/dag/entity_actions_test.go
+++ b/entity/dag/entity_actions_test.go
@@ -406,3 +406,34 @@ func TestRemove(t *testing.T) {
 	err = Remove(def, repoA, e.Id())
 	require.NoError(t, err)
 }
+
+func TestRemoveAll(t *testing.T) {
+	repoA, _, _, id1, _, resolvers, def := makeTestContextRemote(t)
+
+	var ids []entity.Id
+
+	for i := 0; i < 10; i++ {
+		e := New(def)
+		e.Append(newOp1(id1, "foo"))
+		require.NoError(t, e.Commit(repoA))
+		ids = append(ids, e.Id())
+	}
+
+	_, err := Push(def, repoA, "remote")
+	require.NoError(t, err)
+
+	err = RemoveAll(def, repoA)
+	require.NoError(t, err)
+
+	for _, id := range ids {
+		_, err = Read(def, wrapper, repoA, resolvers, id)
+		require.Error(t, err)
+
+		_, err = readRemote(def, wrapper, repoA, resolvers, "remote", id)
+		require.Error(t, err)
+	}
+
+	// Remove is idempotent
+	err = RemoveAll(def, repoA)
+	require.NoError(t, err)
+}
--- a/cache/bug_subcache.go
+++ b/cache/bug_subcache.go
@@ -38,6 +38,7 @@ func NewRepoCacheBug(repo repository.ClockedRepo,
 		ReadWithResolver:    bug.ReadWithResolver,
 		ReadAllWithResolver: bug.ReadAllWithResolver,
 		Remove:              bug.Remove,
+		RemoveAll:           bug.RemoveAll,
 		MergeAll:            bug.MergeAll,
 	}
 
--- a/cache/identity_subcache.go
+++ b/cache/identity_subcache.go
@@ -39,7 +39,8 @@ func NewRepoCacheIdentity(repo repository.ClockedRepo,
 		ReadAllWithResolver: func(repo repository.ClockedRepo, resolvers entity.Resolvers) <-chan entity.StreamedEntity[*identity.Identity] {
 			return identity.ReadAllLocal(repo)
 		},
-		Remove: identity.RemoveIdentity,
+		Remove:    identity.Remove,
+		RemoveAll: identity.RemoveAll,
 		MergeAll: func(repo repository.ClockedRepo, resolvers entity.Resolvers, remote string, mergeAuthor identity.Interface) <-chan entity.MergeResult {
 			return identity.MergeAll(repo, remote)
 		},
--- a/cache/repo_cache.go
+++ b/cache/repo_cache.go
@@ -32,6 +32,7 @@ type cacheMgmt interface {
 	Load() error
 	Build() error
 	SetCacheSize(size int)
+	RemoveAll() error
 	MergeAll(remote string) <-chan entity.MergeResult
 	GetNamespace() string
 	Close() error
--- a/cache/repo_cache_common.go
+++ b/cache/repo_cache_common.go
@@ -3,12 +3,12 @@ package cache
 import (
 	"sync"
 
-	"github.com/go-git/go-billy/v5"
 	"github.com/pkg/errors"
 
 	"github.com/MichaelMure/git-bug/entities/identity"
 	"github.com/MichaelMure/git-bug/entity"
 	"github.com/MichaelMure/git-bug/repository"
+	"github.com/MichaelMure/git-bug/util/multierr"
 )
 
 func (c *RepoCache) Name() string {
@@ -56,7 +56,7 @@ func (c *RepoCache) GetRemotes() (map[string]string, error) {
 }
 
 // LocalStorage return a billy.Filesystem giving access to $RepoPath/.git/git-bug
-func (c *RepoCache) LocalStorage() billy.Filesystem {
+func (c *RepoCache) LocalStorage() repository.LocalStorage {
 	return c.repo.LocalStorage()
 }
 
@@ -80,6 +80,15 @@ func (c *RepoCache) Fetch(remote string) (string, error) {
 
 	// fetch everything at once, to have a single auth step if required.
 	return c.repo.FetchRefs(remote, prefixes...)
+}
+
+// RemoveAll deletes all entities from the cache and the disk.
+func (c *RepoCache) RemoveAll() error {
+	var errWait multierr.ErrWaitGroup
+	for _, mgmt := range c.subcaches {
+		errWait.Go(mgmt.RemoveAll)
+	}
+	return errWait.Wait()
 }
 
 // MergeAll will merge all the available remote bug and identities
@@ -160,6 +169,19 @@ func (c *RepoCache) SetUserIdentity(i *IdentityCache) error {
 
 	c.userIdentityId = i.Id()
 
+	return nil
+}
+
+func (c *RepoCache) ClearUserIdentity() error {
+	c.muUserIdentity.Lock()
+	defer c.muUserIdentity.Unlock()
+
+	err := identity.ClearUserIdentity(c.repo)
+	if err != nil {
+		return err
+	}
+
+	c.userIdentityId = ""
 	return nil
 }
 
--- a/cache/repo_cache_test.go
+++ b/cache/repo_cache_test.go
@@ -135,6 +135,39 @@ func TestCache(t *testing.T) {
 	_, err = cache.Bugs().ResolvePrefix(bug1.Id().String()[:10])
 	require.NoError(t, err)
 
+	require.Len(t, cache.bugs.cached, 1)
+	require.Len(t, cache.bugs.excerpts, 2)
+	require.Len(t, cache.identities.cached, 1)
+	require.Len(t, cache.identities.excerpts, 2)
+	require.Equal(t, uint64(2), indexCount(t, identity.Namespace))
+	require.Equal(t, uint64(2), indexCount(t, bug.Namespace))
+
+	// Remove + RemoveAll
+	err = cache.Identities().Remove(iden1.Id().String()[:10])
+	require.NoError(t, err)
+	err = cache.Bugs().Remove(bug1.Id().String()[:10])
+	require.NoError(t, err)
+	require.Len(t, cache.bugs.cached, 0)
+	require.Len(t, cache.bugs.excerpts, 1)
+	require.Len(t, cache.identities.cached, 0)
+	require.Len(t, cache.identities.excerpts, 1)
+	require.Equal(t, uint64(1), indexCount(t, identity.Namespace))
+	require.Equal(t, uint64(1), indexCount(t, bug.Namespace))
+
+	_, err = cache.Identities().New("René Descartes", "rene@descartes.fr")
+	require.NoError(t, err)
+	_, _, err = cache.Bugs().NewRaw(iden2, time.Now().Unix(), "title", "message", nil, nil)
+	require.NoError(t, err)
+
+	err = cache.RemoveAll()
+	require.NoError(t, err)
+	require.Len(t, cache.bugs.cached, 0)
+	require.Len(t, cache.bugs.excerpts, 0)
+	require.Len(t, cache.identities.cached, 0)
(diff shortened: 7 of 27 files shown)


entity/dag/entity_actions.go
```diff
@@ -258,3 +257,23 @@ func Remove(def Definition, repo repository.ClockedRepo, id entity.Id) error {
...
 }
+
+// RemoveAll delete all Entity items for all identities
+func RemoveAll(def Definition, repo repository.ClockedRepo) error {
+	localIds, err := ListLocalIds(def, repo)
+	if err != nil {
+		return err
+	}
+	for _, id := range localIds {
+		fmt.Println("removing bug " + id.String())
```
MichaelMure, line 268: You can't have anything printing from the core. If some information is worthy enough, that should be bubbled up in the form of an event, and each UI can elect to display it in the form that make sense.
Here, I think we can just remove it.

entity/dag/entity_actions_test.go
```diff
@@ -406,3 +406,36 @@ func TestRemove(t *testing.T) {
...
+					FormatVersion:        0,
+				},
+				repo: repository.NewMockRepo(),
+			},
+			wantErr: false,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if err := RemoveAll(tt.args.def, tt.args.repo); (err != nil) != tt.wantErr {
```
MichaelMure, line 436: Unless I'm missing something, you are executing `RemoveAll` on an empty repo. Arguably that's not testing very much.

commands/cleanup.go
```diff
@@ -0,0 +1,43 @@
...
+			return runCleanup(env)
+		}),
+	}
+
+	return cmd
+}
+
+func runCleanup(env *execenv.Env) error {
+
+	fmt.Println("cleaning bug items..")
```
MichaelMure, line 31: "cleaning bug items.." --> "cleaning bugs ..."
Plus, you should use `env.Out`

commands/cleanup.go
```diff
@@ -0,0 +1,43 @@
...
+	"github.com/spf13/cobra"
+
+	"github.com/MichaelMure/git-bug/commands/execenv"
+)
+
+func newCleanUpCommand() *cobra.Command {
+	env := execenv.NewEnv()
+
+	cmd := &cobra.Command{
+		Use:     "cleanup",
```
MichaelMure, line 17: Let's do `wipe`: https://github.com/MichaelMure/git-bug/issues/676#issuecomment-1329077325

entity/dag/entity_actions_test.go
MichaelMure: Take example on `TestRemove` in the same file to generate some data ?

MichaelMure (changes requested):
Not too bad really, but need some adjustment. 
Thanks for doing this.

entity/dag/entity_actions_test.go
zinderic: Yeah, you're right. I added data at the beginning of the test.

entity/dag/entity_actions.go
zinderic: Removed.

commands/cleanup.go
zinderic: Changed text, using env.Out now.

commands/cleanup.go
zinderic: Updated, now it's `wipe`.

MichaelMure:
Looking at your PR, I realized that it should be up to the cache layer to update or clear the full-text index (Bleve), meaning that those commands should go through the cache, not directly to the repository.

As it turns out, I'm in the process of refactoring that cache (https://github.com/MichaelMure/git-bug/pull/938), so I'll come back to your PR when I'm done with that. Don't worry it will be merged, it just will require a bit more time.

zinderic:
Okay, thanks for letting me know.

MichaelMure:
PR rebased and adjusted:
- follow my previous changes in the cache
- also cleanup the git config (user identity, bridge ...)
- also cleanup the cache files

Thanks for your help :-)

MichaelMure:
hmm, it seems like I unearthed some issue around deleting files on windows. The CI fails with things like:
```
    repo_cache_test.go:159: 
        	Error Trace:	D:\a\git-bug\git-bug\cache\repo_cache_test.go:159
        	Error:      	Received unexpected error:
        	            	CreateFile C:\Users\RUNNER~1\AppData\Local\Temp\TestCache1445993478\001\.git\refs\bugs\51932f5d173e9965e67256a5deb8300e91b70f4b1f029331268f66088289086a: The system cannot find the file specified.
        	Test:       	TestCache
        	```

The code do looks good though, including in go-git where things seem to be idempotent ...

**Comment by zinderic** [comment]
Yeah, I don't think it's specific to Windows. I don't have access to a Windows right now but I managed to reproduce the error on my Mac:

```
kstaykov:git-bug/ (feat-cleanup-subcommand✗) $ go test -v -run TestCache -count=1 ./cache                                                      [16:31:32]
=== RUN   TestCache
    repo_cache_test.go:159: 
                Error Trace:    /Users/kstaykov/proj/git-bug/cache/repo_cache_test.go:159
                Error:          Received unexpected error:
                                open /var/folders/d6/y7xbyvtj337f8njxpkq7kk9c0000gn/T/TestCache3212261930/001/.git/refs/bugs/e87867963cb6e73d40fe9ab0cee9942d1827839610dbbf76019ae395e30c4c41: no such file or directory
                Test:           TestCache
--- FAIL: TestCache (0.24s)
=== RUN   TestCachePushPull
--- PASS: TestCachePushPull (0.41s)
=== RUN   TestCacheEviction
--- PASS: TestCacheEviction (0.14s)
FAIL
FAIL    github.com/MichaelMure/git-bug/cache    1.262s
FAIL
kstaykov:git-bug/ (feat-cleanup-subcommand✗) $  
```

It's not failing every time but it fails often, about half the time. I still don't have much luck identifying the cause but it appears like some sort of race condition that is trying to access the file before it is there. It feels like the file is being created by the logic but it was not synced to the disk yet and unavailable for read at the time it is being accessed. But that's just a theory right now.

**Comment by zinderic** [comment]
I see what's going on. It's trying to remove `Identity` and `Bug` subcaches. Since that's done using multierr.ErrWaitGroup it will run it in goroutine and therefore there is no guaranteed order for which one would go first. I don't think that's a problem but when they go concurrently there is a chance of the cache becoming "broken" or in a state where it will be rebuilt.

I tried patching my function to not use multierr.ErrWaitGroup and go sequentially and return first error. That way the test no longer fails even when executed many times. But if I make that patch we might be hiding a problem with the cache. What do you think?

**Comment by MichaelMure** [comment]
So you are suggesting that there is a concurrent access to something somewhere? If so that's indeed a bug, and concurrently calling `RemoveAll()` on different entities should work. `multierr.ErrWaitGroup` should stay.

**Comment by zinderic** [comment]
Yes, there is a race condition that is not detectable in the current test for some reason but when I make a new test that adds 2 identities and 2 bugs and tries to concurrently remove them the go tooling is able to detect the race condition. Here's the test:

```
func TestCacheConcurrent(t *testing.T) {
	repo := repository.CreateGoGitTestRepo(t, false)
	cache, err := NewRepoCacheNoEvents(repo)
	require.NoError(t, err)
	// Create, set and get user identity
	iden1, err := cache.Identities().New("Kaz", "kaz@ketterdam.grisha")
	require.NoError(t, err)
	err = cache.SetUserIdentity(iden1)
	require.NoError(t, err)
	iden2, err := cache.Identities().New("Inej", "inej@ketterdam.grisha")
	require.NoError(t, err)

	bug1, _, err := cache.Bugs().New("title", "message")
	require.NoError(t, err)
	bug2, _, err := cache.Bugs().New("title", "message")
	require.NoError(t, err)

	go func() {
		cache.Identities().Remove(iden1.Id().String()[:10])
		cache.Identities().Remove(iden2.Id().String()[:10])
		cache.Bugs().Remove(bug1.Id().String()[:10])
		cache.Bugs().Remove(bug2.Id().String()[:10])
	}()

	require.NoError(t, err)
}
```

This test will reliably fail and the race condition is now detectable:

```
kstaykov:git-bug/ (feat-cleanup-subcommand✗) $ go test -race -v -run TestCacheConcurrent -count=1 ./cache                                      [10:49:18]
=== RUN   TestCacheConcurrent
==================
WARNING: DATA RACE
Read at 0x00c000423d10 by goroutine 29:
  runtime.mapaccess1_faststr()
      /opt/homebrew/Cellar/go/1.19.3/libexec/src/runtime/map_faststr.go:13 +0x41c
  github.com/MichaelMure/git-bug/repository.(*GoGitRepo).GetIndex()
      /Users/kstaykov/proj/git-bug/repository/gogit.go:330 +0xac
  github.com/MichaelMure/git-bug/repository.(*replaceKeyring).GetIndex()
      <autogenerated>:1 +0x5c
  github.com/MichaelMure/git-bug/cache.(*SubCache[...]).Remove()
      /Users/kstaykov/proj/git-bug/cache/subcache.go:395 +0x2f8
  github.com/MichaelMure/git-bug/cache.TestCacheConcurrent.func1()
      /Users/kstaykov/proj/git-bug/cache/repo_cache_test.go:186 +0xe8

Previous write at 0x00c000423d10 by goroutine 11:
  runtime.mapassign_faststr()
      /opt/homebrew/Cellar/go/1.19.3/libexec/src/runtime/map_faststr.go:203 +0x42c
  github.com/MichaelMure/git-bug/repository.(*GoGitRepo).Close()
      /Users/kstaykov/proj/git-bug/repository/gogit.go:225 +0xa4
  github.com/MichaelMure/git-bug/repository.CreateGoGitTestRepo.func1()
      /Users/kstaykov/proj/git-bug/repository/gogit_testing.go:35 +0x2c
  testing.(*common).Cleanup.func1()
      /opt/homebrew/Cellar/go/1.19.3/libexec/src/testing/testing.go:1041 +0x13c
  testing.(*common).runCleanup()
      /opt/homebrew/Cellar/go/1.19.3/libexec/src/testing/testing.go:1210 +0xe4
  testing.tRunner.func2()
      /opt/homebrew/Cellar/go/1.19.3/libexec/src/testing/testing.go:1440 +0x48
  runtime.deferreturn()
      /opt/homebrew/Cellar/go/1.19.3/libexec/src/runtime/panic.go:476 +0x30
  testing.(*T).Run.func1()
      /opt/homebrew/Cellar/go/1.19.3/libexec/src/testing/testing.go:1493 +0x40

Goroutine 29 (running) created at:
  github.com/MichaelMure/git-bug/cache.TestCacheConcurrent()
      /Users/kstaykov/proj/git-bug/cache/repo_cache_test.go:185 +0x3d8
  testing.tRunner()
      /opt/homebrew/Cellar/go/1.19.3/libexec/src/testing/testing.go:1446 +0x188
  testing.(*T).Run.func1()
      /opt/homebrew/Cellar/go/1.19.3/libexec/src/testing/testing.go:1493 +0x40

Goroutine 11 (running) created at:
  testing.(*T).Run()
      /opt/homebrew/Cellar/go/1.19.3/libexec/src/testing/testing.go:1493 +0x55c
  testing.runTests.func1()
      /opt/homebrew/Cellar/go/1.19.3/libexec/src/testing/testing.go:1846 +0x90
  testing.tRunner()
      /opt/homebrew/Cellar/go/1.19.3/libexec/src/testing/testing.go:1446 +0x188
  testing.runTests()
      /opt/homebrew/Cellar/go/1.19.3/libexec/src/testing/testing.go:1844 +0x6c0
  testing.(*M).Run()
      /opt/homebrew/Cellar/go/1.19.3/libexec/src/testing/testing.go:1726 +0x870
  main.main()
      _testmain.go:59 +0x2fc
==================
    testing.go:1319: race detected during execution of test
--- FAIL: TestCacheConcurrent (0.16s)
=== CONT  
    testing.go:1319: race detected during execution of test
FAIL
FAIL    github.com/MichaelMure/git-bug/cache    0.762s
FAIL
kstaykov:git-bug/ (feat-cleanup-subcommand✗) $ 
```

**Comment by MichaelMure** [comment]
@zinderic I looked at it and the race you found is because you don't wait for your goroutine in your test, so you have `Remove()` being called at the same time as `Close()` on the cache (due to reaching the end of the test). Arguably `Close()` could lock the mutexes, but that wouldn't really help as the cache would be in a bad state after `Close()` anyway, you are not supposed to do that.

Point is, I don't think it's the problem that the CI found. It looks more like a race at the filesystem level. @smoyer64 speculated that it might be related to https://cs.opensource.google/go/go/+/refs/tags/go1.19.4:src/testing/testing.go;drc=38cfb3be9d486833456276777155980d1ec0823e;l=1229

**Comment by zinderic** [comment]
I checked the link and the comments suggest something specific to Windows. But if it's specific to Windows then why the test fails also on MacOS if run enough times? Running 10 tests like that has about 50-60% error rate on my laptop:

```
go test -race -v -run TestCache -count=10 ./cache
```

In other words 5 or 6 out of 10 fail when run with -count=10. Running them one at a time seems to produce less failures but it still has significant chance of failing. Those observations were made on Mac Air M2 so I don't think that's specific to just Windows.

**Comment by smoyer64** [comment]
This might be related to #809 - running tests in parallel should work since the test environments are built individually and theoretically isolated (and use different temporary directories).  If there's a concurrency issue with the code-under-test, that might explain a bit more.  Intermittent failures are still occurring on the Windows build (one happened this morning).

**Comment by zinderic** [comment]
Looks like a different test fails now - TestGoGitRepo but it fails in similar way as before. This time I cannot reproduce the failure on MacOS.

**Comment by smoyer64** [comment]
I'll take a look at this after work today - if we track enough of these down, then `t.Parallel()` should eventually work (with the obvious exclusions (see `t.Setenv()`, etc.)

**Comment by smoyer64** [comment]
Even the test that I thought I fixed fails intermittently again - I wonder if there's something in the refactoring.  At least the more recent failure gives a better reason (a locked file handle).

**Comment by MichaelMure** [comment]
> I wonder if there's something in the refactoring

The refactor and this PR add more file or folder deletion. I think it's the reason why we see those issues more now. Of course it's possible there is actually an issue in the code, but it's not obvious to me what it would be. It seems more compelling to me that it would be a variation of https://cs.opensource.google/go/go/+/refs/tags/go1.19.4:src/testing/testing.go;drc=38cfb3be9d486833456276777155980d1ec0823e;l=1229

**Comment by smoyer64** [comment]
> Once the test is fixed

Except we're not leaving files behind any more - and the error message is now about being denied access to a file.  I guess I'm going to have to boot into my Windows partition and see what's actually happening.

**Comment by MichaelMure** [comment]
Still happening:
```
    repo_cache_test.go:163: 
        	Error Trace:	/home/runner/work/git-bug/git-bug/cache/repo_cache_test.go:163
        	Error:      	Received unexpected error:
        	            	stat /tmp/TestCache162481194/001/.git/refs/bugs/fd954a37b9dc52742fbb2e2c60f62fc6ec0df85a2a2423b2b7ee5f1b8e4fad62: no such file or directory
        	Test:       	TestCache
```

MichaelMure:
That looks like a `os.PathError` or `fs.PathError` (they are the same type).

theory:
- `TestCache` create a bunch of identities and bugs
- `TestCache` call `Remove` on one identity and one bug, which in turn call `repo.RemoveRef(ref)` and `GoGitRepo.RemoveRef` <-- Let's assume this deletion is not immediate
- `TestCache` call `RemoveAll`, which create a list of refs to remove with `ListLocalIds`, which call `GoGitRepo.ListRefs`, and later call `GoGitRepo.RemoveRef` to remove those <-- this is where things fail

If somehow go-git is racing on listing+removal of git refs, it's possible that `RemoveAll` get a ref that is already being removed, and later would fail when trying to delete it. go-git is subject to race in other places so that wouldn't be unheard of.

Another issue with go-git, is that removing a ref should ideally be idempotent (aka not  fail if the ref is gone already), but that doesn't seem to be the case.

smoyer64:
Interesting ... and that sounds a lot like what I was going to go spelunking for in `go-git`.  I think something is racing and I was going to put a xx mS delay before the call just to see if maybe the OS is still in the process of closing the file when the test tries to delete it.

MichaelMure:
I found the issue. Turns out, that was not at all the right place: https://github.com/go-git/go-git/pull/659

MichaelMure:
https://github.com/go-git/go-git/pull/659 might not be perfect as it triggered the CI there, but does solve our problem. Problem which isn't related to that PR, so I'll merge and we can follow up later with go-git.

PR #936, opened 2022-11-28
feat(935): disambiguate config errors

smoyer64:


repository/gogit_config.go
```diff
@@ -119,32 +119,32 @@ func (cr *goGitConfigReader) ReadString(key string) (string, error) {
...
 	switch {
 	case len(split) == 2:
 		optionName := split[1]
 		if !section.HasOption(optionName) {
-			return "", ErrNoConfigEntry
+			return "", fmt.Errorf("%w: missing key %s", ErrNoConfigEntry, key)
 		}
 		if len(section.OptionAll(optionName)) > 1 {
-			return "", ErrMultipleConfigEntry
+			return "", fmt.Errorf("%w: duplicated key %s", ErrMultipleConfigEntry, key)
```
MichaelMure, line 133: hmmm, would it be better to have those errors be `struct { key string }` and carry that formatting in the `Error() string` function? That would avoid repeating all that and silly mistakes.

smoyer64: Well - I wished that's how Go error wrapping worked ... do we want these two errors to be custom?

MichaelMure: I think that would make sense. Error wrapping is I think meant to happen when one layer bubble up an error from a lower level, not the initial layer adding more information.

smoyer64: Error wrapping isn't so dogmatic so long as `Unwrap()`, `Is()` and `As()` do the right things.  If you turn on all the linters in `golangci-lint`, it will complain about both unwrapped `errors` from external packages and about both types of "dynamic errors".  In any case, I fixed the problem with "repeating all that" since it could indeed lead to silly mistakes (and of course make changing the format in the future a lot more work).

PR #946, opened 2022-12-09
Implement comment deletion feature

### LINKED ISSUE #939: webui: ability to delete a comment
I would like to "delete" a comment.

In the sense of marking it as deleted, so that in it's place, i would see "Comment deleted by Name" rather than the content.

If i try to edit the comment and remove it's text, i get "Empty message, aborting" in termui.

If i do the same in the webui, the "update comment" button gets grayed out.

Currently I work around this by editing comment and replacing content with ".".

mendelmaleh:
Fixes #939, only in termui for now. I mostly followed the edit implementation, let me know what can be improved.

### DIFF
--- a/entities/bug/timeline.go
+++ b/entities/bug/timeline.go
@@ -21,6 +21,7 @@ type CommentHistoryStep struct {
 	Author identity.Interface
 	// The new message
 	Message  string
+	Deleted  bool
 	UnixTime timestamp.Timestamp
 }
 
@@ -29,6 +30,7 @@ type CommentTimelineItem struct {
 	combinedId entity.CombinedId
 	Author     identity.Interface
 	Message    string
+	Deleted    bool
 	Files      []repository.Hash
 	CreatedAt  timestamp.Timestamp
 	LastEdit   timestamp.Timestamp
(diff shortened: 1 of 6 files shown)


MichaelMure:
hey @mendelmaleh, sorry for the delay. Large changes happened in git-bug core, and we are currently trying to catch bugs that came out of it. I'll get to your PR at some point.

entities/bug/timeline.go
```diff
@@ -74,6 +78,10 @@ func (c *CommentTimelineItem) Edited() bool {
 	return len(c.History) > 1
 }
 
+func (c *CommentTimelineItem) DeletedBy() identity.Interface {
+	return c.History[len(c.History)-1].Author
```
MichaelMure, line 82: That doesn't seem robust enough. What if something else happen after deleting? At the moment, it seems like only the local termui would prevent that. Going further, someone running another version of git-bug might make more change on that deleted comment.

We also need to decide what is the semantic to enforce here. If a comment is deleted, do we make no-op any further change? Do we accept them and gracefully handle that (as in, AddComment+DeleteComment+EditComment+UndeleteComment would show the modified comment). I'd go for the latter.

MichaelMure (changes requested):
Pretty good code quality, thanks.
We need to figure out how to deal with the problem in my comment though. I can't merge like that as it could lead to issues.

MichaelMure:
Not stale

sudoforge:

```
$ git fetch https://github.com/git-bug/git-bug.git refs/pull/946/head
$ git checkout FETCH_HEAD
```

Issue #934, opened 2022-11-27
0.8.0 issue with Authentication


idjohnson:
I setup 0.8.0 (Linux amd64) in WSL running Ubuntu 20.04 (focal) and git 2.25.1

I tried with an Org GH Repo, Personal GH repo and a freshly pulled AzDO repo.

```
$ git bug pull
Fetching remote ...
Error: authentication required
```

I tried with 0.7.2 without issue.

I even tried migrating a working repo then using 0.8.0 and got the error again
```
...snip
Migration #3
Purpose: Make bug and identities independent from the storage by making the ID generation self-contained. Also, migrate to the new full DAG entity data model.

identity 1b5c437: migrated to 73519a0
bug b3d0ba7: migrated to 025f7c3


Done!
builder@DESKTOP-72D2D9T:~/Workspaces/testgitbug$ git bug ls
A lock file is present but the corresponding process is not, removing it.
Building identity cache... Done.
Building bug cache... Done.
025f7c3 open    Bug Title 1                                             Isaac Johnson
builder@DESKTOP-72D2D9T:~/Workspaces/testgitbug$ git bug pull
Fetching remote ...
Error: authentication required
```

smoyer64:
Is your SSH key defined in WSL2 and added to GitHub or is your SSH key in Windows.  If it's the latter, you have to configure your WSL2 environment by running the `ssh-agent`.  If you want to share your signing keys between Windows and WSL2, you also have to set up your system to serve the PGP keys from Windows and use them from WSL2.

idjohnson:
This is just a WIP writeup that covers all the steps I tried - windows, wsl, pre-compiled and compiled from source: http://freshbrewed-test.s3-website-us-east-1.amazonaws.com/2022/12/13/gitbug.html

idjohnson:
I tried: WSL https (my usual) with GH Public, Private and Azure DevOps private repos.
I tried WSL SSH
I tried Windows HTTPS.

I tried 0.8.0 and in WSL, also freshly compiled from main.

idjohnson:
I got same error with windows and Mac and Ubuntu. Do you have and path that
works with 0.8.0 I could try?

smoyer64:
I can replicate this issue ...

devopsguiden:
[...]

PR #963, opened 2022-12-28
feat: upgrade go-git to v5.5.1

smoyer64:
Related to #934


MichaelMure:
There is also https://github.com/MichaelMure/git-bug/pull/947, but it looks like some things are breaking in git-bug.

smoyer64:
Yeah ... I didn't notice the one started by dependabot.  I'm tracking down the Go panic (it's the crypto checking the signing).

smoyer64:
We should talk about what our goals are for commit signing and use cases for those who  already have keys.  This will obviously also impact how the bridges work since ideally we'd be able to verify signed commits from  the "forges".

MichaelMure:
> We should talk about what our goals are for commit signing and use cases for those who already have keys. This will obviously also impact how the bridges work since ideally we'd be able to verify signed commits from the "forges".

Does that answer your question? https://github.com/MichaelMure/git-bug/issues/130

As for the bridges, I'd think the only interaction with keys would be to import the public keys when creating or updating the identities. That way, anyone having the matching private key would be able to adopt the identity without losing the history.

MichaelMure (approved):

smoyer64:
I'll add a few comments to #130 instead of them languishing here on a closed PR.

PR #967, opened 2022-12-29
CI: remove lint security step as it's crashing

MichaelMure:


MichaelMure:
@smoyer64 FYI, we can have it back but it's just annoying with the current state of things.

PR #976, opened 2023-01-03
fix: resolve Go vulnerabilities

### LINKED ISSUE #975: Critical security vulnerabilities in Go's standard library
Go v1.19.4 is required to resolve all the security vulnerabilities listed by `govulncheck`:

```
govulncheck ./... 
govulncheck is an experimental tool. Share feedback at https://go.dev/s/govulncheck-feedback.

Scanning for dependencies with known vulnerabilities...
Found 3 known vulnerabilities.

Vulnerability #1: GO-2022-1144
  An attacker can cause excessive memory growth in a Go server
  accepting HTTP/2 requests. HTTP/2 server connections contain a
  cache of HTTP header keys sent by the client. While the total
  number of entries in this cache is capped, an attacker sending
  very large keys can cause the server to allocate approximately
  64 MiB per open connection.

  Call stacks in your code:
      commands/webui.go:194:26: github.com/MichaelMure/git-bug/commands.runWebUI calls net/http.Server.ListenAndServe

  Found in: net/http@go1.18.6
  Fixed in: net/http@go1.19.4
  More info: https://pkg.go.dev/vuln/GO-2022-1144

Vulnerability #2: GO-2022-0969
  HTTP/2 server connections can hang forever waiting for a clean
  shutdown that was preempted by a fatal error. This condition can
  be exploited by a malicious client to cause a denial of service.

  Call stacks in your code:
      commands/webui.go:194:26: github.com/MichaelMure/git-bug/commands.runWebUI calls net/http.Server.ListenAndServe

  Found in: net/http@go1.18.6
  Fixed in: net/http@go1.19.1
  More info: https://pkg.go.dev/vuln/GO-2022-0969

Vulnerability #3: GO-2022-1039
  Programs which compile regular expressions from untrusted
  sources may be vulnerable to memory exhaustion or denial of
  service. The parsed regexp representation is linear in the size
  of the input, but in some cases the constant factor can be as
  high as 40,000, making relatively small regexps consume much
  larger amounts of memory. After fix, each regexp being parsed is
  limited to a 256 MB memory footprint. Regular expressions whose
  representation would use more space than that are rejected.
  Normal use of regular expressions is unaffected.

  Call stacks in your code:
      repository/index_bleve.go:115:28: github.com/MichaelMure/git-bug/repository.bleveIndex.Search calls github.com/blevesearch/bleve.indexImpl.Search, which eventually calls regexp/syntax.Parse

  Found in: regexp/syntax@go1.18.6
  Fixed in: regexp/syntax@go1.19.2
  More info: https://pkg.go.dev/vuln/GO-2022-1039

=== Informational ===

The vulnerabilities below are in packages that you import, but your code
doesn't appear to call any vulnerable functions. You may not need to take any
action. See https://pkg.go.dev/golang.org/x/vuln/cmd/govulncheck
for details.

Vulnerability #1: GO-2022-0988
  JoinPath and URL.JoinPath do not remove ../ path elements appended to a relative path. For example, JoinPath("https://go.dev", "../go") returns the URL "https://go.dev/../go", despite the JoinPath documentation stating that ../ path elements are removed from the result.
  Found in: net/url@go1.18.6
  Fixed in: net/url@go1.19.1
  More info: https://pkg.go.dev/vuln/GO-2022-0988
```

smoyer64:
Resolves #975

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -18,7 +18,7 @@ jobs:
     steps:
       - uses: actions/setup-go@v3
         with:
-          go-version: 1.18.x
+          go-version: 1.19.x
       - uses: actions/checkout@v3
       # Run benchmark with `go test -bench` and stores the output to a file
       - name: Run benchmark
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -18,7 +18,7 @@ jobs:
     steps:
       - uses: actions/setup-go@v3
         with:
-          go-version: 1.19.x
+          go-version: 1.19.4
       - uses: actions/checkout@v3
       # Run benchmark with `go test -bench` and stores the output to a file
       - name: Run benchmark
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -12,7 +12,7 @@ jobs:
 
     strategy:
       matrix:
-        go-version: [1.19.x]
+        go-version: [1.19.4]
         platform: [ubuntu-latest, macos-latest, windows-latest]
 
     runs-on: ${{ matrix.platform }}
@@ -20,7 +20,7 @@ jobs:
     steps:
 
       - name: Set up Go ${{ matrix.node-version }}
-        uses: actions/setup-go@v2
+        uses: actions/setup-go@v3
         with:
           go-version: ${{ matrix.go-version }}
 
@@ -44,13 +44,12 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Install Go
-        uses: actions/setup-go@v2
+        uses: actions/setup-go@v3
         with:
-          go-version: 1.19.x
+          go-version: 1.19.4
 
       - name: Checkout code
         uses: actions/checkout@v2
 
       - name: Check Code Formatting
         run: find . -name "*.go" | while read line; do [ -z "$(gofmt -d "$line" | head)" ] || exit 1; done
-
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -12,7 +12,7 @@ jobs:
 
     strategy:
       matrix:
-        go-version: [1.18.x]
+        go-version: [1.19.x]
         platform: [ubuntu-latest, macos-latest, windows-latest]
 
     runs-on: ${{ matrix.platform }}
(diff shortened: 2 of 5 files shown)


bridge/gitlab/event.go
```diff
@@ -205,7 +205,9 @@ func SortedEvents(inputs ...<-chan Event) chan Event {
 
 // getNewTitle parses body diff given by gitlab api and return it final form
 // examples: "changed title from **fourth issue** to **fourth issue{+ changed+}**"
-//           "changed title from **fourth issue{- changed-}** to **fourth issue**"
+//
+//	"changed title from **fourth issue{- changed-}** to **fourth issue**"
```
MichaelMure, line 209: looks like your IDE got a bit hungry

commands/bridge/bridge_new.go
```diff
@@ -33,7 +33,7 @@ func newBridgeNewCommand() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "new",
 		Short: "Configure a new bridge",
-		Long: `	Configure a new bridge by passing flags or/and using interactive terminal prompts. You can avoid all the terminal prompts by passing all the necessary flags to configure your bridge.`,
+		Long:  `	Configure a new bridge by passing flags or/and using interactive terminal prompts. You can avoid all the terminal prompts by passing all the necessary flags to configure your bridge.`,
```
MichaelMure, line 36: Same here, there should be no space at the beginning of the `Long` description

entities/identity/version.go
```diff
@@ -16,7 +16,7 @@ import (
 
 // 1: original format
 // 2: Identity Ids are generated from the first version serialized data instead of from the first git commit
-//    + Identity hold multiple lamport clocks from other entities, instead of just bug edit
+//   - Identity hold multiple lamport clocks from other entities, instead of just bug edit
```
MichaelMure, line 19: another IDE problem, that's not a markdown list

MichaelMure (changes requested):
Apart from the IDE problems, I'm a bit wary to update too quickly the requirements as it can make it difficult to package for distro. But alright, let see how it goes.
Note that you don't actually require golang 1.19.4 (1 month old), but just golang 1.19 (6 months old).

smoyer64:
That's really weird ... I only changed and committed the module file and three of the workflows but it failed linting and looking at my history, I ran `go fmt ./...` - perhaps it's rules are different in Go 1.19.  I reverted to 33e3e4b and only see the four expected changed files.  Now I just need to see why the linter is complaining!

PR #987, opened 2023-01-10
commands: add a nice terminal progress bar when building the cache

MichaelMure:
One issue remaining is that writing the cache takes significant time, but I don't know how to reflect that nicely to the user.


smoyer64:
Sorry I missed this ... I guess my only option is to criticize it after the fact ;)

MichaelMure:
Critique welcome

smoyer64:
I just got my master branch back in shape and can't wait to see it!

Issue #973, opened 2023-01-03
problem with gitlab push


buestad:
I have a self hosted GitLab 15.6.0-ee and I cannot push any locally created issues.
I am using the pre built 0.8.0 binary for win64

- Bridge configured correctly (I think)
- Pulled issues from project (with some errors)
- Adopted my gitlab user which comes when I pull
- Add a new bug/issue using the WebUI
- git bug ls shows the new bug

```
>git bug bridge push
exported 0 issues with default bridge
```

As mentioned, the git bug bridge pull did not run cleanly:
```
>git bug bridge pull
[8844b1e] new identity: 8844b1eb4852a9a84f497c78a93057d2f588a4c049652d062d75885b35b1f4e5
[abbefcc] new issue: abbefcc80a4fee9566b8205f324c4b2e8ee765a81118d77296eddb30605493a5
import error on entity 608: issue event creation: unexpected event
import error on entity 609: issue event creation: unexpected event
import error on entity 610: issue event creation: unexpected event
[abbefcc] new comment: a3b6b3efcbc80a54fee29566eb82085f3274c4bc2e8e4e7651a811018d7d7296
[abbefcc] changed status with op: d08663dff96760da23ec10b611f48aa5662ae54f3c1c1653602bb36bf02e5fea
[04e5722] new issue: 04e5722e69e3a4d9f32fea893bc11b7147d66efe1f76723838d82b7ed9d0b48f
[04e5722] changed status with op: 0b67543735ca90a9662d3a4c2af39a10d945c02dfc70ae602ca7ff9195ff9492
[2b81d80] new issue: 2b81d800b10d954848ecdfb05158cb58d5521562001725e29cc773a1b7e155b5
[2b81d80] new comment: 2cb0851d8f00b110d95d48489ecdf9b051158cb158d5e521576200f17253e29c
import error on entity 713: issue event creation: unexpected event
import error on entity 714: issue event creation: unexpected event
[2b81d80] changed status with op: b1c1eca65b76a98e22404db6ef791568b838539629070ca070391907f23764a5
[13abb35] new issue: 13abb3506964e26e26d1a6e80416eacdf0ec88a2a398f92425819759bb1d7a94
[13abb35] new comment: 1b3fa3bb3a5069f64e246e261d1a6de804816eabcdf0eec883a2a3098f982425
[39f55ae] new issue: 39f55aeffc6b18e7a65fb9c4401b62fbd592fb1bb5241a8b0c870b41b24d5992
[...]

PR #990, opened 2023-01-13
added EventMentionedInCommit

buestad:
Added the event **EventMentionedInCommit** in the gitlab bridge to support / not give an error on events in gitlab issue s where an issue ID is mentioned in a commit message.

See issue #973


MichaelMure:
Works for now. Ideally we would avoid returning an error altogether if we don't know the event type, which also likely means we can remove entirely the events we don't use.

Thanks!

PR #991, opened 2023-01-13
Fixed version info be set when go install

### LINKED ISSUE #928: version information is not set with `go install`
Due to those information being set through the makefile (https://github.com/MichaelMure/git-bug/blob/master/Makefile#L3-L5), installing git-bug with `go install` result in `version` returning empty version data:
```
$ git bug version
git-bug version: -dev-
```

We should consider moving to the golang built-in feature: https://blog.carlmjohnson.net/post/2021/golang-118-minor-features/#1-version-control-information-included-in-the-binary

vasser:
Resolves #928 

In Go 1.18+ `debug.ReadBuildInfo` contains revision so we can read it from there when we are installing git-bug with `go install`.

In this PR we are utilizing the following attributes of `debug.BuildInfo`:
- `vcs.revision` - get commit hash
- `vcs.modified` - identifies whether the state of project is clean
- `vcs.time` - in discussion

This change introduces new version schema:

- `v0.8.0` - release
- `v0.8.0-dev-xyxyxy` - non release, at commit `xyxyxy`
- `v0.8.0-dev-xyxyxy-dirty` - same as above, but also dirty
- `dev-xyxyxy` - we don't know the tag, at commit `xyxyxy`
- `dev-xyxyxy-dirty` - same but also dirty

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,8 @@
 all: build
 
+GIT_COMMIT:=$(shell git rev-parse --short HEAD)
+GIT_LAST_TAG:=$(shell git describe --dirty --tags)
+GIT_EXACT_TAG:=$(shell git name-rev --name-only --tags HEAD)
 UNAME_S := $(shell uname -s)
 XARGS:=xargs -r
 ifeq ($(UNAME_S),Darwin)
@@ -7,27 +10,30 @@ ifeq ($(UNAME_S),Darwin)
 endif
 
 COMMANDS_PATH:=github.com/MichaelMure/git-bug/commands
+LDFLAGS:=-X ${COMMANDS_PATH}.GitCommit=${GIT_COMMIT} \
+	-X ${COMMANDS_PATH}.GitLastTag=${GIT_LAST_TAG} \
+	-X ${COMMANDS_PATH}.GitExactTag=${GIT_EXACT_TAG}
 
 .PHONY: build
 build:
 	go generate
-	go build .
+	go build -ldflags "$(LDFLAGS)" .
 
 # produce a build debugger friendly
 .PHONY: debug-build
 debug-build:
 	go generate
-	go build -gcflags=all="-N -l" .
+	go build -ldflags "$(LDFLAGS)" -gcflags=all="-N -l" .
 
 .PHONY: install
 install:
 	go generate
-	go install .
+	go install -ldflags "$(LDFLAGS)" .
 
 .PHONY: releases
 releases:
 	go generate
-	go run github.com/mitchellh/gox@v1.0.1 -osarch '!darwin/386' -output "dist/{{.Dir}}_{{.OS}}_{{.Arch}}"
+	go run github.com/mitchellh/gox@v1.0.1 -ldflags "$(LDFLAGS)" -osarch '!darwin/386' -output "dist/{{.Dir}}_{{.OS}}_{{.Arch}}"
 
 secure: secure-practices secure-vulnerabilities
 
@@ -56,7 +62,7 @@ pack-webui:
 # produce a build that will fetch the web UI from the filesystem instead of from the binary
 .PHONY: debug-webui
 debug-webui:
-	go build -tags=debugwebui
+	go build -ldflags "$(LDFLAGS)" -tags=debugwebui
 
 .PHONY: clean-local-bugs
 clean-local-bugs:
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,5 @@
 all: build
 
-GIT_COMMIT:=$(shell git rev-list -1 HEAD)
-GIT_LAST_TAG:=$(shell git describe --abbrev=0 --tags)
-GIT_EXACT_TAG:=$(shell git name-rev --name-only --tags HEAD)
 UNAME_S := $(shell uname -s)
 XARGS:=xargs -r
 ifeq ($(UNAME_S),Darwin)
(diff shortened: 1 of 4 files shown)


commands/root.go
```diff
@@ -90,3 +95,40 @@ func Execute() {
...
+	}
+
+	for _, kv := range info.Settings {
+		switch kv.Key {
+		case "vcs.revision":
+			c = kv.Value
+		}
+	}
+
+	cmd := exec.Command("git", "describe", "--tags", "--abbrev=0")
```
smoyer64, line 115: I'm not sure we want to run this every time a `git-bug` command is executed and, since we're already using `go-git` using `exec` might be a bit dangerous.  Theoretically, `git-bug` is run inside a `git` repository and the executable should be available but perhaps not?

The bigger issue is that this would return tag information for the `git` repository where `git-bug` is run rather than tag information for the `git-bug` repository.  I think I'd continue to ask the `Makefile` for this information.

smoyer64 (commented):
I'd looked into how to best accomplish this a couple weeks ago and agree that the new `debug.BuildInfo` type is the way to go (pun intended) since the project is not >= Go v1.18.  I'm not sure that we shouldn't still rely on the Make file for `LAST_TAG` and `EXACT_TAG` ... if there's no tag on `HEAD`, one common alternative would be to use the commit.  I also like think we should be outputting a version that includes the `dirty` flag.  Perhaps this library can be some inspiration - https://pkg.go.dev/github.com/carlmjohnson/versioninfo.

The `Makefile` also has a problem that we don't currently manage well (though it hasn't broken a release build yet) ... both `git` commands that used to be in the `Makefile` only use the last tag found on the commit.  If there's more than one tag we're going to get whatever it contains (hopefully some sort of semantic version!)

commands/root.go
MichaelMure: @smoyer64 is correct, we want to bake into the binary which code was compiled.

vasser: @smoyer64 @MichaelMure  thank you for the detailed response, I see what you say.

I think I was too fast, opening this PR.

Here are my suggestions for now, would like to validate that with you, before I move forward:

1. Use `Makefile` in the first place (revert all the changes I did)
2. If `GitLastTag`, `GitCommit` and `GitExactTag` are not defined - try to extract info from `ReadBuildInfo` (if that's possible and Go is 1.18+)
3. If by the end we do not have required info - return some sort of better defined version, as now it's `-dev-`

There is one more thing wort mentioning: `ReadBuildInfo` does return only `vcs.revision`, that is commit, and we cannot get tag from there.

Does that sound okay for you?

MichaelMure: It seems that:
- `vcs.revision` is the equivalent of `GIT_COMMIT`
- `vcs.modified` tells if the original repo was dirty, which could be used to add a `-dirty` suffix on the version name

We are indeed missing the tag, which is quite annoying.

So ... what to do ... We could use those two variables when they are available, which I assume would at least partially fix #928. Ideally golang would also expose the tag, but I don't even see an issue in their tracker for that so it's not going to happen anytime soon.

MichaelMure: But yeah, the tag extraction in the makefile need to stay for now.

So we would have different version schema:
- `v0.8.0` <-- release
- `v0.8.0-dev-xyxyxy`  <-- non release, at commit xyxyxy
- `v0.8.0-dev-xyxyxy-dirty` <-- same but also dirty
- `dev-xyxyxy` <-- we don't know the tag
- `dev-xyxyxy-dirty` <-- same but also dirty

smoyer64: There's also `vcs.time` ... after I wrote the above I was also coming to the conclusion that we might want a different format.  Fortunately, `debug.BuildInfo.Main.Version` will either contain the tag or the word "devel".  The real problem (for me) is that I'll never install a release package, and I'd love to know the "base" release for what I'm working on!

smoyer64: Related (with links to a lot more):

- https://github.com/golang/go/issues/29228#issuecomment-457390055 (and whole issue)
- https://github.com/golang/go/issues/50603

smoyer64: Note that you can still use linker flags with `go install` - would users do that if so instructed?

vasser: Good reading about the build info and it's caveats.

Basically, the case when we can retrieve a tag from `debug.BuildInfo.Main.Version` will work only when we have a release, right?

> There's also vcs.time

We can use time from the build info as well, but will this more useful info for the use?
In this case, it sounds better to use linker flags as you suggest - it will be more useful for those who really need full release. But I should admit that will have to research that, since have never used linker flags

> In the examples above, the first three would have to be tagged AND would be produced by the Makefile right?

Yes, exactly. Updating PR description with that info again

For now, updated PR: reverted back changes in Makefile. The only change there, is `--dirty` flag. And added changes to return version per new schema described above

Makefile
```diff
@@ -1,7 +1,7 @@
 all: build
 
-GIT_COMMIT:=$(shell git rev-list -1 HEAD)
-GIT_LAST_TAG:=$(shell git describe --abbrev=0 --tags)
+GIT_COMMIT:=$(shell git rev-parse --short HEAD)
```
MichaelMure, line 3: can you explain those changes?

commands/root.go
```diff
@@ -90,3 +107,28 @@ func Execute() {
 		os.Exit(1)
 	}
 }
+
+func getCommitAndDirty() (commit, dirty string) {
+	var d, c string
+
+	info, ok := debug.ReadBuildInfo()
+
+	if !ok {
```
MichaelMure, line 116: if you can't read the build infos:
- you should not continue reading the returned value, that could crash. Instead, return an error or a valid value.
- you should not print anything

commands/root.go
```diff
@@ -90,3 +107,28 @@ func Execute() {
 		os.Exit(1)
 	}
 }
+
+func getCommitAndDirty() (commit, dirty string) {
```
MichaelMure, line 111: looks like `dirty` should be a boolean

commands/root.go
```diff
@@ -90,3 +107,28 @@ func Execute() {
...
+
+	if !ok {
+		fmt.Println("could not get commit")
+	}
+
+	// get the commit and
+	// modified status (that is the flag for repository dirty or not)
+	for _, kv := range info.Settings {
+		switch kv.Key {
+		case "vcs.revision":
```
MichaelMure, line 124: instead of going over all the values, you could simply try to read: `if val, ok := info.Settings["vcs.revision"]; ok { ... }`.

vasser: `info.Settings` returns an array of structs, so I can't access required values by key name

Makefile
vasser: Oh, that change was unnecessary. I was thinking of returning a short commit hash to not to format it in the root

MichaelMure:
@vasser I made some changes, mostly style, but also fixed one or two edge cases I found. Thanks for the help :-)

vasser:
Great, happy to help 👍

PR #993, opened 2023-01-14
command: adapt the output of the bug list to the terminal size

MichaelMure:
@smoyer64 thought on this and on generalizing this pattern?


smoyer64:
I like the idea ... we should probably detect whether the output will be piped to another application and just go full width TSV.  I'm not sure how the label indicators would work in this case though.

PR #1001, opened 2023-01-18
style: clean up linter complaints

smoyer64:


commands/bug/bug_select.go
```diff
@@ -44,7 +44,7 @@ The complementary command is "git bug deselect" performing the opposite operatio
 
 func runBugSelect(env *execenv.Env, args []string) error {
 	if len(args) == 0 {
-		return errors.New("You must provide a bug id")
+		return errors.New("a bug id must be provided")
```
MichaelMure, line 47: I've been guilty of making those error messages a bit more human as they show up directly to the user. That was a long time ago though, dunno if it really makes sense.

commands/bug/bug_show.go
```diff
@@ -91,7 +91,9 @@ func runBugShow(env *execenv.Env, opts bugShowOptions, args []string) error {
 		case "title":
 			env.Out.Printf("%s\n", snap.Title)
 		default:
-			return fmt.Errorf("\nUnsupported field: %s\n", opts.fields)
+			env.Out.Println()
```
MichaelMure, line 94: trying out that command, it doesn't make that much sense to have a new line, I think we can remove it entirely.

commands/bug/bug_comment.go
```diff
@@ -26,7 +26,7 @@ func newBugCommentCommand(env *execenv.Env) *cobra.Command {
 }
 
 func runBugComment(env *execenv.Env, args []string) error {
-	b, args, err := ResolveSelected(env.Backend, args)
+	b, _, err := ResolveSelected(env.Backend, args)
```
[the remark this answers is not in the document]
smoyer64: And that pattern continues to exist where the `args` is actually used ... the functions I changed are the ones where the `args` are never used.

commands/bug/bug_select.go
smoyer64: I tried to make it somewhat human and match the convention for error messages - your choice (the capitalization was what the linter complained about and I rearranged it to get rid of the word "You")

commands/bug/bug_show.go
smoyer64: I removed one of the two new-lines ... I thought maybe that was to separate an error output from the field output that might be directly above it.

MichaelMure: Let's remove both.

commands/bug/bug_select.go
MichaelMure: Yeah, that also make sense. I suppose that to make things really proper we would need some kind of augmented error to render exactly how it should be, which would also be required  if we ever do internationalization.

commands/bug/bug_comment.go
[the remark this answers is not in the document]
smoyer64: So part of the solution should be to not shadow the function parameter that's already named `args`?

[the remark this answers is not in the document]
MichaelMure: heh, I don't see an optimal solution that also make the linter happy. But that's alright.

Issue #971, opened 2022-12-31
Submodules and working trees are not handled correctly


tionis:
In git submodules and working trees the .git path is a file linking to the real working tree via following syntax:
```
gitdir: /some/absolute/or/relative/path
```
This is not handled correctly and throws a `Error: .git exist but is not a directory` error

MichaelMure:
What would you expect to happen in that case? Read that file and open the corresponding repo?

In that case it could be an easy fix over there: https://github.com/MichaelMure/git-bug/blob/f94616ec76e01145a7b2b90c11b27a25d4a4be96/repository/gogit.go#L174

tionis:
Yes, that would be the same behavior as git.

tionis:
As far as I have seen, the paths in these files don't recurse. Instead, nested submodules/working trees point to the top most "real" git dir.
I think a simple parsing of the .git file and a change-dir should suffice

DarthRubik:
I have a WIP solution for this that I will post probably tonight after I clean
it up

DarthRubik:
Oh....Oops...I did not notice that a solution for this was already posted

tionis:
This seems to be fixed

PR #1004, opened 2023-01-21
fix(#971): parse submodule .git files instead of erroring

Kintar:
Altered logic for detecting git directory. Instead of erroring on non-direcctory .git files, now parses the file and returns the linked gitdir.

Note that since the module is a different repository, you have to `git bug user new` in the submodule before using git-bug. I've never used working trees, so I haven't tested that.

Also, I'm brand new to this project and haven't quite worked out how to add a unit test for this. Happy to do that if needed before this PR can be accepted.

### DIFF
--- a/repository/gogit.go
+++ b/repository/gogit.go
@@ -1,6 +1,7 @@
 package repository
 
 import (
+	"bufio"
 	"bytes"
 	"errors"
 	"fmt"
@@ -170,6 +171,28 @@ func detectGitPath(path string) (string, error) {
 		fi, err := os.Stat(filepath.Join(path, ".git"))
 		if err == nil {
 			if !fi.IsDir() {
+				// See if our .git item is a dotfile that holds a submodule reference
+				dotfile, err := os.Open(fi.Name())
+				if err != nil {
+					// Can't open" error
+					return "", fmt.Errorf(".git exists but is not a directory or a readable file: %w", err)
+				}
+				// We aren't going to defer the dotfile.Close, because we might keep looping, so we have to be sure to
+				// clean up before returning an error
+				reader := bufio.NewReader(dotfile)
+				line, _, err := reader.ReadLine()
+				if err != nil {
+					_ = dotfile.Close()
+					return "", fmt.Errorf(".git exists but is not a direcctory and cannot be read: %w", err)
+				}
+				dotContent := string(line)
+				if strings.HasPrefix(dotContent, "gitdir:") {
+					_ = dotfile.Close()
+					// This is a submodule parent path link. Strip the prefix, clean the string of whitespace just to
+					// be safe, and return
+					dotContent = strings.TrimSpace(dotContent[7:])
+					return dotContent, nil
+				}
 				return "", fmt.Errorf(".git exist but is not a directory")
 			}
 			return filepath.Join(path, ".git"), nil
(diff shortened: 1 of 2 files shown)


repository/gogit.go
```diff
@@ -170,6 +171,28 @@ func detectGitPath(path string) (string, error) {
...
+				// clean up before returning an error
+				reader := bufio.NewReader(dotfile)
+				line, _, err := reader.ReadLine()
+				if err != nil {
+					_ = dotfile.Close()
+					return "", fmt.Errorf(".git exists but is not a direcctory and cannot be read: %w", err)
+				}
+				dotContent := string(line)
+				if strings.HasPrefix(dotContent, "gitdir:") {
+					_ = dotfile.Close()
```
MichaelMure, line 190: This `Close()` is misplaced, you want to close as soon as possible (just after the `ReadLine()` here, after handling the error). In your code still have one code path that can lead to not closing the file.

repository/gogit.go
```diff
@@ -170,6 +171,28 @@ func detectGitPath(path string) (string, error) {
...
+					// This is a submodule parent path link. Strip the prefix, clean the string of whitespace just to
+					// be safe, and return
+					dotContent = strings.TrimSpace(dotContent[7:])
```
MichaelMure, line 193: It's not wrong, but what about using `strings.TrimPrefix(dotContent, "gitdir:")` that tells a bit more what you are doing?

MichaelMure (changes requested):
Small nitpicking.
It'd be nice indeed to have some test for that. Maybe add a new test case in https://github.com/MichaelMure/git-bug/blob/60d40d60e9f91247b61f541888f1469bff44f573/repository/gogit_test.go#L12 ? That is:
- generate a repo as normal
- create a temp folder (`t.TempDir()`), put the proper .git file there
- make sure that opens properly

MichaelMure:
> Note that since the module is a different repository, you have to `git bug user new` in the submodule before using git-bug. I've never used working trees, so I haven't tested that.

Could you expand a bit on that? Once your code setup the root at the "remote" repository, there should be no difference with a normal repo.

repository/gogit.go
Kintar: Would you believe I've been using Go since 1.8 and never noticed the strings.TrimPrefix function? XD  Absolutely!

Kintar:
@MichaelMure said:

> Could you expand a bit on that?

Sure.  Say you have a repository "repo1" with a submodule "module1".  If you're in a directory which is tracked by `repo1` then the git root directory is `repo1/.git/`, and git-bug's metadata lives in `repo1/.git/git-bug/`.  If you then switch to a directory tracked by the submodule then the git root becomes `repo1/.git/modules/module1`, and git-bug will start using metadata from `repo1/.git/modules/module1/git-bug`.

I think this is the desired behavior, since technically you're now tracking bugs on the submodule and not on the parent repository, but I called it out in case the assumption was made that bugs logged while in a submodule path should be tracked in the parent repository.  If that's the case, then the logic in gogit.detectGitPath needs to return the parent repository's git path and not the submodule's.

Kintar:
Changes submitted as requested.

Kintar:
Realized I did not update the error message returned when .git is neither a directory nor a workspace file. Updated error message.

MichaelMure:
Sorry for the delay.

I improved your PR a bit:
- add a limited reader to avoid attacks with very large .git file
- support recursive indirection (up to depth 10)
- check that the pointed to repo does exist

PR #1016, opened 2023-01-31
DOC: it is "new" not "configure" command (also was missing \)

yarikoptic:


smoyer64 (approved):

MichaelMure:
This is technically correct, but it's not only that command. Most of the commands have changed names recently and the doc will need to be adjusted before the next release. Unfortunately, the Readme in the master branch should IMHO reflect the currently released version, which means we should update it only right before the next one.

yarikoptic:
FWIW my 2c:
- I've got to catch that by using `git-bug`, although indeed just freshly built from `master` ... I would say people who follow README.md right here on github are as likely to just run that `master` version so I would have just kept code/docs consistent within the branch. People who run released version - should read some proper online documentation or README.md shipped along with the package since there is no guarantee that `master` is reflective of the release (might be outdated) they run.
- If you worry about visitors to follow some README.md too advanced for released version -- you could make `dev` branch which would not be default (as some projects do), and keep `master` aligned with releases.  Then it would be "win/win" - consistent and not visible by default until released. But IMHO it is an overkill.

MichaelMure:
This conundrum could be resolved by having the new doc and a message that point to the v0.8.0 tag.

sudoforge (approved):
Thanks for this!

---

I'm strongly of the opinion that we should never have any sort of disconnection between documentation and features, meaning that checking out _any_ revision of the repository should result in having the accurate source code and documentation for working with `git-bug`.

In the future, I think we can do better. I'm currently mulling over [simplifying the README](#1222), and will follow that with some work to restructure and better organize the documentation to reduce some of the noise.

PR #890, opened 2022-09-26
Initial Gitea bridge.

MichaelMure:
I rescued https://github.com/MichaelMure/git-bug/pull/675 of which the original repo has been deleted. Even though @tianyuanhao seems to not work on it anymore (apologize if that's not the case, you are of course welcome here), it seems that the configuration part and the importer are relatively complete.

@6543 @Vitali64 @bqv @gc-ss As you seemed interested, could you all give it a test drive and report on your findings? I don't and have never used gitea, but maybe together we can push that far enough.

PR #1017, opened 2023-02-04
Gitea bridge

DarthRubik:
Starting with https://github.com/MichaelMure/git-bug/pull/890, I added a fix for deleted accounts or private accounts that commented or created an issue.....But this is probably not my final solution (this is perhaps a bug in the API being called....I don't know enough yet...)

Other than that issue, it pulled in all the issues from https://gitea.com/xorm/xorm correctly (which I chose because it looked to be one of the most popular repos on gitea.com), except it marked them all as open.....So more digging needs to happen there....

### DIFF
--- a/bridge/bridges.go
+++ b/bridge/bridges.go
@@ -3,6 +3,7 @@ package bridge
 
 import (
 	"github.com/MichaelMure/git-bug/bridge/core"
+	"github.com/MichaelMure/git-bug/bridge/gitea"
 	"github.com/MichaelMure/git-bug/bridge/github"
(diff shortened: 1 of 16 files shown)


MichaelMure:
That's pretty cool, thank you. At first glance, you will at least need to rebase on the current master version. Some internal API have changed a bit, but not too much.

> except it marked them all as open...

It seems like your importer doesn't import state change (open, close). Implement that and you will have the proper state in git-bug.

Also, could you update https://github.com/MichaelMure/git-bug/blob/master/doc/feature_matrix.md#bridges ?

Issue #1014, opened 2023-01-30
Error: the repository you want to access is already locked


arnaudlecam:
Hi,
Lauching Git Bug (v0.8.0) gives me this Error message:
`Error: the repository you want to access is already locked by the process pid 1138`
I don't have any 1138 PID running, and Git is working fine (Guitar, Gitk, etc.)
Is there a way to unlock ?

arnaudlecam:
Sorry, finally I made a reboot : it works now, but I don't know how to reproduce... So issue could be closed.
I'll try a few tests, and then I'll send you a feedback if it succeeds.

arnaudlecam:
Here I reproduced:
First, you run git-bug: here `.git/git-bug/lock` contains the PID (ex.: 1322).
The save all your works and close all youre running tasks (but you don't close git-bug).
Now if you hard-reboot (pressing "Off" button a long time) while running git-bug, then git-bug is killed and `.git/git-bug/lock` isn't deleted (containing the killed PID: 1322).
Now you reboot, but you don't start with git-hub : you run another programms and don't close them, until those PID's are greater than the killed PID (ex.: you can launch chromium and open a large number of tabs).
With (un)luck, it will open a process with the same PID (1322).
Then you launch git-hub, and it's reproduced (ex.: `Error: the repository you want to access is already locked by the process pid 1322`).
The first time I reported it, I was realy unlucky: after hard-reboot, I think the "killed PID" was owned by a system's process so it wasn't listed with `ps -e`.
May I should test the PID's name to ? If the PID is found, then whe check this PID's name. And if the PID's name is "git-bug", then git-bug isn't lauch, else git-bug could be launched.

MichaelMure:
That's a very good point. I suppose we need a reliable and multi-platform way to check that the running process is indeed git-bug. Now, how to do that ...

arnaudlecam:
May [process.Name](https://github.com/shirou/gopsutil/blob/v3.21.11/process/process.go#L346) or [process.NameWithContext](https://github.com/shirou/gopsutil/blob/v3.21.11/process/process_linux.go#L83) should do the trick ?
But sorry, I never wrote a Go line: you should be more efficient than me ;-)

MichaelMure:
I'm busy with other things, so anyone would be faster than that :-)

smoyer64:
[...]

PR #1025, opened 2023-03-02
Func IsRunning : checking process name

arnaudlecam:
Resolves MichaelMure/git-bug#1014
It's my first Go commit, so have a look please !
I renamed process.go -> findproc.go to prevent collide with github.com/shirou/gopsutil/process
I don't really understood what was wrong or not with github.com/selesy/gokart-pre hash...


MichaelMure:
I've looked at this, and well, I'm not sure what to do.

First, that PR introduce a very large dependency. That's not a deal breaker, but for that, that seems a bit overkill.

Then comes the question of if it is the right thing to do to begin with. Checking the name of the process is kinda brittle. There is already plans to rename the binary to `gb`, and have wrapper (or filesystem hard link) named `git-bug`, `git-board`, `git-pr` ... and whatever else might the user decide. Even checking if that other process **has the same name as the current process** instead of hard-coding `git-bug` would not fully solve that problem.

I thought about exposing a dead-simple http endpoint on a fixed port, returning a fixed payload, to test the other process, but then a fixed port can be very problematic on some systems.

So ... I kinda wonder if not doing anything is actually the right move. Testing the process existence already does a lot at making git-bug seamless when something crashed. Maybe trying to do more would actually make us go backward?

MichaelMure:
While I was researching this, I found some improvement for that process testing: https://github.com/MichaelMure/git-bug/pull/1031

smoyer64:
I felt the same way about the huge number of dependencies that package added (maybe because it's really thorough and cross-platform?).  At some point, adding a message to delete the lock if you're sure there's not another copy running might be prudent.

arnaudlecam:
I agree @smoyer64. For such a rare case,

- adding dependencies is expensive;
- using the binary name looks unwieldy.
A message looks much more efficient.

smoyer64:
If the consensus (almost misspelled that) is to add a message to the output if there's a lock, we should close this issue and open one to effect that change.

sudoforge:
Due to a lack of response from the author, I'm closing this PR.

PR #1045, opened 2023-03-09
github: try to make token generation more robust against bad connection

MichaelMure:


akhavr:
Sometimes it fails even earlier:

```
Select option: 1
Github login: akhavr
Error: Get "https://api.github.com/users/akhavr": read tcp 192.168.1.193:52634->140.82.121.6:443: read: connection reset by peer
```

Will play more with the bridge creation trying to catch most failure modes.

smoyer64 (approved):
Should we adopt something like `backoff`? (https://pkg.go.dev/github.com/cenkalti/backoff/v4).  This is probably not the only place that would benefit from retries.  We should also output a warning if the users `GITHUB_TOKEN` isn't valid (has expired, etc) since you technically have a token but you're still rate-limited like an unauthorized user.

sudoforge:
> Should we adopt something like `backoff`? (https://pkg.go.dev/github.com/cenkalti/backoff/v4)

yes :)

Issue #1072, opened 2023-05-29
Separate command (or option) to pull user identities from remote


mstenek:
In a freshly cloned repo, prior to interacting with any issues/bugs, git-bug forces you to create a brand new user -- even if your git-bug user is already stored in the remote; so, you end up having a different user every time you clone the repo!

Unless this is the desired functionality (which seems unnecessarily superfluous), **I suggest allowing git-bug to pull the users stored in the remote in a separate command (or option)** so that you can adopt your an already created user and from there interact with the issues/bugs. Thus potentially saving on a surfeit of duplicate (or orphaned) users.

MichaelMure:
This is more of less a dup of #1003. The solution is not to have a new command, but to make `pull` fail later, only when we do need to have an identity (to merge data) instead of a hard requirement to do any pull. This would cover your scenario as you would pull data without having to do any merge (you don't have data locally yet, nothing to merge).

https://github.com/MichaelMure/git-bug/issues/1003#issuecomment-1411803779

Issue #1003, opened 2023-01-19
`git bug user`: unable to adopt identity


vitalyster:
I have pushed git-bug bugs and identities to some remote.
Now I want to pull that remote on another machine and I got `merge error: No identity is set. `
I'm able to create new identity only, it is not possible to adopt existing one because it is not merged.
Now I have 2 identities and it is not possible to merge them/delete one.
Am I miss something, or this is expected issue?

MichaelMure:
I think this is a valid thing we need to fix, or at least check. I think at the moment git-bug requires having an identity to pull, as it might be required to make a merge commit. That doesn't sit well with your scenario.
I think we need to make having a user optional there, and only error when we do need to make a merge.

MichaelMure:
Copy/paste my other comment:

> This is more of less a dup of #1003. The solution is not to have a new command, but to make `pull` fail later, only when we do need to have an identity (to merge data) instead of a hard requirement to do any pull. This would cover your scenario as you would pull data without having to do any merge (you don't have data locally yet, nothing to merge).

MichaelMure:
That is:
- remove that check: https://github.com/MichaelMure/git-bug/blob/master/cache/subcache.go#L543-L547
- make sure that the called functions behave properly when `author` can be nil, fail and present the error properly to the user when we need an `author` but don't have one

sudoforge:
Adopting an identity is possible by manually fetching the `refs/identities` namespace, initiating the identity and bug cache, and then adopting an existing identity. Here's an example below, with any hash and user randomized:

```
➜ git init
Initialized empty Git repository in /tmp/tmp.4m5LR4TcGG/.git/


➜ git remote add origin git@github.com:example/repo.git


➜ git fetch origin +refs/identities/\*:refs/identities/\*
remote: Enumerating objects: 3, done.
remote: Total 3 (delta 0), reused 0 (delta 0), pack-reused 3
Unpacking objects: 100% (3/3), 306 bytes | 306.00 KiB/s, done.
From ssh://github.com/example/repo
 * [new ref]         refs/identities/mbeaik8jm9onmu6xifh2fa2du53vgo4utzu9hspvapdjbdv51v087599nqmstg09 -> refs/identities/mbeaik8jm9onmu6xifh2fa2du53vgo4utzu9hspvapdjbdv51v087599nqmstg09


➜ git bug ls
Building identity cache... Done.
Building bug cache... Done.


➜ git bug user adopt mbeaik8
[...]

PR #1076, opened 2023-07-28
pull, push: try reading git-bug.remote config value before defaulting to 'origin' when no explicit REMOTE argument

wahern:


MichaelMure:
Neat, thank you!

I like the idea, but if I get that right, `git-bug.remote` is a brand new config entry that the user would have to setup, correct? In term of usability, it's not ideal.

What about instead do something like `cfg.ReadAll("remote.")` and parse the available remote, directly from the normal git config? It does get murky if there is more than one remote, but that should be a fairly rare case in practice. Maybe we can use `git-bug.remote` only in that rare case?

smoyer64:
If there's more than one remote, perhaps let the user select one?  We'll have to guide the user to picking identities eventually (especially when there are keys involved.)

wahern:
My intention behind this was to make it easier to push git-bug data to a separate, standalone repository, and to reduce the risk of accidentally pushing to one that I wanted to keep clean.

The inception of this idea comes from running into the user initialization issue--you can't pull into a repository without first creating a new user. See #1072. Which means I either have to manually delete[1] that new identity after the first pull into a newly cloned repository, or I end up with a new user for every repository I want to manage bugs in.

I wouldn't mind taking a shot at #1072, but that's much more involved and I wanted to just get to a place where I could try out git-bug for awhile before fully committing to it; I didn't want to have any remnants in my central repository if things didn't work out.

Regarding asking the user: At least for my open source projects, my working repositories usually have at least two upstream remotes: my own hosted server and GitHub. Similarly, for this particular project there's also at least two remotes counting the separate git-bug repository (standalone or not). Interactively asking the user every time would be annoying for my workflow. And if it kept state so it only needed to ask once, that's state I'd like to be able to manually control, just like with regular `git`. If you mean you would prefer a patch to include the ability to ask if `git-bug.remote` isn't yet explicitly set, rather than defaulting to `origin`, that makes total sense and I'd be happy to try that.

[1] I wasn't actually able to fully remove an identity from a repository. I can delete the identity from .git/refs/identities, but that still leaves around orphan objects in .git/objects which I couldn't figure out how to remove with git prune or git fsck. I don't think these objects would get pushed up. But once they did (e.g. by accident) you're back to square 1 in the sense that now your upstream repository will have been polluted with no simple way to remove the objects without recreating the repository.

MichaelMure:
@wahern as you will see in #1072 , #1003 , the solution for that problem is different from what you suggest here.

That said, your point about having multiple remote for different purpose makes sense. In that situation, the user is very likely to always interact with the same remote for bugs, so having a way to bake that is relevant.

There is really ~two~ three different problems here:
- user want to always interact with the same remote
- there is only one remote, but it's not `origin`
- there is more than one remote, we need to pick one

What about:
1. List the remotes (with something lke `cfg.ReadAll("remote.")`). If there is only one, pick that. If there is none, fail.
2. We have more than one remote --> if `git-bug.remote` exists, use that.
3. Print an error indicating we need the user to choose which remote explicitly, or set `git-bug.remote` --> clean and easy fail, and the user is guided towards a more sticky solution

Do you think you could implement that?

MichaelMure:
> I wasn't actually able to fully remove an identity from a repository. I can delete the identity from .git/refs/identities, but that still leaves around orphan objects in .git/objects which I couldn't figure out how to remove with git prune or git fsck. I don't think these objects would get pushed up. But once they did (e.g. by accident) you're back to square 1 in the sense that now your upstream repository will have been polluted with no simple way to remove the objects without recreating the repository.

If think you are looking for `git gc` to remove those unreachable objets. In any case, they should never be pushed (as not enumerable).

sudoforge:
@MichaelMure [said](https://github.com/MichaelMure/git-bug/pull/1076#issuecomment-1673733075):

> 1. List the remotes (with something lke `cfg.ReadAll("remote.")`). If there is only one, pick that. If there is none, fail.
> 2. We have more than one remote --> if `git-bug.remote` exists, use that.
> 3. Print an error indicating we need the user to choose which remote explicitly, or set `git-bug.remote` --> clean and easy fail, and the user is guided towards a more sticky solution

I think the `git-bug.remote` should be preferred, no? So the order of operations should be:

1. If `git-bug.remote` exists, use that.
2. If not, list all remotes. If there is more than one remote detected, prompt the user to select a remote and set that as the value for `git-bug.remote`.
3. If there are no remotes, 

This provides a pretty seamless workflow. An informational message could be logged if `git-bug.remote` is updated, I guess, to provide that information to the user. I don't think erroring out and making them manually set

sudoforge:
@wahern this is a feature i'd be interested in merging. do you have the cycles to address the above suggestion?

sudoforge:

```
$ git fetch https://github.com/git-bug/git-bug.git refs/pull/1076/head
$ git checkout FETCH_HEAD
```

Issue #349, opened 2020-03-05
Implement Gitea bridge


strk:
It would be great to have one, be it a first step toward [integration](https://github.com/MichaelMure/git-bug/issues/80) or not

MichaelMure:
Any volunteer around ?

tredoe:
There is also other self-hosted Git service implemented in Go: [Gogs](https://github.com/gogs/gogs)

AtomToast:
afaik gitea is a fork from gogs with a different development philosophy and from what I saw also more features

6543:
I might have a look at this ...

GalaxySnail:
Not stale.

WillForan:
for reference, I think PR #1077 is the most recent (July 2023) effort to implement the bridge and issue #80 addresses gitea integration on the whole

sudoforge:
I'm hestitant to add new bridges at all given that it adds to the maintenance burden of `git-bug` and its core features (imagine a world in which there is a bridge for _every other issue management platform_ -- who is going to maintain it?). I'm much more in favor of maintaining a common bridge interface and supporting a "plugin system", allowing the community to maintain specific bridges outside of `git-bug`.

sudoforge:
Closing in favor of keeping #80 as the source of truth for the Gitea integration request.

Issue #80, opened 2018-12-10
gitea integration


vtolstov:
Great to see such project, what do you think - how integrate such thinkg into gitea to store issues inside git?

vtolstov:
Firstly i think about as backend, to minimize usage of sql database and have ability to create/update/delete issues via console interface =)

MichaelMure:
related issue: https://github.com/go-gitea/gitea/issues/6454

hoijui:
I would also preffer it that way that gitea is the interface. why do you preffer the bridge way, @MichaelMure ?
... i would like ht einterface way very much! I think it woudl be great for both of the projects. distributedness is THE thing for concious, freedom loving people, and that tends to be an important thing in the open source community, and especially among hte people using gitea or this project, I woudl imagine. It is hte perfect marriage.

MichaelMure:
@hoijui I would very much welcome gitea to use git-bug as a backend for bugs (and even code review eventually!). 

I would consider first the bridge solution, simply because it's by far the easiest and doesn't require heavy changes in gitea. Also note that if the bridge solution is done well enough, you get most of the benefits you would have if gitea used natively git-bug as its backend.

hoijui:
ahhh nice!
perfect explanation, thank you @MichaelMure !

bqv:
@tianyuanhao desperate to see this happen, thanks for your work in #675

nsrosenqvist:
Would love this too, also note that Forgejo is another popular for of gitea/gogs, and I think a bridge would be able to provide support for more than just gitea without much issue.

GalaxySnail:
Not stale.

mattcen:
As per https://github.com/MichaelMure/git-bug/issues/349#issuecomment-2267075506:
@sudoforge said:
> I'm hestitant to add new bridges at all given that it adds to the maintenance burden. […] I'm much more in favor of maintaining a common bridge interface and supporting a "plugin system", allowing the community to maintain specific bridges outside of `git-bug`.

Do you have ideas or the beginnings of a plan for this? Is it the sort of thing that's likely to get any more traction than the current approach? My concern is that while I agree that this will improve separation and reduce maintenance burden on git-bug core, it would likely involve a significant interface/API design and reimplementation of the 4-5 existing bridges, which is a lot of initial up-front work that could take significant effort (especially on volunteer time).

sudoforge:
[...]

PR #1077, opened 2023-07-29
WIP: Gitea/Forgejo bridge

mattcen:
This PR deprecates #1017 (and #890 and #645 before it). I've rebased off current `master`, updated the feature matrix to the best of my ability, and built the code to confirm I can pull bugs from a Gitea repo.

I've not done anything else so far. I'm *very* new to Go, so am starting small, but wanted to at least give this a shot because having a Gitea bridge would be valuable to me.


bridge/gitea/import.go
```diff
@@ -0,0 +1,195 @@
...
+
+			// create issue
+			b, err := gi.ensureIssue(repo, issue)
+			if err != nil {
+				err := fmt.Errorf("issue creation: %v", err)
+				out <- core.NewImportError(err, "")
+				return
+			}
+
+			// Loop over all comments
```
MichaelMure, line 80: looks like this is a quite important missing feature, no?

bridge/gitea/config.go
```diff
@@ -0,0 +1,305 @@
...
+		cred = token
+	default:
+		if params.Login == "" {
+			if !interactive {
+				return nil, fmt.Errorf("Non-interactive-mode is active. Please specify the login name via the --login option.")
+			}
+			// TODO: validate username
+			login, err = input.Prompt("Gitea login", "login", input.Required)
+		} else {
+			// TODO: validate username
```
MichaelMure, line 86: todo: at least make sure it's not empty, return an error if so

bridge/gitea/export.go
```diff
@@ -0,0 +1,29 @@
...
+	"github.com/MichaelMure/git-bug/bridge/core"
+	"github.com/MichaelMure/git-bug/cache"
+)
+
+var (
+	ErrMissingIdentityToken = errors.New("missing identity token")
+)
+
+// giteaExporter implement the Exporter interface
+type giteaExporter struct {
```
MichaelMure, line 19: this whole exporter could be removed (as it's not used), instead of returning those weird errors

bridge/gitea/gitea.go
```diff
@@ -0,0 +1,57 @@
...
+	"time"
+
+	"code.gitea.io/sdk/gitea"
+
+	"github.com/MichaelMure/git-bug/bridge/core"
+	"github.com/MichaelMure/git-bug/bridge/core/auth"
+)
+
+const (
+	target = "gitea-preview"
```
MichaelMure, line 13:
```suggestion
	target = "gitea"
```

MichaelMure (commented):
Well, as a maintainer, I see the value of having such a bridge, but it's also a bit scary to merge (and have to maintain) a bridge with zero tests, and incomplete like this.

There is a few tiny fixes to do, but the biggest issue is not having the capability to import comments. It seems like it greatly reduce the usefulness of such importer.

Do you think you could implement that part? It seems like there is already the corresponding iterator, it should be "just" a matter of replicating that data into git-bug.

mattcen:
Oh for sure; there's significant work to be done before it's remotely merge-able, hence the "WIP" and "Draft" status. I'm not sure when I'll be in a position to do more than I have already, but wanted to at least share the progress of bringing the previous PRs up to a state where they at least *compile* and run on the current code-base.

Thanks for the direction in your comments, though; that will certainly make it easier to make a start on this. :)

will-ca:
Any thoughts on [Forgejo vs Gitea](https://forgejo.org/faq/)? Looks like they should be basically compatible, and have a good amount of momentum behind them.

mattcen:
Software-wise they're currently identical since Forgejo is a soft fork.

WillForan:
To complete the git issue/PR graph: this address issues #349 (bridge), part of #80 (integration). And has related issue on the gitea side at https://github.com/go-gitea/gitea/issues/6519

bridge/gitea/config.go
mattcen: Fixed

mattcen:
I've had a little time this weekend to poke at this, and have started adding some unit tests (based on those for the Github/Gitlab bridges) to make this easier to develop. Given I'm still learning the language, I figure having a test harness is a particularly good way to proceed. Not sure how far I'll get with the time I have, but… small steps!

bqv:
̇

PR #1145, opened 2024-03-26
Use HTTPS repo url

jvasile:
When cloning this repo for installation (as opposed to development), users should use the HTTPS url for this repo, as they almost certainly do not have the necessary access rights to clone via SSH.


MichaelMure:
Thanks!

Issue #1023, opened 2023-02-23
Error: error creating SSH agent: "SSH agent requested but SSH_AUTH_SOCK not-specified" on `git bug push`


NobbZ:
Please guide me to a way to provide additional information…

Yes, I use SSH URLs for git.
No, they are not password protected.

If though starting SSH-Agent *and* adding the key, the error becomes:

> Error: ssh: handshake failed: knownhosts: key mismatch

Regular git operations against the remote work.

patricioantoniotoledo:
cd ~/.ssh
ssh-add ~/.ssh/id_rsa

might help

NobbZ:
To be honest, I do not even (want to) use SSH-agent.

I want the configured SSH-config to be respected and the command to just work.

I stoped using `git-bug` because of this a couple of days after posting this issue without a reply.

octvs:
Might be [related](https://github.com/MichaelMure/git-bug/issues/1109#issuecomment-1732362208), I had some similar issues with `pull` & `push`, building from source fixed.

stackengineer:
Not stale. Also looking for an agentless solution, preferably respecting `~/.ssh/config`.

sudoforge:
> git's shorthand (scp like syntax) won't work for me (until ~/.ssh/config gets parsed) as I need to set a custom port. The scp like syntax does not allow setting a custom port number.

yeah, this is something i know about -- parsing custom ssh config is on my radar.

in the meantime, if your only need is to specify a custom port, using the full RFC-3986-compliant URI would work for you:

```
ssh://you@host:1234/foo/bar.git
```

> A SSH Agent is still needed. If I try a git bug pull or push without starting an agent, I get:
>
> ```
> Error: error creating SSH agent: "SSH agent requested but SSH_AUTH_SOCK not-specified"
> ```

yep - this is on my radar as well.

generally speaking, the way git-bug is currently written makes a lot of assumptions about your git configuration, your ssh configuration, and the types of remotes you're interacting with. these are the core issues i want to solve in the near term, as they are silly, and outright blocking people.

PR #1146, opened 2024-04-13
ssh authenticate

### LINKED ISSUE #934: 0.8.0 issue with Authentication

wiegelmann:
fixes #934 and #1023

### DIFF
--- a/repository/gogit.go
+++ b/repository/gogit.go
@@ -9,6 +9,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"regexp"
 	"sort"
 	"strings"
 	"sync"
@@ -21,6 +22,7 @@ import (
 	"github.com/go-git/go-git/v5/plumbing"
 	"github.com/go-git/go-git/v5/plumbing/filemode"
 	"github.com/go-git/go-git/v5/plumbing/object"
+	"github.com/go-git/go-git/v5/plumbing/transport/ssh"
 	"golang.org/x/sync/errgroup"
 	"golang.org/x/sys/execabs"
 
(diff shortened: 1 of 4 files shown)


MichaelMure (commented):
Hi, thank you for doing this.

A few things:
- as we have seen with the recent XY backdoor, it's important for that kind of code to be cristal clear: could you comment in the body of those function about what you are doing and why?
- is that a workaround for a go-git issue? If so, could you mark it clearly, ideally with a link to an upstream issue?

wiegelmann:
Hi, according to https://github.com/go-git/go-git/issues/218 go-git does not load public keys from ~/.ssh, meaning you need an active ssh-agent running otherwise you will get errors like:

- error creating SSH agent: "SSH agent requested but SSH_AUTH_SOCK not-specified"
- ssh: handshake failed: ssh: unable to authenticate, attempted methods [none publickey], no supported methods remain

If the ssh-agent is running and contains valid keys, everything is fine. If the ssh-agent has invalid keys or is not running, we try to find public keys in ~/.ssh and retry fetch/push again.

Tested on Linux and Mac with the following scenarios:

- ssh-agent running with valid and invalid keys in ~/.ssh
- ssh-agent not running with valid keys in ~/.ssh

sudoforge:
I'm not sure that I agree with this approach.

Given that there are various methods for running an ssh-compatible agent (I use gpg-agent to emulate ssh-agent, for example), and many different ways that people may be managing keys, this really only works for users that follow the default convention of having some key(s) with a well-known name in `~/.ssh`, which isn't always going to be the case.

This is also problematic because it ignores host-specific configuration a user may have set up in their `~/.ssh/config`, and doesn't handle cases where the file contains a paswword-encrypted PEM block.

Instead, I'd prefer checking for an SSH-like remote URI (ensuring that we've resolved `insteadOf` references, but that may be out-of-scope for this change), and if we fail to fetch, emitting a useful error message pointing the user in the right direction (telling them plainly that we failed and that the reason may be due to a lack of configuring ssh-agent).

sudoforge:
i'm going to close this for now. if you wish to address the feedback and submit an update, you can always check out this tree like so:

```
$ git fetch https://github.com/git-bug/git-bug.git refs/pull/1146/head

$ git checkout FETCH_HEAD
```

PR #1148, opened 2024-04-17
Update README.md

mcepl:
I know this is a stupid nit, but people have a tendency to copy & paste to the command line, so it should work.


MichaelMure:
Not stupid, thank you :-)

PR #1189, opened 2024-07-19
A lil' bit of TLC for the pipelines

sudoforge:
```
- feat: update action library versions
- feat: add concurrency limits to all pipelines
- fix: bump to go v1.22.5
- fix: correct typo: acceps => accepts
```

---

This is a **stack of logically independent commits** and should be **rebased** instead of merged with a merge commit, or squashed.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -20,13 +20,16 @@ jobs:
     name: Performance regression check
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/setup-go@v3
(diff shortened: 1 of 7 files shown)


sudoforge:
@MichaelMure i'm getting test failures due to [expired bridge tokens](https://github.com/MichaelMure/git-bug/actions/runs/10002724790/job/27648485703#step:5:141). I can update the pipeline to skip this, however, I'd appreciate it if you rotated the token so that I can ensure the test suite passes (especially because I updated the pipeline to use `go 1.21`).

Issue #1194, opened 2024-07-21
flaky test on windows: //cache/repo_cache_test.go%TestCache


sudoforge:
```
=== RUN   TestCache
    repo_cache_test.go:163: 
        	Error Trace:	D:/a/git-bug/git-bug/cache/repo_cache_test.go:163
        	Error:      	Received unexpected error:
        	            	remove C:\Users\RUNNER~1\AppData\Local\Temp\TestCache1576603400\001\.git\refs\bugs\6cc51bbddc3c2a6de7c8effacd3766e2ea7935a6a3fee9a4f46af3ea846921d0: The process cannot access the file because it is being used by another process.
        	Test:       	TestCache
```

_Sourced from: https://github.com/MichaelMure/git-bug/actions/runs/10024586195/job/27706826825#step:5:196_

sudoforge:
i've seen this crop up on non-windows systems, so i'm removing that qualifier from the title.

PR #1229, opened 2024-08-04
fix: correct path for reusable workflow: lifecycle

sudoforge:
```
fix: correct path for reusable workflow: lifecycle

Change-Id: I02c6e1f183e4b6ee67c0a4d0dcb7c5fbeff361ac
```


sudoforge:
merging, because this is just a workflow change, and the pipeline is only failing due to a flaky test: #1194

PR #1252, opened 2024-08-25
core: make label a common type, in a similar fashion as for status

MichaelMure:
This will be useful for Board, and likely code review support later


sudoforge (dismissed):
looks good to me, but perhaps these comments should be updated (i'll defer to your judgment here):

```
➜ rg 'bug\.label\b'
api/graphql/connections/connections.go
7://go:generate genny -in=connection_template.go -out=gen_label.go gen "Name=Label NodeType=bug.Label EdgeType=models.LabelEdge ConnectionType=models.LabelConnection"

api/graphql/connections/gen_label.go
14:// BugLabelEdgeMaker define a function that take a bug.Label and an offset and
```

sudoforge (approved):

PR #1329, opened 2025-02-25
docs: use `configure` instead of `new` for  bridges in `README`

zyv:
It seems that the command to set up a bridge has changed and the readme hasn't been updated. This is very confusing, so I hope you can accept this small PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -227,17 +227,17 @@ The web UI interact with the backend through a GraphQL API. The schema is availa
 Interactively configure a new github bridge:
 
 ```bash
-git bug bridge new
+git bug bridge configure
 ```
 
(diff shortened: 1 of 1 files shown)


zyv:
To follow up on this:

It seems that the documentation in the README was updated by @yarikoptic in f00e42e and is for the current master that was never released. So when I installed the Homebrew version, the instructions didn't work for me.

I hope you revive this project sometime. I found it a cool way to work with issues, and was seriously considering it as an alternative to backup issues for GitHub-hosted repositories, but in its current state, unfortunately, it can't be relied upon. I will try to use GitHub's migration API for backups.

Issue #1367, opened 2025-04-19
TestGitlabPushPull is broken


sudoforge:
https://github.com/git-bug/git-bug/actions/runs/14547588831/job/40815002321#step:5:142

this tree was branched off of HEAD at 52f42daf8bf317a065277da50d916c90e26fe4dd

sudoforge:
i started on this, but will be AFK for a few days so i won't be able to complete it. if you're interested in the status: there's a detailed comment in the commit; take a look at the PR.

sudoforge:
opened an upstream issue: https://gitlab.com/gitlab-org/gitlab/-/issues/536827

i have no idea if that's the right place. they have a ton of projects and no clear direction on where to file "API issues".

sudoforge:
i'll summarize the root cause here for people who'd rather not read the thread on gitlab (_there's a bit of back and forth about how to potentially improve this scenario, if you're interested_).

---

when importing from gitlab, in order to get the historical events related to a gitlab issue, we hit the "issue notes api", and get a list of objects that look like this:

```
{
  "id": 123412341234,
  "type": null,
  "body": "changed title from **bug title edited** to **bug title edited{+ again+}**",
  "author": {
    ...
  },
  "created_at": "2024-03-19T20:56:59.456Z",
  "updated_at": "2024-03-19T20:56:59.463Z",
  "system": true,
  "noteable_id": 12341234,
  "noteable_type": "Issue",
  "project_id": 12341234,
  "resolvable": false,
  "confidential": false,
  "internal": false,
  "imported": false,
  "imported_from": "none",
  "noteable_iid": 1,
  "commands_changes": {}
}
```

the critical field here is the `body` field. this value represents that the title was changed from `bug title edited` to `bug title edited again`.

gitlab uses this sort of message to render the "comments" that are added to issues when different event types occur: a title being changed, a mention from somewhere else, commits being added, etc. this comprises the "notes" that get added to an issue's feed (you can see a similar thing on github issues, above, for example). for this use case, gitlab doesn't care about the `body` field's value -- they just dump it into `SomeRenderer` that spits out something for their UI.

conversely, `git-bug` cares about the _type_ of event. we track these events as strongly typed "operations", and therefore, we need to _parse_ this event.

historically, this field's value looked like what you see above. with the advent of the 17.11 release, however, gitlab changed the format of the value.

this means that the following two values represent the exact same title change, from `bug title edited` to `bug title edited again`:

_original_

```
[...]

PR #1370, opened 2025-04-19
fix: refactor how gitlab title changes are detected

sudoforge:
This change refactors how issue title changes from gitlab events are
detected, fixing an issue (due to upstream changing the format of the
event body from markdown-esque to html), and improving on error
handling.

The error boiled down to a change in the issue title format. Gitlab
changed this on April 17 2025 with the release of version 17.11 [0],
although the only place a reference to this change exists is in the
changelog [1], which is not linked to from the releases page.

To account for the potential future in which other fields need to be
parsed in this way, an internal parser library was introduced at
`//bridge/gitlab/parser:parser.go` with initial support for parsing
title change messages.

An issue was opened with the Gitlab team discussing the fact that this
was a breaking change [2]. This may lead to moving title changes (or
maybe all changes) to `resource_*_events`, which would likely provide a
smoother experience for our use case.

Debugging this issue surfaced a few pain points with this bridge:

- Errors are few and far between, and when they do exist and are
  managed, they are often not propagated, often existing as simple
  `fmt.Printf` calls
- Inconsistent and uninformative logging structure when there _are_
  errors, leading to challenges in debugging unexpected behavior
- Fragility: we are parsing random text from event fields (for title
  changes and more). This will likely lead to future breakage should
  Gitlab change the format of other fields. Ideally, the gitlab SDK
  would start classifying notes and have fields like `type`, `old`,
  `new`... but this is unlikely to happen in the short term

[0]: https://about.gitlab.com/releases/2025/04/17/gitlab-17-11-released/
[1]: https://gitlab.com/gitlab-org/gitlab/-/commit/b3e1fdcf45f8b18110a2f5217b9964a11616d316#ab09011fa121d0a2bb9fa4ca76094f2482b902b7_5_232
[2]: https://gitlab.com/gitlab-org/gitlab/-/issues/536827

Closes: #1367
Change-Id: I3bd7fa1c39a9e4dd2176d6e482e30ab68965f6e7

### DIFF
--- a/bridge/gitlab/event.go
+++ b/bridge/gitlab/event.go
@@ -7,6 +7,7 @@ import (
 
 	"gitlab.com/gitlab-org/api/client-go"
 
+	"github.com/git-bug/git-bug/bridge/gitlab/parser"
 	"github.com/git-bug/git-bug/util/text"
 )
 
@@ -50,7 +51,11 @@ type NoteEvent struct{ gitlab.Note }
 func (n NoteEvent) ID() string           { return fmt.Sprintf("%d", n.Note.ID) }
 func (n NoteEvent) UserID() int          { return n.Author.ID }
 func (n NoteEvent) CreatedAt() time.Time { return *n.Note.CreatedAt }
+
 func (n NoteEvent) Kind() EventKind {
+	if _, err := parser.NewWithInput(parser.TitleParser, n.Body).Parse(); err == nil {
+		return EventTitleChanged
+	}
 
 	switch {
 	case !n.System:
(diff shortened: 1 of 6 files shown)


sudoforge:
i introduced these changes to handle the new title format we're getting from gitlab... unfortunately, as can be seen in the failing import test, it appears that the old markdown-esque format is still returned by their API.

i'm out of time to work on this for this weekend. i'll be able to pick it back up towards the end of the coming week. if someone wanted to add in an `if/else` here to handle conditionally using the html or markdown parser and send the patch here (as a comment maybe), i could take a look earlier and merge it. otherwise, we'll just have a broken pipeline for the time being.

PR #1373, opened 2025-04-21
refactor: rename //:git-bug.go to //:main.go

sudoforge:
`main.go` is and always has been the conventional entrypoint for
binaries (this is actually true for libraries as well, e.g. some internal
`//foo` package would typically have an entrypoint of `//foo:foo.go`)

Change-Id: Ic75db640b06dc5c39a8c88db0d9d73d78f39d772


sudoforge:
CI is failing due to #1367; merging this as it's a no-op.

PR #1381, opened 2025-04-28
feat: add action: auto-label

sudoforge:
This change adds a custom github action, `auto-label`, and refactors the
lifecyle workflow to use it for improving the automation around
management of the lifecycle label.

The stale bot's message is updated to make it clear that its goal is
only to ping maintainers, with additional information added detailing
each action that can be performed to remove the lifecycle status from an
issue or pull request.

Historically, users have expressed disastisfaction with the stale bot,
usually due to the presumption that it closes issues. This is not the
case, and this change hopes to improve both the automation around
managing the labels, and the messaging presented when items are marked
with an idle or dormant life due to inactivity.

Change-Id: I09a654c5ddbfdc8a83b319685e5703304eff4775


sudoforge:
internal tooling caused a duplicate. closing.

Issue #944, opened 2022-12-07
"git bug push" fails with "Error: repository not found" when using insteadOf


larsks:
I have `insteadOf` aliases in my git configuration, like this:

```
$ git config --list | grep me:
url.ssh://git@github.com/larsks/.insteadof=me:
```

If I configure a repository with a remote using that alias, as in:

```
git remote set-url origin me:git-bug-example
```

Then when I attempt to `git bug push` with this configuration, it fails with an error:

```
$ git bug push
Error: repository not found
```

It looks like `git bug` is performing git operations with something that doesn't parse git configuration files. Aliases like this can be expanded using the `git remote get-url` command:

```
$ git remote get-url origin
ssh://git@github.com/larsks/git-bug-example
```

MichaelMure:
Those are two very different code path, as git-bug is reading the local+global config on its own. On the other hand when push/pull is done, it's pure go-git. It does look like a go-git bug.

miwig:
I got this problem with a repository URL containing ~ to refer to the home directory. Using an absolute path fixed it.

sudoforge:
@larsks [said earlier](https://github.com/git-bug/git-bug/issues/944#issuecomment-1344501529):

> it looks like go-git isn't reading the global configuration: if I move the url configuration into .git/config, it seems to work as expected

i know it's been a while, but thanks for running the debugger. having used `git-bug` in some other projects, i can confirm indeed that they do not parse the `url.<base>.insteadOf` configuration (or even, last i checked, configuration via `includeIf` or other mechanisms).

support for this was (supposedly) added in [go-git/go-git@v3.0](https://github.com/go-git/go-git/commit/51cbc24bbecfecbbcea9cd733ad44eaf74b8ae4b), but it doesn't seem to actually work, with the `url.<base>.insteadOf` configuration in either the global or local scope`).

this is still an issue. it looks like we'll need to manually account for this using the default configuration chain (local > global > system, following any includes). i've accepted this bug and have slotted it in for the [0.8.2 release](https://github.com/git-bug/git-bug/milestone/10).

---

@miwig [said earlier](https://github.com/git-bug/git-bug/issues/944#issuecomment-2272295531):

> I got this problem with a repository URL containing ~ to refer to the home directory. Using an absolute path fixed it.

are you able to share your `url.<base>.insteadOf` config(s) for the remote you specified, and the CLI you ran? please be sure to redact anything you wish, such as paths on your local filesystem.

PR #1394, opened 2025-04-29
fix: resolve the remote URI using url.*.insteadOf

sudoforge:
This change introduces a new method for resolving the URI for a given
remote (by name), ensuring that the URI is rewritten following the rules
defined in `url.*.insteadOf` configuration.

When `go-git@69aefb67315d775bf73df836b1e3ba5a3b321198` lands in a
release (and we update go-git to that release), this should be
refactored to use `config.URL.InsteadOfs()`, which supports multiple
`url.*.insteadOf` configuration blocks instead of the current limitation
of always taking the last definition.

Closes: #944
Change-Id: Iffd111a1f4c42eff12613dcbd1b0d27df415f5e8

### DIFF
--- a/repository/gogit.go
+++ b/repository/gogit.go
@@ -382,8 +382,14 @@ func (repo *GoGitRepo) FetchRefs(remote string, prefixes ...string) (string, err
 
 	buf := bytes.NewBuffer(nil)
 
-	err := repo.r.Fetch(&gogit.FetchOptions{
+	remoteUrl, err := repo.resolveRemote(remote, true)
+	if err != nil {
+		return "", err
+	}
+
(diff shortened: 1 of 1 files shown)


MichaelMure:
FYI, this kind of thing (calling the git binary) has been responsible for a CVE. I'd avoid it as much as possible.

sudoforge:
> FYI, this kind of thing (calling the git binary) has been responsible for a CVE. I'd avoid it as much as possible.

yeah, i'm in the process of trying to make this work via go-git. i really don't like how half-baked it feels given that it's the "de facto" libgit2 implementation for golang.

PR #1408, opened 2025-05-10
Spelling corrections

mcepl:
I got [this patch to my repo at Sourcehut](https://lists.sr.ht/~mcepl/devel/<20250510074052.2005600-1-keeperotphones@gmail.com>).

### DIFF
--- a/.gitmessage
+++ b/.gitmessage
@@ -17,11 +17,11 @@
 #   about what the old behavior was, why the change was necessary, and what the
 #   new behavior is
 #
-# - The `footer` MAY be prsent, and should include any relevant trailers and
+# - The `footer` MAY be present, and should include any relevant trailers and
 #   other metadata
 #
 # Valid values for <type>:
-#   - deprecation: a change related to deprecation of a reosurce or interface
+#   - deprecation: a change related to deprecation of a resource or interface
 #   - design: a change related to design that does not change an interface or
 #     any application/business logic
 #   - doc: a change related to documentation


.gitmessage
```diff
@@ -1,4 +1,3 @@
-
```
sudoforge, line 1: Please revert this.

Without this newline, the editor is opened with the section break (`########....`) on the first line.

sudoforge (changes requested):
thanks! small required change.

.gitmessage
kotp: Chesterton's Fence.  Generally speaking I tend to have the break at the end of every file, rather than before, to prevent this very thing, but should have recognized it as there was no blank line at the end of this file.

Thanks for attending to it!

sudoforge: i'm not sure i'm following you.

> I tend to have the break at the end of every file

are you talking about the `###...` section break markers? or the newline at the beginning of the file? the newline is at the beginning because this file is used for [`commit.template`](https://git-scm.com/docs/git-config#Documentation/git-config.txt-codecommittemplatecode), which is configured automatically once you [activate the development shell](https://github.com/git-bug/git-bug/blob/89ff47a4ee5ccfc3f0c3ac2e5783b3cf15591ee6/CONTRIBUTING.md).

its contents are what pops up in your editor when git prompts you for a commit message. your cursor starts at the beginning of the file (byte 0).

this file uses `#` to surround the commit message guidelines to provide a "better" visual break between your commit message and the guidelines.

if you were to create a new git repository and a new commit, you'd see another sort of template:

```
$ cd $(mktemp -d)
$ git init
$ git commit --allow-empty
```

at this point, your editor would open with the following contents:

```

# Please enter the commit message for your changes. Lines starting
# with '#' will be ignored, and an empty message aborts the commit.
#
# On branch trunk
#
# Initial commit
#
```

... which is the default commit template (although there's a bit of dynamic content in there; if you had changes staged, for example, it would show up there).

kotp: The reason for the blank first line is different than I understood.  Hence "Chesterton's Fence" _Don't take the fence down until you understand it's purpose and history._  I should have at worst created a separate patch, and at best left it alone while perhaps asking about it.  I considered the impact of the "small change" in my understanding, but did not consider the history and purpose outside of my incomplete understanding.

sudoforge: i get that -- my comment was simply attempting to provide the reason it exists there, and potentially clean up some confusion :)

kotp: Yep, I appreciate it, my response was only to the

> > i'm not sure i'm following you.

> I tend to have the break at the end of every file

Since it was a "drive by" contribution done via e-mail patch, all the platform dependent things never happened (for github) and was never seen *by me*.

sudoforge: thanks for the clarification -- this actually isn't a platform-dependent thing, or related to github at all. it's a new part of the environment that's set up for contributors, through the development shell.

> a "drive-by" contribution done via e-mail patch

unrelated: you are welcome to send any future changes (however small!) to this repository directly on github (which is the source of truth for git-bug at this point in time). if you prefer avoiding github's interface (like me), you can use the `gh` cli to manage the lifecycle of your PR.

kotp:
On Sun, May 11, 2025 at 10:56:09AM -0700, sudoforge wrote:
> @sudoforge commented on this pull request.
> thanks for the clarification -- this actually isn't a
> platform-dependent thing, or related to github at all. it's a new part
> of the environment that's set up for contributors, through the
> development shell.

Apparently I failed to read the `CONTRIBUTING.md` file which makes the
repository source clear.  Sorry about that.  It also has information
about how to prepare the development environment using `nix`.

> > a "drive-by" contribution done via e-mail patch
> 
> unrelated: you are welcome to send any future changes (however small!)
> to this repository directly on github (which is the source of truth
> for git-bug at this point in time). if you prefer avoiding github's
> interface (like me), you can use the `gh` cli to manage the lifecycle
> of your PR.

PR #1427, opened 2025-05-16
feat(web): simplify header navigation

lygaret:
* removes the tabbed navigation completely
* makes it possible to see the custom name of the repository

tabs (and really all not "main" nav) were not in use and the features may not work the same way in the future, so rather than redesigning, the whole tab bare was removed.

for the custom name, if it's `__default`, the the default `git-bug` text is displayed; otherwise, the name of the repo.

### DIFF
--- a/webui/src/components/Header/Header.tsx
+++ b/webui/src/components/Header/Header.tsx
@@ -1,13 +1,11 @@
 import AppBar from '@mui/material/AppBar';
-import Tab, { TabProps } from '@mui/material/Tab';
-import Tabs from '@mui/material/Tabs';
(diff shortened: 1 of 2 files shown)


lygaret:
the margins look weird in these screenshots because I've got it in mobile width due to devtools; a quick quality-of-life pass re: responsive spacing and such probably needs to happen, but I don't think in this PR.

Screenshot 2025-05-15 at 7 45 29 PM

sudoforge (approved):
LGTM - thanks for your continued work on the web ui. it's very much appreciated!

PR #1442, opened 2025-05-17
build(deps): upgrade bleve v1.0.14 => v2.5.1

sudoforge:
This change upgrades `bleve`, primarily to remove the need for the
replace directive on `github.com/willf/bitset`, but this change upgrades
further to move past a non-impacting vulnerability to the `bleve/http`
package [0] [1], even though we do not use it, even indirectly.

Note that even though the GitHub advisory notes that this is patched in
`v2.5.0`, all that seems to have happened (see [1]) is that a README was
added to the package noting the vulnerabilities.

Other CVE databases [2] [3] do not mark this as resolved. Again,
however, git-bug DOES NOT USE this package, so it is not really a
concern.

[0]: https://github.com/advisories/GHSA-9w9f-6mg8-jp7w
[1]: https://github.com/blevesearch/bleve/commit/1c7509d6a17d36f265c90b4e8f4e3a3182fe79ff
[2]: https://nvd.nist.gov/vuln/detail/CVE-2022-31022
[3]: https://pkg.go.dev/vuln/GO-2022-0470

Change-Id: I9ab436ade9221bfd91b84ebaf47434f1b3d91cd3


MichaelMure (dismissed):

sudoforge:
holding off on merging this due to the breaking change.

i'll work on introducing version-based namespacing for the cache so that this sort of change can be abstracted away from the user and handled automatically. once that's in, i will revisit this.

sudoforge:
last three force pushes:

- rebasing the two commits in this tree
- updating the commit message
- rebasing on top of `HEAD` to resolve conflicts in `go.mod` and `go.sum`.

this is ready to merge now.

go.mod
```diff
@@ -42,29 +42,40 @@ require (
 	golang.org/x/text v0.24.0
 )
 
+require (
```
MichaelMure, line 45: it's nice if you can keep the indirect deps in the same block

sudoforge: huh, that's weird. i just regenerated this with `go mod tidy` when rebasing to resolve the conflicts with `HEAD`.

why would it have created this second block for these indirect deps?

sudoforge: it seems like the `tool` directive prevents `go mod tidy` from "using" the `require` block below it. that's unexpected

sudoforge: manually reordered the directives in f281039

PR #1449, opened 2025-05-18
build(web): upgrade graphql-eslint plugin

lygaret:
with this upgraded, we no longer depend on an vulnerable version of the `ws` library through an old graphql version


sudoforge (approved):
LGTM, and that was some quick work noticing the format issue :P

PR #1456, opened 2025-05-19
graphql: implement changes subscription

MichaelMure:
Most is there and compile, untested.


sudoforge (commented):
overall, this looks good! i think we want to take a look at improving the handler though.

PR #1460, opened 2025-05-19
feat(config): add git-bug.remote for defining the default remote

mcepl:
Other way is to have explicit REMOTE argument.

### DIFF
--- a/repository/config_test.go
+++ b/repository/config_test.go
@@ -4,6 +4,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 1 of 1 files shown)


sudoforge (changes requested):
thanks for submitting @wahern's contribution!

as a general rule, for changes like this, we'd like to see unit tests added in concert. there's an existing test for this interface in `//repository:config_test.go`; that file would be a great place to implement tests for the new `GetDefaultString()`.

you can also update your commit message to add a `Closes: #11` trailer -- i think this effectively implements that.

mcepl:
@sudoforge I have added some tests. What’s remaining here? The patch works pretty well for me.

bridge/todosrht/schema.graphqls
sudoforge, line 1: it looks like you accidentally added this during some of the recent force pushes. this file shouldn't be part of this tree.

mcepl: Well, I used it for development from https://git.sr.ht/~sircmpwn/todo.sr.ht/tree/master/item/api/graph/schema.graphqls, and I thought it could be useful for maintenance, but if you don’t want to have it here, I will remove it.

sudoforge: This file might be useful for development of the sourcehut bridge, but this tree is introducing the `git-bug.remote` configuration option, which has nothing to do with the sourcehut bridge.

mcepl: You are absolutely right. Let me fix it.

PR #1468, opened 2025-05-29
feat(cli): allow an alternate Git repository path to be specified

### LINKED ISSUE #1467: Allow the Git repository path to be specified.
In support of `git-bug`'s use as a tool (similar to what's described in #828), this issue promotes composability by allowing the calling program to specify the Git repository's path.  This should be accomplished in the same manner as using the `-C` flag with the `git` command.  Separate handling of the Git directory and working tree, similar to using the Git `--git-dir` and `--work-tree` flags are out-of-scope (at least for the time being.)

For reference, the relevant portion of the documentation for Git's `-C` flag is:

> Run as if git was started in <path> instead of the current working directory. When multiple -C options are
> given, each subsequent non-absolute -C <path> is interpreted relative to the preceding -C <path>. If
> <path> is present but empty, e.g.  -C "", then the current working directory is left unchanged.

smoyer64:
A  flag was added that shares the same semantics as Git's  `-C` flag

Resolves #1467

### DIFF
--- a/commands/execenv/loading_test.go
+++ b/commands/execenv/loading_test.go
@@ -43,7 +43,7 @@ func TestGetRepoPath(t *testing.T) {
 		assert.Equal(t, gitDir, path)
 	})
 
-	t.Run("GIT_DIR supercedes --git-dir", func(t *testing.T) {
+	t.Run("GIT_DIR supersedes --git-dir", func(t *testing.T) {
 		cacheDir, err := os.UserCacheDir()
 		require.NoError(t, err)
 


sudoforge (commented):
this is already a supported workflow using git's `-C` flag:

```
➜ cd $(mktemp -d)


➜ git init .
Initialized empty Git repository in /tmp/tmp.5DzuMyMWOO/.git/


➜ git bug user new -e 'no-reply@sudoforge.com' -n 'sudoforge' --non-interactive
Building cache...

9d2b85433bc7775128e778a1c122c1efc23a748dd035c30dcea673f5d6833e20
```

```
➜ cd $(mktemp -d)


➜ pwd
/tmp/tmp.gOrmIEBMtc


➜ git -C /tmp/tmp.5DzuMyMWOO bug user
9d2b854 sudoforge
```

is there any particular reason you want to add support for this to git-bug?

MichaelMure:
> git -C /tmp/tmp.5DzuMyMWOO bug user

This requires to use the git integration, which I don't think is the right choice for the future (even if it's cute).

That said, I'm not quite fan of the ergonomics of `-c` being a []string (some weird join of path segments) instead of a simple string (a single simple path). It seems to me that it's not the kind of thing we should carry over from git, even for an homogeneous experience.

Additionally, from a UX perspective we should consider how we will in the future instruct the `webui` command to load multiple repo, or a folder containing multiple repos, or ...

commands/root.go
```diff
@@ -5,10 +5,10 @@ import (
 
 	"github.com/spf13/cobra"
 
-	"github.com/git-bug/git-bug/commands/bridge"
-	"github.com/git-bug/git-bug/commands/bug"
+	bridgecmd "github.com/git-bug/git-bug/commands/bridge"
+	bugcmd "github.com/git-bug/git-bug/commands/bug"
 	"github.com/git-bug/git-bug/commands/execenv"
-	"github.com/git-bug/git-bug/commands/user"
+	usercmd "github.com/git-bug/git-bug/commands/user"
```
sudoforge, line 11: this is showing up as a diff because it was removed from HEAD in a commit that's newer than your tree. you should be able to get rid of this automagically by rebasing on top of HEAD.

commands/root.go
```diff
@@ -55,6 +55,7 @@ the same git remote you are already using to collaborate with other people.
 	}
 
 	env := execenv.NewEnv()
+	cmd.PersistentFlags().StringArrayVarP(&env.RepoPath, "repo-path", "C", []string{}, `Run as if git-bug was started in <path> instead of the current working directory.`)
```
sudoforge, line 58: as discussed in the comment thread, i think this should probably be a `string` (that is, not a slice of strings).

commands/execenv/loading.go
```diff
@@ -171,3 +172,15 @@ func CacheBuildProgressBar(env *Env, events chan cache.BuildEvent) error {
 
 	return nil
 }
+
+func getRepoPath(env *Env) (string, error) {
```
sudoforge, line 176: `getRepoPath` should return an error if the path doesn't exist.

commands/execenv/env.go
```diff
@@ -19,11 +19,12 @@ const gitBugNamespace = "git-bug"
...
-	Backend *cache.RepoCache
-	In      In
-	Out     Out
-	Err     Out
+	Repo     repository.ClockedRepo
+	Backend  *cache.RepoCache
+	In       In
+	Out      Out
+	Err      Out
+	RepoPath []string
```
sudoforge, line 27: `RepoPath` should be a `string`

sudoforge (changes requested):

sudoforge:
> This requires to use the git integration, which I don't think is the right choice for the future (even if it's cute).

i mean, okay, **if** this move is made, but moving away from shipping as git-bug isn't something that's happening in the (near) future, doesn't have a complete design, should be aligned with a repository rename, and isn't the way users are expected to be using git-bug today.

since this wouldn't really cause an _issue_, i guess i'd accept it today even though the need for it is rather non-existent.

> Additionally, from a UX perspective we should consider how we will in the future instruct the webui command to load multiple repo, or a folder containing multiple repos, or ...

this isn't the right place for design discussions, but generally, i see the server (a long-lived daemon running in the background) maintaining a list of repositories it should track/interact with. the user would add repositories to the server either imperatively (through a command, or api endpoint via a client application) or dynamically, by running any git-bug command in a repository that isn't already tracked.

ashpool37:
A bit of confusion here and there regarding this entire proposal and pull request.

Git's [`-C`](https://git-scm.com/docs/git#Documentation/git.txt--Cltpathgt) flag is not (just) for specifying the path to the Git repository. It's for changing the current working directory. `git` [literally](https://github.com/git/git/blob/cb3b40381e1d5ee32dde96521ad7cfd68eb308a6/git.c#L313) does `chdir` when it encounters `-C` while parsing `argv`. Sure, you can use it to chdir to the root of a Git repo from somewhere else, but it's mostly used with deeply nested source trees to simplify commands and scripting.

The correct ways to specify the path to the Git repository (the `.git` directory) if you are running `git` from somewhere it can't be discovered automatically are the [`--git-dir`](https://git-scm.com/docs/git#Documentation/git.txt---git-dirltpathgt) flag and the `GIT_DIR` environment variable. As noted in #1467, one common use case for this is separation of working trees from repositories. While this might be out of scope because git-bug doesn't do much with the working tree (the actual tracked files), specifying the path to the `.git` directory is useful on its own, perhaps even more so when using git-bug than when using Git. It simplifies working with [bare repositories](https://git-scm.com/docs/gitglossary/2.39.0#Documentation/gitglossary.txt-aiddefbarerepositoryabarerepository), allowing to interact with bugs without checking out the files, and even using bare Git repos purely for task tracking, making them accessible from anywhere in the file system tree.

### Real use case

[yadm](https://github.com/yadm-dev/yadm) is a nifty POSIX Shell tool for managing "dotfiles". It's a Git wrapper designed to simplify tracking certain paths in your Unix home directory, like `~/.profile`, `~/.bashrc`, certain paths in `~/.config`, etc. Because the files are all over the place and no one wants their home directory to be a giant Git working tree, yadm keeps the bare repository in `~/.local/share/yadm/repo.git`.

`yadm` may be called from anywhere. Before running Git commands, yadm exports `GIT_DIR` to tell Git where the bare repository is. Because of git-bug's Git integration it should be automatically supported by yadm by calling `yadm bug`, but git-bug does not respect `GIT_DIR`, so it fails:

```
$ yadm bug bug
Error: .git not found
```

Worse, if someone calls `yadm bug` from inside another Git working tree, git-bug will actually modify _that_ repository's bugs rather than those of the intended yadm repo.

## This Pull Request

The implementation of `-C` as per this PR superficially mimics Git's `-C` and supports one of its use cases, but it may be confusing since the patched `git-bug` does not actually chdir as `git` does. Instead, all paths are concatenated and the resulting path is used as if it is the Git repository. This is made worse by the fact that `git -C bug` and `git bug -C` may result in completely different behaviours.

### Suggestions

- rename `-C` to `--git-dir`;
- only allow specifying `--git-dir` once and make `RepoPath` a single string (removing the path concatenation behaviour);
- allow specifying the repository path via the `GIT_DIR` environment variable;
- leave everything else as is, do not worry about implementing `--work-dir` for now.

I might be available to implement this soon if everything makes sense.

sudoforge:
:wave: revisiting this.

> Git's [-C](https://git-scm.com/docs/git#Documentation/git.txt--Cltpathgt) flag is not (just) for specifying the path to the Git repository. It's for changing the current working directory. git [literally](https://github.com/git/git/blob/cb3b40381e1d5ee32dde96521ad7cfd68eb308a6/git.c#L313) does chdir when it encounters -C while parsing argv. Sure, you can use it to chdir to the root of a Git repo from somewhere else, but it's mostly used with deeply nested source trees to simplify commands and scripting.

I don't think there is any confusion about how it works -- neither of the respondents claimed that it did anything other than what it does. That said, I would disagree that it is "mostly used with deeply nested source trees", as my own primary use case for it is automation that runs against many independent repositories, from outside of the context of any one of them - and you cannot possibly hope to know how the bulk of other `git` users use it.

> yadm may be called from anywhere. Before running Git commands, yadm exports GIT_DIR to tell Git where the bare repository is. Because of git-bug's Git integration it should be automatically supported by yadm by calling yadm bug[1](https://github.com/git-bug/git-bug/pull/1468#user-content-fn-1-93deff1273b93d9f9c9941e6d9cbc803), but git-bug does not respect GIT_DIR, so it fails.

this is true - but this PR doesn't address `GIT_DIR`, and i'd rephrase the comment from @MichaelMure about refraining from requiring integration with git as something along the lines of "let's refrain from a deeply nested architectural decision that would prevent us from supporting other backends that are not git", as i believe that is the spirit of the comment (but any subsequent comment he makes to clarify this should overwrite this opinion). It was a response to my suggestion that `git -C <path> bug ...` was the recommended approach, anyway.

more to the point, I don't think calling `git-bug` from within another program that manages the current context the user runs it in is a use case that has been considered thus far. This isn't to say that we would reject patches which support doing this, but that those patches need to make sense for `git-bug` independently.

commands/execenv/loading.go
```diff
@@ -171,3 +172,15 @@ func CacheBuildProgressBar(env *Env, events chan cache.BuildEvent) error {
...
+	if len(env.RepoPath) > 0 {
+		return filepath.Join(env.RepoPath...), nil
+	}
+
+	cwd, err := os.Getwd()
```
sudoforge: i don't think that the current working directory is an appropriate default - i think we should instead get the repository's root directory, to support this workflow for users who are in some subdirectory of the repo.

smoyer64:
> more to the point, I don't think calling git-bug from within another program that manages the current context the user runs it in is a use case that has been considered thus far.

This is exactly what I was doing until I realized that:

1. I needed a different `execenv` for other reasons.
2. The parameter parsing done by the Cobra commands wasn't needed and in fact allowed too many options.

So now I'm using `git-bug` as a library.  I have another app I've been toying with that might need this feature though.

That being said, I'm happy to update this PR once we've reached a consensus.  As an alternate view-point, I never run `git-bug` as a `git` sub-command but that's really just habit as `git-bug bug` and `git bug bug` are similar to type.

> i don't think that the current working directory is an appropriate default - i think we should instead get the repository's root directory, to support this workflow for users who are in some subdirectory of the repo.

Great point!

So what I think I've heard so far is:

1. Support `--git-dir` and `GIT_DIR`
2. The default should be the CWD if it contains `.git` or traverse to a parent that does have `.git`.  And error if we're not in a repository at all.

commands/execenv/loading.go
smoyer64: The current working directory is where the search for a `.git` folder starts - https://github.com/git-bug/git-bug/blob/7468b3793fbf1eeef4637d91291494fd8454ae7c/commands/execenv/loading.go#L18-L28

smoyer64: But for a GoGitRepo, it searches the CWD and parents looking for a Git repository: https://github.com/git-bug/git-bug/blob/7468b3793fbf1eeef4637d91291494fd8454ae7c/repository/gogit.go#L58-L59

smoyer64:
@sudoforge Can you re-review this when you get a chance?

commands/root.go
```diff
@@ -21,6 +21,10 @@ git-bug use git objects to store the bug tracking separated from the files
 history. As bugs are regular git objects, they can be pushed and pulled from/to
 the same git remote you are already using to collaborate with other people.
 
+By default, git-bug manages bugs and users in the Git repository containing
+the current working directory.  This behavior can be changed using the --git-dir
+flag described below, or by setting the GIT-DIR environment variable, which
```
MichaelMure, line 26:
```suggestion
flag described below, or by setting the GIT_DIR environment variable, which
```

commands/execenv/env.go
```diff
@@ -24,6 +24,7 @@ type Env struct {
 	In      In
 	Out     Out
 	Err     Out
+	GitDir  string
```
MichaelMure, line 27: As we discussed, let's remove this from `Env`. There is better ways and there is no good reason to have this here.

commands/root.go
```diff
@@ -55,6 +59,7 @@ the same git remote you are already using to collaborate with other people.
 	}
 
 	env := execenv.NewEnv()
+	cmd.PersistentFlags().StringVar(&env.GitDir, "git-dir", "", `run as if git-bug was started in <path>`)
```
MichaelMure, line 62: should we mention the env variable in the comment?

PR #1497, opened 2025-11-20
docs(cli): fix user show command documentation

waldyrious:
As pointed out in [this comment](https://github.com/git-bug/git-bug/issues/530#issuecomment-3558997281), the documentation and CLI help incorrectly suggested that the command to display a user identity was `git-bug user user show USER_ID`, but the actual command is `git-bug user show USER_ID`.

This change fixes the `Use` field in `user_show.go` to use correct `show [USER_ID]` form instead of `user show [USER_ID]`, and updates the documentation and manpages accordingly.

### DIFF
--- a/commands/user/user_show.go
+++ b/commands/user/user_show.go
@@ -20,7 +20,7 @@ func newUserShowCommand(env *execenv.Env) *cobra.Command {
 	options := userShowOptions{}
 
 	cmd := &cobra.Command{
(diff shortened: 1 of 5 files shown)


sudoforge (approved):
good catch, and thanks for this contribution!

+1 for the great commit subject and message, and for regenerating the documentation :)

PR #1506, opened 2025-12-18
Verbose pull push

mcepl:
feat(cli): add verbose output to push and pull commands

Add -v/--verbose flag to git-bug push and pull commands for better user
visibility and debugging.

- push command: verbose mode logs push details to stderr
- pull command: verbose mode logs each operation to stderr, normal mode shows summary
- Added comprehensive test coverage for both commands
- Documentation updated automatically via build system

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,6 @@
+{
+  "name": "git-bug",
+  "lockfileVersion": 3,
+  "requires": true,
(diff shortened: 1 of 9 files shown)


package-lock.json
sudoforge: can you revert changes made to this file, please? it doesn't seem like it should be part of this tree.

PR #1522, opened 2026-02-24
feat: use git-describe for setting the version string

yarikoptic:
This change uses git-describe to set the version string, which improves the user experience by always showing the most recent tag, with additional git data (such as the delta of a commits between that tag and HEAD). 

Before this change, we see the following output:

    git-bug undefined 7468b3793fbf go1.24.9 linux amd64

With this change, the output becomes:

    git-bug v0.10.1-24-gb957ba2e-dirty b957ba2e5947/dirty go1.24.9 linux amd64

The duplicated VCS information does add noise, however, this will be dealt with in a future commit.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@ endif
 
 SYSTEM=$(shell nix eval --impure --expr 'builtins.currentSystem' --raw 2>/dev/null || echo '')
 
-TAG:=$(shell git describe --match 'v*' --always HEAD)
+TAG:=$(shell git describe --match 'v*' --always --dirty --broken)
 LDFLAGS:=-X main.version="${TAG}"
 
 all: build
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@ endif
 
 SYSTEM=$(shell nix eval --impure --expr 'builtins.currentSystem' --raw 2>/dev/null || echo '')
 
-TAG:=$(shell git name-rev --name-only --tags HEAD)
+TAG:=$(shell git describe --match 'v*' --always HEAD)
 LDFLAGS:=-X main.version="${TAG}"
 
 all: build


sudoforge (changes requested):
hey, thanks for thinking about this and submitting a patch!

i went with `name-rev` specifically because i wanted to avoid confusing end users with:

1) a version string when they aren't really using a binary built from a versioned commit,
2) the additional commit hash from `git-describe`, which on an actual release binary would duplicate the third column of data (which shows the commit hash that the binary was built on). yes, this can be stripped, but i wasn't sure if at this point `git-describe` is really providing useful information.

i also wanted to force downstream maintainers to explicitly set `main.version` when building git-bug in order to get a version string embedded, basically as an indicator that the binary was actually built as a release binary, and otherwise show "undefined" if you were using a binary built from some random tree.

however, after noodling this for a while, i think i'm swayed in the other direction. i'm okay with using `git-describe` - but let's strip the commit hash from the output, though, because as mentioned, this would just be redundant information on an actual release binary (and for any custom-built binary, end users can just run `git <log|rev-list> <tag>..<commit shown in version string>` as needed).

---

actually, i have to correct myself - `git-describe` won't show the commit hash suffix if the object passed to it is what the tag is pointing to. i still think we should remove it, though, because the version string already embeds the commit hash (and notes whether or not the workspace was dirty and may contain local changes -- i like this format better than what `git-describe` provides, and it's automatic, and not part of the `main.version` variable).

yarikoptic:
thanks for giving it a thought!  indeed if on a tag, will not add commit hexsha, e.g.
```manu
❯ git describe v0.10.1
v0.10.1
```

and interesting "feature" is that git knows how to treat them directly (without needing to parse out the hexsha), e.g.

```
❯ git describe
v0.10.1-24-g4514cd9b
❯ git diff --stat origin/master..v0.10.1-24-g4514cd9b
 Makefile | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)
```

sudoforge:
> and interesting "feature" is that git knows how to treat them directly (without needing to parse out the hexsha)

TIL! i only rarely actually deal with tags and named descriptors, i generally work off of detached trees (direct commits).

to summarize my earlier comment, what i basically want is the "most recent tag" and the "number of commits from there to you". unfortunately, it doesn't look like `git-describe` provides a flag for explicitly _excluding_ the sha, so we'll need to pipe that into `awk` or `sed` to explicitly remove it.

yarikoptic:
gotcha, because you anyways ALWAYS include commit hexsha in the output of the `version`.

FWIW -- related, in the past I asked git-annex's Joey to provide some `--version` so it would list only the actual version without any other info, so it could be included as is in a uniform fashion among version and in 'standard' ready for comparison form (as to say if current version is greater or less than some other version), hence 

```shell
❯ git-annex version --raw
10.20251029
```
and for some after release point

```
❯ git-annex version --raw
10.20260213+git57-gffa771e735-1~ndall+1
```
was born! ;)  

(I personally just prefer more prevalent `--version` as option to the command)

And that is where then, to a degree, having hexsha embedded is good and bad
- good: allows for identification and exact comparison of the `--version` output alone
- bad: comparison might be misleading since e.g. two branches/rewrites might have the same number of commits off the tag but different hexshas

sudoforge:
i'm totally open to that idea, and i think using `--raw` makes more sense than `--version`, since you're already writing `git bug version` (`git bug version --version` feels too repetitive, and "raw" is a little more clear as a flag, where someone can infer that they're going to "get the raw version"). i've created an issue (#1528) to track that.

Makefile
```diff
@@ -6,7 +6,7 @@ endif
 
 SYSTEM=$(shell nix eval --impure --expr 'builtins.currentSystem' --raw 2>/dev/null || echo '')
 
-TAG:=$(shell git name-rev --name-only --tags HEAD)
+TAG:=$(shell git describe --match 'v*' --always HEAD)
```
sudoforge: i've thought about this a little more and as part of an upcoming change to the build, vcs information will not be available within the build sandbox, so the only value for any sort of version information will have to be set via the `main.version` ldflag.

let's change this to be a bit more comprehensive:


```suggestion
TAG:=$(shell git describe --match 'v*' --always --dirty --broken)
```

commitish references can't be passed with `--dirty` or `--broken`, but this still defaults to `HEAD`.

sudoforge (approved):
thanks again for submitting this change!

sudoforge:
we are having a small technical issue with our gitlab import test (the api token is expired) which is preventing this from landing. i'll enqueue it when the underlying issue is resolved :)

PR #1526, opened 2026-02-26
feat(bugs): add assignee support

### LINKED ISSUE #112: Assign Users
With the identity section almost done woudn't it make sense to add the ability to assign bugs to a user .
`assign bug <id> <username>`
Some of the things we need to take care of:
1) Only the author of the bug is allowed to assign it to someone.
2) Section of assignees when the snapshot of the bug is compiled.
3) Addition to the GraphQL API.

andrust:
Add ability to assign users to bugs with:
- SetAssigneeOperation for tracking assignee changes
- CLI commands: bug assignee, bug assignee set, bug assignee clear
- GraphQL mutations and queries
- Query filter: assignee:<user>
- TermUI assignee selection
- WebUI assignee menu and timeline display
- Bridge support for Jira (GitHub/GitLab marked TODO)

Closes #112


MichaelMure:
To be clear, a LLM wrote that, right?

At first glance it's looks good but I suspect that it would be better to have a **list** of assignees instead of just one. At the very least github supports that, so that would make things easier in the bridge.

andrust:
Being a c++ dev I had to resort to AI usage to make up for the missing go experience.

Using an array would probably be the right call, I will give it a proper look. But this is already way beyond my go level tbh.

MichaelMure:
To be clear, using a LLM is ok ... as long as you read and fix the output before sending it for review ;-)

PR #1555, opened 2026-04-22
Github PR support

### LINKED ISSUE #373: support pull requests
Might be slightly offtopic because the pull requests are no bugs but the repository limits the scope to bugs. However, i want to expand the vision of distributed development.

I think it would be very interesting if pull requests would also be supported in the same way as bugs. For projects such as [lemmy](https://github.com/LemmyNet/lemmy) this would be a really great feature, because lemmy has at the moment 3 repositories on github, gitea and gitlab. So you can make a pull request on any of those platforms, but for the issues you have to look at github.

I think a really great improvement would be to decentralize pull requests the same way as bugs. That could allow a development style where you just have different mirrors on github, gitea, gitlab, etc. but for the bugs and pull requests you can use `git-bug`, so you wouldnt need to use github or so.

And it would make this decentralized style of development actually easier than the status quo, since you do not need to look up multiple places (gitlab, github, gitea) to find out if a pull request has already done something you want to do. You just can look that up in `git-bug`.

So i think this would be really exciting to avoid being dependent on github or so. You can use github to gain popularity and being easily found, but you github is actually not important for the project as it is just another code mirror.

gilescope:
Claude did this. Feel free to take it / parts / or pass.

Closes #373 (feat: support for pull requests).

PR support itself doesn't have any philosophical questions, but CI for a PR does. Linux doesn't have blocking CI gates — it has trusted approvers. I don't know what decentralised CI looks like though I suspect proof of reproducibility probably comes into it. Either way this is probably out of the scope of this PR — for now it has some links to centralised CI.

I thought this would be a good checkpoint as am going to pull in the project PR — will branch off here as I do that.

## Running the webui across many repos

Two flags on `git-bug webui` let one process serve a whole directory of checked-out repos, so the UI becomes a single pane of glass across an org:

- `--repo <path>` (repeatable) — register an extra repo. Its display name comes from the parent dir plus basename (so `/a/b/myorg/myrepo` shows up as `myorg/myrepo`).
- `--root <dir>` — scan `<root>/<org>/<repo>/.git` siblings and register each one found. Only directories with git-bug refs are included; symlinks that escape `<root>` are rejected.

Multi-repo mode forces `--read-only` (per-repo identity selection isn't wired up yet).

Example layout I use locally:

```
~/git/work/
  my-github-org/
    repo1/      <- git-bug synced
    repo2/        <- git-bug synced
    ...
  dir-of-forked-stuff/
    foo/                <- git-bug synced
    bar/                <- git-bug synced
```

Serve the lot:

```
git-bug webui --root ~/git/work --read-only
```

Or register individual repos:

```
git-bug webui \
  --repo ~/git/work/my-github-org/repo1 \
  --repo ~/git/work/my-github-org/repo2 \
  --read-only
```

The list view shows every registered repo; a repo-scoped search can be done via the `repo:` / `org:` query filters, and the Sync button can either sync every registered repo or just the one you're looking at.


MichaelMure:
While that could be interesting for if/when we implement PR support, imho it really should be a separate entity (that is, not retrofitted into `bug`). Unfortunately, that invalidates most of your PR.

Before any token are spent on implementation, we really need to come to an agreement on:
- what the data model would be (the most critical point)
- to some extent, what the workflow would be
- how the different UIs would work

Issue #1433, opened 2025-05-16
web: read-only mode shows a blank page


sudoforge:
starting the web ui with `--read-only` results in a blank page being shown. there are no errors logged.

PR #1559, opened 2026-05-23
Fix blank web UI in read-only mode

arunisaac:
Hi,

I've come up with a fix for #1433.

Go and typescript are not the languages I normally code in. So, please bear with me if I got some detail wrong.

Thanks!


MichaelMure:
The new webui has just been merged and therefore this is obsolete. Sorry!

PR #1569, opened 2026-06-08
Update README.md

cfkaran2:
fix(doc): Broken link in spec

There are a couple of broken links in the spec that are referencing `[Data model - the rational](../data-model-rational.md)`.  Unfortunately, there doesn't seem to be a document there.  The closest thing I've found is to `../design/data-model.md`, which is where I've pointed the links to.


MichaelMure (approved):

cfkaran2:
@MichaelMure Apparently, this PR is blocked because merging is blocked; any ideas on how I'm supposed to make the PR happen?

PR #1572, opened 2026-06-14
webui: markdown viewer

nijotz:
The directory view already renders a folder's README.md as markdown. This adds that same viewer for other markdown files in the repo with a Preview/Code toggle with a FileContent component (with a story and snapshot test). I could see the FileContent component being a switch for other file types as well, like images.


MichaelMure:
> I could see the FileContent component being a switch for other file types as well, like images.

Agree. I think the server side would add the correct content-type header, but I didn't test that. It could make sense in the future to rework how the buttons and all are displayed in that header (github stuff that in a single line, with the history higher), but it works well for now.

I've invited you to the repo if you want to continue contributing.

